PR #118, opened 2022-07-25
feat(util): add fetch and runtime functions

mkilpatrick:
This allows the use of cross-fetch in templates. Since fetch is only supported natively in Node 18+, we need to allow users to be able to fetch in 16+ uses a 3rd party library. cross-fetch fits this build. Additionally, we only want cross-fetch imported during dev mode since this library does not work with Deno for production builds and Deno supports fetch natively.

The impetus for this change is removing fetch-anywhere from the starter since it has 2 high vulnerability issues.

Tested against the starter repo.


packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -35,6 +36,11 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
               assetFileNames: "assets/static/[name]-[hash][extname]",
               chunkFileNames: "assets/static/[name]-[hash].js",
             },
+            plugins: [
+              externalGlobals({
+                "cross-fetch": "fetch",
```
mancusi, line 41: I don't think we should do this. It's an unclear dependency and surprising/atypical behavior. 

It's odd that we'd bundle cross-fetch dependencies specifically to the fetch global due to 
 our dependency on deno.

Instead of doing this why don't we provide a a mechanism natively through our package that allows users to fetch across all environments that we support. 

```js
const fetch = (// fetchAPI) => {
  if node {
    // cross-fetch or node-fetch
  }
  
  // otherwise we're deno
  fetch(fetchApi)
}
```

mancusi (changes requested):

packages/pages/src/vite-plugin/build/build.ts
mancusi: probably: https://www.npmjs.com/package/node-fetch

mkilpatrick: Updated to export some new util functions instead.

packages/pages/src/util/fetch.ts
```diff
@@ -0,0 +1,18 @@
...
+  // Check if running in Node 17 or less, which does not support native fetch
+  if (runtime.name == "node" && runtime.getNodeMajorVersion() < 18) {
+    const { default: crossFetch } = await import("cross-fetch");
+    return crossFetch(input, init);
+  }
+
+  return fetch(input, init);
+};
+
+export { fetchInternal as fetch };
```
mancusi, line 18: Add documentation since this component is exported publicly

mancusi: same for runtime

packages/pages/src/util/runtime.ts
```diff
@@ -0,0 +1,30 @@
+class Runtime {
```
mancusi, line 1: can you add tests for these?

packages/pages/src/util/fetch.ts
mancusi: We will probably also want at least a readme for this package as well since it'd otherwise be undocumented outside of the starter

mancusi (changes requested):

packages/pages/src/util/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+- Node (during local dev)
+- Deno (during production build/generation)
+- the browser (any executed frontend code)
+
+This can be useful when the function or library differs depending on the runtime. `fetch` is one example as it's not native to Node < v18.
+
+## fetch()
+
+A custom fetch implementation that determines which fetch library to use depending on the current runtime. When locally developing, Node is used. Since fetch is only native starting in v18 and the version on the user's machine is up to them, we need to polyfill fetch. Under the hood this uses cross-fetch.
```
mancusi, line 15: > When locally developing, Node is used
node is used locally for *devmode* and for the production build. in the local realistic generation mode deno is used to actually render the templates.

packages/pages/src/util/fetch.ts
```diff
@@ -15,4 +15,13 @@ const fetchInternal = async (
   return fetch(input, init);
 };
 
+/**
+ * A custom fetch implementation that determines which fetch library to use
+ * depending on the current runtime. When locally developing, Node is used.
```
mancusi, line 20: similar local dev comment

mancusi (changes requested):
just minor comments clarifying when deno comes into play

mancusi (approved):

PR #120, opened 2022-07-25
docs: add contributing guide and add packages to README

mkilpatrick:


CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
...
+
+The PagesJS repo is a monorepo using pnpm workspaces. The package manager used to install and link dependencies must be [pnpm](https://pnpm.io/).
+
+To develop and test:
+
+1. Run `pnpm i` in the root folder.
+
+2. Make your changes. Then in /packages/pages run `pnpm pack`.
+
+3. Reference the local pack file in another project that uses this one, such as https://github.com/yext/pages-starter-react-locations. You can reference it in the starter's package.json like `"@yext/pages": "file:../pages/packages/pages/yext-pages-1.0.0.tgz"` (assuming both projects live in the same root folder).
```
nbramblett, line 15: would suggest replacing 1.0.0 with [version] to indicate that it can change, so people know not to just blindly paste it in

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
...
+
+4. If you make changes in PagesJS make sure to `rm -rf package-lock.json node_modules/ && npm i` in the starter first to ensure you pull in the latest packed changes.
```
nbramblett, line 17: We can (and probably should) only remove node_modules/@yext/pages rather than the full folder

mkilpatrick: good call, much faster to iterate on too

nbramblett (approved):

PR #122, opened 2022-07-26
feat(components): add image component and export it

nbramblett:
Adds the image component from proserve with a few adjustments - styles are inline instead of using TailwindCSS and I cleaned up the types a bit so that it no longer has a transitive dependency on @yext/types.

Also updated the version to beta.1 since it's a new phase of the plugin.

J=SUMO-4686
TEST=manual

Ran against both dev and prod builds on the starter repo, verifying that I could import and use the <Image> component as expected.


packages/pages/src/components/Image/Image.tsx
```diff
@@ -0,0 +1,188 @@
...
+    src,
+    srcSet,
+    sizes,
+    alt = "",
+    loading = "lazy",
+    ...rest
+  } = props;
+
+  // One of `src`, `imageField` is required to be populated in props, so manually
+  //    casting imageField as `imageType` here to ignore typescript type checking
```
mkilpatrick, line 116: weird spacing. Can you run `pnpm fmt`? Though the auto formatter should have run. Maybe it's not checking in this directory for some reason? Though I think it should be.

packages/pages/src/components/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./Image";
+export * from "./AspectRatio";
```
mkilpatrick, line 2: Does this need to be exported to users? It looks like they only use Image.

packages/pages/src/components/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface StandardProps {
+  yaTrack?: string;
```
mkilpatrick, line 2: Is this actually used by the component anywhere? Should we remove it?

packages/pages/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/pages",
-  "version": "1.0.0-beta.0",
+  "version": "1.0.0-beta.1",
```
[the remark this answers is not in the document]
nbramblett: Ah ok, will do.

packages/pages/src/components/AspectRatio/AspectRatio.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const mainDivStyle = maxWidth
+    ? { maxWidth: `${maxWidth}px`, maxHeight: `${maxWidth * ratio}px` }
+    : undefined;
+  const outerWrapperStyle: React.CSSProperties = {
+    paddingBottom: `${ratio * 100}%`,
+    position: "relative",
+    width: "100%",
+    height: 0,
+  };
+  const innerWrapperStyle: React.CSSProperties = {
```
mancusi, line 26: move all constants outside of the component

packages/pages/src/components/AspectRatio/AspectRatio.tsx
```diff
@@ -0,0 +1,45 @@
...
+    position: "absolute",
+    top: "0",
+    right: "0",
+    bottom: "0",
+    left: "0",
+  };
+
+  return (
+    <div className="AspectRatio" style={mainDivStyle}>
```
mancusi, line 35: these class names should be configurable

packages/pages/src/components/AspectRatio/AspectRatio.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+
+interface AspectRatioProps {
+  children?: any;
```
mancusi, line 4: React.Children?

packages/pages/src/components/Image/Image.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { StandardProps } from "../types";
+import { AspectRatio } from "../AspectRatio";
+
+export interface ImageType {
+  height: number;
+  width: number;
+  url: string;
+}
+
+export interface ImagePropsShared
```
mancusi, line 18: similar "detailed documentation" comment

packages/pages/src/components/Image/Image.tsx
```diff
@@ -0,0 +1,188 @@
...
+}
+
+export interface ImagePropsWithField extends ImagePropsShared {
+  src?: never;
+  imageField: ImageType;
+}
+
+export type ImageProps = ImagePropsWithSrc | ImagePropsWithField;
+
+/**
```
mancusi, line 51: The JS doc is okay but we should be moving that to the to ts types as it's a more consumer-friendly

mancusi (changes requested):

packages/pages/src/components/Image/Image.tsx
nbramblett: It looks like intentional reverse-indentation to indicate it's part of the same comment paragraph.

packages/pages/src/components/index.ts
nbramblett: Only Image and AspectRatio should be exported, because those are dictated by the index.ts in the subdirectories.

I'm not sure if this is objectively cleaner but I do think one nice benefit is that in the future we could automate a check to verify that all the components have an export, and if they're in an easy format like `export * from "./FolderName"` then that's trivial to verify.

packages/pages/src/components/types.ts
nbramblett: It's extended by ImagePropsShared, I put it in a higher-level types file since I suspect other components might use it, as it seems to be a common control prop scheme for analytics settings.

packages/pages/src/components/index.ts
nbramblett: Unexporting AspectRatio since it's basically just an internal component for Image

packages/pages/src/components/AspectRatio/AspectRatio.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+
+interface AspectRatioProps {
```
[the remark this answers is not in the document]
nbramblett: Ack for this particular file since I'm not exporting AspectRatio anymore. Done for Image.

mancusi:
This is less RFPT than ever before --  moving to a draft PR until we get a clear spec.

PR #144, opened 2022-07-30
feat(components): add `Image` component

EmilyZhang777:
This PR sets up the components/ export and adds an Image component.

J=SUMO-4687
TEST=manual

Ran against both dev and production builds on the starter repo, made sure I can import the Image component and all functionalities work as specified [here](https://www.notion.so/Images-c04403f8e8f14e7bb3869894d5a4860b).

### DIFF
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -82,17 +82,15 @@
     "vitepress": "1.0.0-alpha.4"
   },
   "devDependencies": {
-    "@babel/preset-env": "^7.16.11",
-    "@babel/preset-typescript": "^7.16.7",
     "@microsoft/api-documenter": "^7.17.9",
     "@microsoft/api-extractor": "^7.22.2",
-    "@types/babel__core": "^7.1.18",
+    "@testing-library/react": "^12.1.5",
     "@types/escape-html": "^1.0.2",
     "@types/express": "^4.17.13",
     "@types/express-serve-static-core": "^4.17.29",
     "@types/fs-extra": "^9.0.13",
     "@types/glob": "^7.2.0",
-    "@types/jest": "^27.5.1",
+    "@types/jest": "^27.5.2",
     "@types/lodash": "^4.14.182",
     "@types/node": "^17.0.45",
     "@types/react": "^17.0.40",
@@ -115,23 +113,10 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2"
   },
-  "babel": {
-    "presets": [
-      [
-        "@babel/preset-env",
-        {
-          "targets": {
-            "node": "current"
-          }
-        }
-      ],
-      "@babel/preset-typescript"
-    ]
-  },
   "jest": {
     "rootDir": "..",
     "collectCoverageFrom": [
-      "<rootDir>/pages/src/**/*.{ts,js}"
+      "<rootDir>/pages/src/**/*.{ts,tsx,js}"
     ],
     "globals": {
       "ts-jest": {
@@ -139,19 +124,21 @@
           "allowSyntheticDefaultImports": true,
           "esModuleInterop": true,
           "allowJs": true,
-          "target": "ESNext"
+          "target": "ESNext",
+          "jsx": "react"
         }
       }
     },
     "verbose": true,
     "testMatch": [
-      "<rootDir>/pages/src/**/*.test.{ts,js}"
+      "<rootDir>/pages/src/**/*.test.{ts,tsx,js}"
     ],
     "transform": {
       "\\.[jt]sx?$": "ts-jest"
     },
     "moduleNameMapper": {
       "^(.+)\\.js$": "$1"
-    }
+    },
+    "testEnvironment": "jsdom"
   }
 }
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -90,7 +90,7 @@
     "@types/express-serve-static-core": "^4.17.29",
     "@types/fs-extra": "^9.0.13",
     "@types/glob": "^7.2.0",
-    "@types/jest": "^27.5.2",
+    "@types/jest": "^27.5.1",
     "@types/lodash": "^4.14.182",
     "@types/node": "^17.0.45",
     "@types/react": "^17.0.40",
@@ -138,7 +138,6 @@
     },
     "moduleNameMapper": {
       "^(.+)\\.js$": "$1"
-    },
-    "testEnvironment": "jsdom"
+    }
   }
 }
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -107,6 +107,8 @@
     "glob": "^7.2.0",
     "jest": "^28.1.3",
     "jest-environment-jsdom": "^28.1.3",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
     "ts-jest": "^28.0.7"
   },
   "peerDependencies": {
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -12,7 +12,8 @@
   "exports": {
     ".": "./dist/index.js",
     "./vite-plugin": "./dist/vite-plugin/plugin.js",
-    "./util": "./dist/util/index.js"
+    "./util": "./dist/util/index.js",
+    "./components": "./dist/components/index.js"
   },
   "typesVersions": {
     "*": {
@@ -24,6 +25,9 @@
       ],
       "util": [
         "dist/types/src/util/index.d.ts"
+      ],
+      "components": [
+        "dist/types/src/components/index.d.ts"
       ]
     }
   },
--- a/packages/pages/src/components/image/image.test.tsx
+++ b/packages/pages/src/components/image/image.test.tsx
@@ -67,10 +67,12 @@ describe("Image", () => {
     const placeholder = <div>{placeholderText}</div>;
 
     render(
-        <Image
-            image={{image: {...image.image, url: "https://a.mktgcdn.com/p/2x1.jpg"}}}
-            placeholder={placeholder}
-        />
+      <Image
+        image={{
+          image: { ...image.image, url: "https://a.mktgcdn.com/p/2x1.jpg" },
+        }}
+        placeholder={placeholder}
+      />
     );
 
     expect(screen.getByText(placeholderText)).toBeTruthy();
@@ -190,7 +192,14 @@ describe("validateRequiredProps", () => {
     const logMock = jest.spyOn(console, "warn").mockImplementation(() => {});
 
     expect(logMock.mock.calls.length).toBe(0);
-    validateRequiredProps(ImageLayout.INTRINSIC, imgWidth, imgHeight, width, undefined, undefined);
+    validateRequiredProps(
+      ImageLayout.INTRINSIC,
+      imgWidth,
+      imgHeight,
+      width,
+      undefined,
+      undefined
+    );
     expect(logMock.mock.calls.length).toBe(1);
 
     jest.clearAllMocks();
@@ -201,7 +210,14 @@ describe("validateRequiredProps", () => {
 
     expect(logMock.mock.calls.length).toBe(0);
 
-    validateRequiredProps(ImageLayout.FIXED, imgWidth, imgHeight, undefined, undefined, undefined);
+    validateRequiredProps(
+      ImageLayout.FIXED,
+      imgWidth,
+      imgHeight,
+      undefined,
+      undefined,
+      undefined
+    );
 
     expect(logMock.mock.calls.length).toBe(1);
     jest.clearAllMocks();
@@ -212,7 +228,14 @@ describe("validateRequiredProps", () => {
 
     expect(logMock.mock.calls.length).toBe(0);
 
-    validateRequiredProps(ImageLayout.FIXED, imgWidth, imgHeight, -100, undefined, undefined);
+    validateRequiredProps(
+      ImageLayout.FIXED,
+      imgWidth,
+      imgHeight,
+      -100,
+      undefined,
+      undefined
+    );
 
     expect(logMock.mock.calls.length).toBe(1);
     jest.clearAllMocks();
@@ -223,7 +246,14 @@ describe("validateRequiredProps", () => {
 
     expect(logMock.mock.calls.length).toBe(0);
 
-    validateRequiredProps(ImageLayout.ASPECT, imgWidth, imgHeight, undefined, undefined, undefined);
+    validateRequiredProps(
+      ImageLayout.ASPECT,
+      imgWidth,
+      imgHeight,
+      undefined,
+      undefined,
+      undefined
+    );
 
     expect(logMock.mock.calls.length).toBe(1);
     jest.clearAllMocks();
@@ -234,7 +264,14 @@ describe("validateRequiredProps", () => {
 
     expect(logMock.mock.calls.length).toBe(0);
 
-    validateRequiredProps(ImageLayout.FILL, -100, imgHeight, undefined, undefined, undefined);
+    validateRequiredProps(
+      ImageLayout.FILL,
+      -100,
+      imgHeight,
+      undefined,
+      undefined,
+      undefined
+    );
 
     expect(logMock.mock.calls.length).toBe(1);
     jest.clearAllMocks();
@@ -243,13 +280,27 @@ describe("validateRequiredProps", () => {
 
 describe("getImageSizeForFixedLayout", () => {
   it("properly sets fixedWidth and fixedHeight", () => {
-    expect(getImageSizeForFixedLayout(imgWidth, imgHeight, widths, undefined, undefined))
-        .toEqual({fixedWidth: imgWidth, fixedHeight: imgHeight, fixedWidths: widths});
-    expect(getImageSizeForFixedLayout(imgWidth, imgHeight, widths, width, undefined))
-        .toEqual({fixedWidth: width, fixedHeight: height, fixedWidths: [width]});
-    expect(getImageSizeForFixedLayout(imgWidth, imgHeight, widths, undefined, height))
-        .toEqual({fixedWidth: width, fixedHeight: height, fixedWidths: [width]});
-    expect(getImageSizeForFixedLayout(imgWidth, imgHeight, widths, width, height))
-        .toEqual({fixedWidth: width, fixedHeight: height, fixedWidths: [width]});
+    expect(
+      getImageSizeForFixedLayout(
+        imgWidth,
+        imgHeight,
+        widths,
+        undefined,
+        undefined
+      )
+    ).toEqual({
+      fixedWidth: imgWidth,
+      fixedHeight: imgHeight,
+      fixedWidths: widths,
+    });
+    expect(
+      getImageSizeForFixedLayout(imgWidth, imgHeight, widths, width, undefined)
+    ).toEqual({ fixedWidth: width, fixedHeight: height, fixedWidths: [width] });
+    expect(
+      getImageSizeForFixedLayout(imgWidth, imgHeight, widths, undefined, height)
+    ).toEqual({ fixedWidth: width, fixedHeight: height, fixedWidths: [width] });
+    expect(
+      getImageSizeForFixedLayout(imgWidth, imgHeight, widths, width, height)
+    ).toEqual({ fixedWidth: width, fixedHeight: height, fixedWidths: [width] });
   });
 });
--- a/packages/pages/src/components/image/image.test.tsx
+++ b/packages/pages/src/components/image/image.test.tsx
@@ -81,7 +81,6 @@ describe("Image", () => {
     );
 
     expect(screen.getByText(placeholderText)).toBeTruthy();
-
     expect(logMock.mock.calls.length).toBe(1);
     expect(logMock.mock.calls[0][0]).toBe(`Invalid image url: ${invalidUrl}.`);
 
--- a/packages/pages/src/components/image/image.test.tsx
+++ b/packages/pages/src/components/image/image.test.tsx
@@ -1,7 +1,7 @@
 /**
  * @jest-environment jsdom
  */
-import React from "react";
+import * as React from "react";
 import {
   Image,
   validateRequiredProps,
--- a/packages/pages/src/components/image/image.test.tsx
+++ b/packages/pages/src/components/image/image.test.tsx
@@ -9,7 +9,7 @@ import {
   handleLayout,
   getImageSizeForFixedLayout,
 } from "./image";
-import { ImageLayout } from "./types";
+import { ImageLayoutOption } from "./types";
 import { render, screen } from "@testing-library/react";
 
 const imgWidth = 20;
@@ -35,8 +35,9 @@ describe("Image", () => {
     render(
       <Image
         image={image}
-        layout={ImageLayout.FIXED}
+        layout={ImageLayoutOption.FIXED}
         width={width}
+        height={height}
         style={{ objectFit: overrideObjectFit }}
         imgOverrides={{ src: overrideSrc }}
       />
@@ -151,9 +152,9 @@ describe("getImageUUID", () => {
 });
 
 describe("handleLayout", () => {
-  it(`properly sets aspectRatio when layout is ${ImageLayout.INTRINSIC} and aspectRatio is provided`, () => {
+  it(`properly sets aspectRatio when layout is ${ImageLayoutOption.INTRINSIC} and aspectRatio is provided`, () => {
     const { imgStyle } = handleLayout(
-      ImageLayout.INTRINSIC,
+      ImageLayoutOption.INTRINSIC,
       imgWidth,
       imgHeight,
       imgUUID,
@@ -166,9 +167,9 @@ describe("handleLayout", () => {
     expect(imgStyle.aspectRatio).toEqual(aspectRatio.toString());
   });
 
-  it(`properly sets aspectRatio when layout is ${ImageLayout.INTRINSIC} and aspectRatio is not provided`, () => {
+  it(`properly sets aspectRatio when layout is ${ImageLayoutOption.INTRINSIC} and aspectRatio is not provided`, () => {
     const { imgStyle } = handleLayout(
-      ImageLayout.INTRINSIC,
+      ImageLayoutOption.INTRINSIC,
       imgWidth,
       imgHeight,
       imgUUID,
@@ -181,9 +182,9 @@ describe("handleLayout", () => {
     expect(imgStyle.aspectRatio).toEqual(`${imgWidth} / ${imgHeight}`);
   });
 
-  it(`properly sets src, imgStyle and widths when layout is ${ImageLayout.FIXED} and only width is provided`, () => {
+  it(`properly sets src, imgStyle and widths when layout is ${ImageLayoutOption.FIXED} and only width is provided`, () => {
     const { src, imgStyle, widths } = handleLayout(
-      ImageLayout.FIXED,
+      ImageLayoutOption.FIXED,
       imgWidth,
       imgHeight,
       imgUUID,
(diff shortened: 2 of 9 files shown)


packages/pages/src/components/image/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  };
+};
+
+export type ImageProps = {
+  image: ImageType;
+  className?: string;
+  width?: number;
+  height?: number;
+  aspectRatio?: number;
+  layout?: "intrinsic" | "fixed" | "aspect" | "fill";
```
mancusi, line 24: use a typescript enum for this

```ts
enum Layout = {
  INTRINSIC = "intrinsic"
  ...
}
```

packages/pages/src/components/image/types.ts
```diff
@@ -0,0 +1,28 @@
...
+export type ImageType = {
+  image: {
+    height: number;
+    width: number;
+    url: string;
+    thumbnails?: ThumbnailType[];
+  };
+};
+
+export type ImageProps = {
```
mancusi, line 18: make sure to document all of these props

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const imgHeight: number = image.image.height;
+  const imgUUID: string = image.image.url.split("/")[4];
+
+  style.objectFit = style.objectFit || "cover";
+  style.objectPosition = style.objectPosition || "center";
+
+  let widths: number[] = [100, 320, 640, 960, 1280, 1920];
+  let src: string = getImageUrl(imgUUID, 500, 500);
+
+  switch (layout) {
```
mancusi, line 41: can you extract this to a separate function?

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  if (layout != "fixed" && (width || height)) {
+    console.warn(
+      "Width or height is passed in but layout is not fixed. These will have no impact. If you want to have a fixed height or width then set layout to fixed."
+    );
+  }
+
+  const imgWidth: number = image.image.width;
+  const imgHeight: number = image.image.height;
+  const imgUUID: string = image.image.url.split("/")[4];
```
mancusi, line 33: can we validate that the image url is of the form we expect before doing this?

packages/pages/src/components/image/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Image } from "./image";
+export * from "./types";
```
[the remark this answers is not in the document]
EmilyZhang777: I tested that this only exports the types that are denoted with "export". If we want to have an internal type, we can just not export it in types.ts

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * A component that renders an image. Here is an example of using the component with simple and
+ * complex image fields from Knowledge Graph:
+ * ```
+ * import { Image } from "@yext/pages/components";
+ *
+ * const template = ({ document }) => {
+ *   return (
+ *     <div>
+ *       <Image image={document.logo.image} />
```
Mxs2019, line 14: Did we consider making the prop just `document.logo` to simplify things a bit?

EmilyZhang777: If we want to make this change, the `ImageType` will need to be `type ImageType = {image: {height: number...};};` with the extra `image` layer.

packages/pages/src/components/image/image.tsx
```diff
@@ -1,14 +1,34 @@
 import * as React from "react";
-import { useRef, useState, useEffect } from "react";
-import { ImageProps } from "./types";
-
+import { useEffect, useRef, useState } from "react";
+import { ImageProps, ImageLayout } from "./types";
+
+/**
+ * A component that renders an image. Here is an example of using the component with simple and
```
mkilpatrick, line 6: I'd rephrase this intro to:

Renders an image based from the Yext Knowledge Graph

packages/pages/src/components/image/image.tsx
mkilpatrick: I agree with Max that we should make it simpler to use.

packages/pages/src/components/image/image.tsx
```diff
@@ -1,14 +1,34 @@
...
+ * complex image fields from Knowledge Graph:
+ * ```
+ * import { Image } from "@yext/pages/components";
+ *
+ * const template = ({ document }) => {
```
mkilpatrick, line 11: I wouldn't give an example of the whole template, just using the component itself.

packages/pages/src/components/image/image.tsx
```diff
@@ -1,14 +1,34 @@
...
+ *   return (
+ *     <div>
+ *       <Image image={document.logo.image} />
+ *       <Image image={document.photoGallery[0].image} />
+ *     </div>
+ *   );
+ * }
+ * ```
+ *
+ * `layout`, `imgOverrides` and `style` can be used to specify the style and behavior of the image.
```
mkilpatrick, line 21: I think this is better left to documenting the prop type itself, rather than the component.

packages/pages/src/components/image/image.tsx
```diff
@@ -22,24 +42,78 @@ export const Image = ({
...
       "Width or height is passed in but layout is not fixed. These will have no impact. If you want to have a fixed height or width then set layout to fixed."
     );
   }
 
-  const imgWidth: number = image.image.width;
-  const imgHeight: number = image.image.height;
-  const imgUUID: string = image.image.url.split("/")[4];
+  const imgWidth: number = image.width;
+  const imgHeight: number = image.height;
+  const imgUrlValid = !!image.url && image.url.split("/").length > 4;
```
mkilpatrick, line 53: Turn this validation into its own function. I also think we should validate it more that just checking there are 4 slashes. We should also check our mktgcdn domain.

packages/pages/src/components/image/image.tsx
```diff
@@ -22,24 +42,78 @@ export const Image = ({
...
+        srcSet={srcSet}
+        loading={"lazy"}
+        onLoad={() => setImgLoaded(true)}
+        {...imgOverrides}
+      />
+    </>
+  );
+};
 
+const getImageUrl = (uuid: string, width: number, height: number) => {
```
mkilpatrick, line 94: Add jsdoc to these, even though they're internal

packages/pages/src/components/image/types.ts
```diff
@@ -7,22 +7,52 @@ export type ThumbnailType = {
 };
 
 export type ImageType = {
```
mkilpatrick, line 9: Sounds like we want the double image here

packages/pages/package.json
```diff
@@ -115,43 +113,32 @@
...
+      "<rootDir>/pages/src/**/*.test.{ts,tsx,js}"
     ],
     "transform": {
       "\\.[jt]sx?$": "ts-jest"
     },
     "moduleNameMapper": {
       "^(.+)\\.js$": "$1"
-    }
+    },
+    "testEnvironment": "jsdom"
```
mkilpatrick, line 142: I don't think we want to globally change this. The default test env is node. If you need jsdom for your specific test then you can override it for your particular test suite. See `packages/pages/src/util/runtime.test.ts`.

packages/pages/src/components/image/image.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+const image = {
+  image: {
+    width: imgWidth,
+    height: imgHeight,
+    url: `https://a.mktgcdn.com/p/${imgUUID}/2x1.jpg`,
+  },
+};
+
+/**
+ * @jest-environment node
```
mkilpatrick, line 21: Same comment. Why did you set the env to jsdom globally and then just swap it back here? Also, I'm pretty sure this needs to be at the top of the file to work.

packages/pages/src/components/image/image.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+ */
+
+describe("Image", () => {
+  it(`properly logs warning when layout is not ${ImageLayout.FIXED} and width or height is provided`, async () => {
+    jest.clearAllMocks();
```
mkilpatrick, line 26: Shouldn't be needed at the beginning if your other tests properly clear at the end

packages/pages/src/components/image/image.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+      image={image}
+      layout={ImageLayout.INTRINSIC}
+      width={100}
+    />);
+    expect(logMock.mock.calls.length).toBe(1);
+
+    jest.clearAllMocks();
+  });
+
+  it("properly renders the image with the pass through style and imgOverrides", () => {
```
mkilpatrick, line 40: Is there a way to test the lazy loading-ness of the image?

packages/pages/src/components/image/image.tsx
```diff
@@ -91,21 +79,43 @@ export const Image = ({
   );
 };
 
-const getImageUrl = (uuid: string, width: number, height: number) => {
+/**
+ * Returns the UUID of an image given its url. Logs a warning message if the image url is invalid.
+ */
+export const getImageUUID = (url: string) => {
```
mkilpatrick, line 85: Does this need to be exported?

packages/pages/src/components/image/image.tsx
```diff
@@ -91,21 +79,43 @@ export const Image = ({
...
+  const uuidRegex = /(?<=^https:\/\/a\.mktgcdn\.com\/p\/)[a-zA-Z0-9]+(?=\/(.*)$)/g;
```
mkilpatrick, line 86: Move this outside of the func and call it MKTGCDN_URL_REGEX

packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.test.tsx
```diff
@@ -16,12 +16,24 @@ const baseTemplateModule: TemplateModuleInternal<any, any> = {
 
 const baseProps: TemplateProps = {
   document: {},
-  __meta: { mode: "development" },
+  __meta: {
+    mode: "development",
+    manifest: {
```
mkilpatrick, line 21: Did this need to change?

packages/pages/src/components/image/image.test.tsx
EmilyZhang777: Didn't get this to work and forgot to remove it lol

packages/pages/src/components/image/image.tsx
EmilyZhang777: I exported them for testing but they are not exported in index.ts

packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.test.tsx
EmilyZhang777: yeah, need to change this to get the test ran properly

packages/pages/src/components/image/image.test.tsx
EmilyZhang777: I don't think so. The onLoad on the img tag is not invoked in test so there's no good way to test the loading option

packages/pages/package.json
```diff
@@ -90,7 +90,7 @@
     "@types/express-serve-static-core": "^4.17.29",
     "@types/fs-extra": "^9.0.13",
     "@types/glob": "^7.2.0",
-    "@types/jest": "^27.5.2",
+    "@types/jest": "^27.5.1",
```
mkilpatrick, line 93: You actually moved the version down which seems weird, though it doesn't actually matter.

EmilyZhang777: It was at 27.5.1 before. I thought. the check failure was caused by this so gave it a try. 27.5.1 should be fine to use

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,210 @@
...
+};
+
+/**
+ * Returns the UUID of an image given its url. Logs a warning message if the image url is invalid.
+ */
+export const getImageUUID = (url: string) => {
+  const matches = url.match(MKTGCDN_URL_REGEX);
+
+  if (matches == null || matches.length == 0) {
+    console.warn(`Invalid image url: ${url}.`);
```
jimboHenris, line 109: is this a recoverable state? If not, we should log a full error

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+      break;
+    case ImageLayout.FIXED:
+      if (!width && !height) {
+        console.warn(
+          "Using fixed layout but width and height are not passed as props."
+        );
+      }
+
+      const fixedWidth = width
```
jimboHenris, line 161: as a general rule, I try to avoid nested ternary operators, I use tao of react to guide most of my style / design decisions and this is something that's mentioned: https://alexkondov.com/tao-of-react/#avoid-nested-ternary-operators

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,210 @@
...
+    .join(", ");
+
+  const onError = () => {
+    setImgStatus(imgLoadingStatus.ERROR);
+    console.warn(`Invalid image src: ${src}.`);
+  };
+
+  return (
+    <>
+      {placeholder != null &&
```
jimboHenris, line 80: Just to be certain there's never an instance in which both placeholder and  could render I'd recommend just switching on imgLoadingStatus.LOADED. If it's loaded, , otherwise return placeholder. I don't think it'll be an issue if it's null, right? That's equivalent to what'd happen if we were in an error state w/o a placeholder anyway I think

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,210 @@
...
+      // Don't let image be wider than its intrinsic width
+      style.maxWidth = imgWidth;
+      style.width = "100%";
+      style.aspectRatio = aspectRatio
+        ? `${aspectRatio}`
+        : `${imgWidth} / ${imgHeight}`;
+
+      break;
+    case ImageLayout.FIXED:
+      if (!width && !height) {
```
jimboHenris, line 155: I'd move all this validation into a helper function, combine with the check at line 47 and call it there

packages/pages/src/components/image/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  width?: number;
+  /** The absolute height of the image. Only impacts if layout is set to "fixed". */
+  height?: number;
+  /** The aspect ratio of the image. Only impacts if layout is set to "aspect". */
+  aspectRatio?: number;
+  /** Specifies how the image is rendered. */
+  layout?: ImageLayout;
+  /** A pass through react component that is displayed when the image is loading. */
+  placeholder?: React.ReactNode;
+  /** Pass through props that are on the native HTML img tag. The Image component may not work if src and/or srcsets are included. */
```
jimboHenris, line 56: can we just add a check and if there's both ignore src and use srcset?

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,210 @@
...
+      const fixedHeight = height
+        ? height
+        : width
+        ? (width * imgHeight) / imgWidth
+        : imgHeight;
+
+      style.width = fixedWidth;
+      style.height = fixedHeight;
+
+      if (fixedWidth && fixedHeight) {
```
jimboHenris, line 175: what's the case where these both wouldn't be set here?

packages/pages/src/components/image/image.tsx
EmilyZhang777: Not sure if I get your point. But the main issue with this is, we want to make sure  is rendered in order to trigger onLoad, which sets the imgStatus. I verified that the  and placeholder never show up together

packages/pages/src/components/image/image.tsx
```diff
@@ -3,7 +3,12 @@ import { useEffect, useRef, useState } from "react";
 import { ImageProps, ImageLayout } from "./types";
 
 const MKTGCDN_URL_REGEX =
-  /(?<=^https:\/\/a\.mktgcdn\.com\/p\/)[a-zA-Z0-9]+(?=\/(.*)$)/g;
+  /((?<=^http:\/\/a\.mktgcdn\.com\/p\/)|(?<=^https:\/\/a\.mktgcdn\.com\/p\/)).+(?=\/(.*)$)/g;
+enum imgLoadingStatus {
```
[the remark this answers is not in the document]
mkilpatrick: Actually I don't think we need this at all. I'm going to put up a commit with some changes. TAL and lmk what you think.

packages/pages/src/components/image/image.tsx
mkilpatrick: Jim, I think the issue here is that the placeholder might not exist. We only display it if there is one.

packages/pages/src/components/image/types.ts
EmilyZhang777: The reason that this might not work is because we set the src and srcset on the ``. I don't think we reached a conclusion on which src has the higher priority. I'll leave it as is for now. If we want to add support for it, it can be a future item 👍

packages/pages/src/components/image/image.tsx
EmilyZhang777: I changed the behavior and forgot to remove this check

packages/pages/src/components/image/image.tsx
```diff
@@ -70,47 +73,89 @@ export const Image = ({
...
+  if (imgWidth < 0) {
+    console.warn(`Invalid image width: ${imgWidth}.`);
+  }
+
+  if (imgHeight < 0) {
+    console.warn(`Invalid image height: ${imgHeight}.`);
+  }
+
+  if (layout == ImageLayout.FIXED) {
+    if (!width && !height) {
```
mkilpatrick, line 115: Is height actually required? When I was testing earlier a fixed image with just a width seemed fine.

packages/pages/src/components/image/image.tsx
```diff
@@ -34,34 +28,43 @@ export const Image = ({
...
+  const imgWidth: number = Math.abs(image.image.width);
+  const imgHeight: number = Math.abs(image.image.height);
   const imgUUID = getImageUUID(image.image.url);
 
+  // The image is invalid, only try to load the placeholder
+  if (!imgUUID) {
+    return <>{placeholder != null && placeholder}</>;
+  }
+
+  const absWidth = width ? Math.abs(width) : undefined;
```
mkilpatrick, line 57: Why is this needed? Validation should handle it. It's confusing to validate something, then also fix it and use some other variable here.

packages/pages/src/components/image/image.tsx
```diff
@@ -70,47 +73,89 @@ export const Image = ({
...
-  };
-
   return (
     <>
-      {placeholder != null &&
-        imgStatus != imgLoadingStatus.LOADED &&
-        placeholder}
-      {imgStatus != imgLoadingStatus.ERROR && (
+      {!imgLoaded && placeholder != null && placeholder}
+      {src && (
```
mkilpatrick, line 79: I don't think the src check is necessary. It's always set by the time we get here.

packages/pages/src/components/image/image.test.tsx
```diff
@@ -125,101 +148,172 @@ describe("handleLayout", () => {
...
     );
 
-    expect(imgStyle.aspectRatio).toEqual(aspectRatio.toString());
+    expect(logMock.mock.calls.length).toBe(1);
+    jest.clearAllMocks();
   });
 
-  it(`properly sets width when layout is ${ImageLayout.FILL}`, () => {
-    const { imgStyle } = handleLayout(
+  it(`properly logs warning when image.width is a negative value`, () => {
```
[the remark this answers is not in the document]
EmilyZhang777: nah, this is the ImageProps.image.width, the other one is for ImageProps.width

packages/pages/src/components/image/image.tsx
```diff
@@ -132,8 +177,8 @@ export const handleLayout = (
   imgHeight: number,
   imgUUID: string,
   style: React.CSSProperties,
-  width?: number,
-  height?: number,
+  absWidth?: number,
+  absHeight?: number,
   aspectRatio?: number
 ): { src: string; imgStyle: React.CSSProperties; widths: number[] } => {
   let widths: number[] = [100, 320, 640, 960, 1280, 1920];
```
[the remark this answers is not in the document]
EmilyZhang777: It's not really mutating, this function takes in the style prop the user passed in and returns a new imgStyle that will be used on the ``

packages/pages/src/components/image/image.tsx
```diff
@@ -70,47 +73,89 @@ export const Image = ({
...
+export const validateRequiredProps = (
+  layout: ImageLayout,
+  imgWidth: number,
+  imgHeight: number,
+  width?: number,
+  height?: number,
+  aspectRatio?: number
+) => {
+  if (imgWidth < 0) {
+    console.warn(`Invalid image width: ${imgWidth}.`);
```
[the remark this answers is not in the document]
EmilyZhang777: This is not a deadly error, we can still render the image. All cases that log a warning now can still render an image. The only cases that we can't render an image are invalid url and invalid uuid, which are logged as errors.

packages/pages/src/components/image/image.tsx
EmilyZhang777: we need at least a width or a height to render

packages/pages/src/components/image/image.tsx
EmilyZhang777: hmm, the same logic is used throughout the file (where even if you have an invalid config, we still try to render an image to our best effort. For example, if you have layout = "aspect" but do not provide a ImageProps.aspectRatio, we still try to render an image with the `aspectRatio = ImageProps.image.width/ImageProps.image.height`.  Do we want to change that?

mkilpatrick: This is probably a call for Max. Idk if "try to render it as best we can" is good or if it leads to unknown outcomes. We can go with "try our best" for now, but can you start a thread about it (tomorrow is fine) and cc me?

packages/pages/src/components/image/image.test.tsx
```diff
@@ -66,16 +66,46 @@ describe("Image", () => {
...
         image={{
-          image: { ...image.image, url: "https://a.mktgcdn.com/p/2x1.jpg" },
+          image: { ...image.image, url: invalidUrl },
         }}
         placeholder={placeholder}
       />
     );
 
     expect(screen.getByText(placeholderText)).toBeTruthy();
+
```
mkilpatrick, line 84: remove empty line

packages/pages/src/components/image/image.tsx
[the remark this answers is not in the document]
mkilpatrick: Right, you return a new imgStyle, but you mutate style throughout first

packages/pages/src/components/image/image.tsx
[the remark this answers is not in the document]
mkilpatrick: Sounds good assuming we're trying to render something as best we can. Thanks for the explanation.

packages/pages/src/components/image/image.tsx
mkilpatrick: Oh yea, ha, misread. 😓

mkilpatrick (approved):

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,271 @@
...
+  width,
+  height,
+  aspectRatio,
+  layout = ImageLayout.INTRINSIC,
+  placeholder,
+  imgOverrides,
+  style = {},
+}: ImageProps) => {
+  const imgRef = useRef<HTMLImageElement>(null);
+  const [imgLoaded, setImgLoaded] = useState(false);
```
mancusi, line 31: isImageLoaded, setIsImageLoaded

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  useEffect(() => {
```
mancusi, line 33:
```js
const useIsImageLoaded = (): [React.Ref<HTMLImageElement>, boolean] => {
    const imgRef = useRef<HTMLImageElement>(null);
    const [imgLoaded, setImgLoaded] = useState(false);
    
    useEffect(() => {
       if (imgRef.current?.complete) {
      setImgLoaded(true);
    }
  }, []);

  return imgRef, isImageLoaded;
}
```

mancusi: then once this is a custom hook move it outside the view;

packages/pages/src/components/image/image.tsx
```diff
@@ -0,0 +1,271 @@
...
+      <img
+        ref={imgRef}
+        style={imgStyle}
+        src={src}
+        className={className}
+        width={absWidth}
+        height={absHeight}
+        srcSet={srcSet}
+        loading={"lazy"}
+        onLoad={() => setImgLoaded(true)}
```
mancusi, line 88: apparently load doesn't trigger when the image is loaded in chrome so I think just checking complete is fine.

mancusi (changes requested):

packages/pages/src/components/image/types.ts
```diff
@@ -0,0 +1,62 @@
...
+   * `objectPosition` props of the `style` rpop.
+   */
+  FIXED: "fixed",
+  /** Shows the image in a fixed aspect ratio. The `aspectRatio` prop must be provided. */
+  ASPECT: "aspect",
+  /** Always fills the image to 100% of the container's width. */
+  FILL: "fill",
+} as const;
+
+export type ImageLayout = typeof ImageLayout[keyof typeof ImageLayout];
```
mancusi, line 37: any exported value should have some sort of documentation on it. same as above

packages/pages/src/components/image/types.ts
```diff
@@ -0,0 +1,62 @@
...
+export type ImageType = {
+  image: {
+    height: number;
+    width: number;
+    url: string;
+    thumbnails?: ThumbnailType[];
+  };
+};
+
+export const ImageLayout = {
```
mancusi, line 18: this and the type have the same name which isn't ideal -- maybe this can be called ImageLayoutOption?

packages/pages/src/components/image/image.tsx
mancusi: nvm this isn't true (just tested)

packages/pages/src/components/image/image.tsx
mancusi: nvm we can't do this as nicely since we have to call the setter in the img component itself

packages/pages/src/components/image/image.tsx
EmilyZhang777: looks like Max wants to go with the current way

packages/pages/src/components/image/types.ts
```diff
@@ -34,29 +49,58 @@ export const ImageLayout = {
...
   /** A pass through react component that is displayed when the image is loading. */
   placeholder?: React.ReactNode;
   /** Pass through props that are on the native HTML img tag. The Image component may not work if src and/or srcsets are included. */
   imgOverrides?: Object;
   /** The pass through style of the underlying img tag. */
   style?: React.CSSProperties;
-};
+}
+
+interface OtherImageProps extends BaseImageProps {
```
mkilpatrick, line 81: jsdoc for all of these with @public

packages/pages/src/components/image/types.ts
```diff
@@ -34,29 +49,58 @@ export const ImageLayout = {
...
+interface AspectImageProps extends BaseImageProps {
+  /** Specifies how the image is rendered. */
+  layout: "aspect";
+  /** The aspect ratio of the image. Only impacts if layout is set to "aspect". */
+  aspectRatio: number;
+}
+
+/**
+ * The shape of the data passed directly to the different template functions with the
+ * exception of the render function (getPath, getHeadConfig, etc).
```
mkilpatrick, line 104: This doc talks about potential usage rather than what it _is_. These are the props required by the Image componet.

packages/pages/src/components/image/types.ts
EmilyZhang777: I don't think we want to export these?

EmilyZhang777: Added docs but didn't export the layout specific interfaces. Let me know if you want to export them.

jimboHenris:
Aside from the failing GH actions, LGTM

mancusi (approved):
forgot to approve this earlier 👍

mkilpatrick (approved):

PR #147, opened 2022-08-03
feat(plugin): validate file sizes (closes #145)

### LINKED ISSUE #145: Show File Size Limit in Local Dev
### Describe the bug

There is a `1mb` file size limit in the function system that generates pages. This limit spits out an error like
```
Error encountered in site configuration: source file assets/server/locator.6a26f199.js for plugin Generator-ry2nmj5shd exceeds max size of 1 MB
```

This error should be caught in local development and should throw an error. It would also be helpful if the error message was more helpful and pointed out how to fix the error. 

### Reproduction

_No response_

### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
    Memory: 538.21 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.17.0 - /usr/local/bin/node
    Yarn: 1.22.11 - /usr/local/bin/yarn
    npm: 8.1.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 98.0.4758.102
    Firefox: 88.0.1
    Safari: 15.6
```

### Used Package Manager

npm

### Used Browser

Chrome

### Validations

mkilpatrick:


packages/pages/src/vite-plugin/build/closeBundle/moduleLoader.ts
```diff
@@ -62,6 +67,40 @@ const validateUniqueFeatureName = (
     });
 };
 
+const validateBundles = (serverBundlePaths: string[]) => {
```
mancusi, line 70: The server bundle paths aren't the only assets included in the plugin. Everything in the static and renderer folder are also included (see the ci.json file in the starter or the artifact output )

mancusi (changes requested):
can you add tests?

packages/pages/src/vite-plugin/build/closeBundle/bundleValidator.ts
```diff
@@ -0,0 +1,51 @@
...
+import glob from "glob";
+import path from "path";
+
+const PLUGIN_FILESIZE_LIMIT = 1.5; // MB
+const PLUGIN_TOTAL_FILESIZE_LIMIT = 10; // MB
+
+/**
+ * Validates the bundled files.
+ */
+export const validateBundles = () => {
```
mancusi, line 11: this isn't all the bundles -- just the server-side bundles. should that be noted?

packages/pages/src/vite-plugin/build/closeBundle/bundleValidator.test.ts
```diff
@@ -0,0 +1,96 @@
+import { glob } from "glob";
+import fs, { Stats } from "fs";
+import { validateBundles } from "./bundleValidator";
+
+describe("bundleValidator", () => {
+  it("throws an error when a single file is over the max filesize limit", async () => {
+    const stats = getStats(1);
+
+    jest.spyOn(glob, "sync").mockImplementation(() => ["file1.js", "file2.js"]);
```
mancusi, line 9: make sure to restoreAllMocks
https://jestjs.io/docs/jest-object#jestspyonobject-methodname

mancusi (changes requested):

mancusi:
General comment, the fact that we're doing this validation now ties the ci.json configuration in the repo to versions of @yext/pages. This doesn't particularly desirable. I think we should start auto-generating this plugin configuration

jimboHenris:
> General comment, the fact that we're doing this validation now ties the ci.json configuration in the repo to versions of @yext/pages. This doesn't particularly desirable. I think we should start auto-generating this plugin configuration

Sorry, I'm not sure I follow this. Which plugin configuration would we autogen? And where is the coupling?

mkilpatrick:
> General comment, the fact that we're doing this validation now ties the ci.json configuration in the repo to versions of @yext/pages. This doesn't particularly desirable. I think we should start auto-generating this plugin configuration

Agreed. We talked offline and this will be desirable to do in the future, but we already have this loose coupling now so this CR shouldn't impact or make anything worse.

> Sorry, I'm not sure I follow this. Which plugin configuration would we autogen? And where is the coupling?

We'd autogen the `artifactStructure` portion of ci.json (except it'll need to live in its own file. The coupling now is that PagesJS assumes the dist/assets folders (hard coded) and that setup is defined in the starter's ci.json. If we change something on either side things won't work.

jimboHenris (approved):

mancusi (approved):
LGTM if we're not going to do the medium-term solution here.

PR #148, opened 2022-08-05
feat(dev): redirect if url ends in final slash (closes #143)

### LINKED ISSUE #143: Redirect urls with trailing slashes in dev server
### Description

Currently it's possible to visit urls with or without a trailing slash. For example an entity page can be accessed at both `http://localhost:5173/location/123` and `http://localhost:5173/location/123/`.

This is confusing because of how it interacts with the `relativePrefixToRoot`. A link like: ```<a href={`${relativePrefixToRoot}path`}>Link</a>``` 
will take you to `http://localhost:5173/location/path` if you're viewing the page with a trailing slash in the url and `http://localhost:5173/path` if you're viewing the page without a trailing slash. 

 

### Suggested solution

Redirect urls with trailing slashes in the dev server to a version without the trailing slash. That would make it closer to how it operates in production. 

### Alternative

The logic used to calculate `relativePrefixToRoot` could just ignore the trailing slash, but seems nicer to remove entirely. 

### Additional context

_No response_

### Validations

- [X] Check that there isn't already an issue that requests the same feature to avoid creating a duplicate.

mkilpatrick:
If the local dev url ends in a slash, 301 redirects to the canonical url without the slash. This will prevent future issues when people use relative links.

### DIFF
--- a/packages/pages/src/dev/server/middleware/finalSlashRedirect.ts
+++ b/packages/pages/src/dev/server/middleware/finalSlashRedirect.ts
@@ -2,11 +2,11 @@
  * Redirect urls with a final slash to their canonical url without the slash.
  */
 export const finalSlashRedirect = (req: any, res: any, next: any): void => {
-    if (req.path.endsWith('/') && req.path.length > 1) {
-        const query = req.url.slice(req.path.length);
-        const safepath = req.path.slice(0, -1).replace(/\/+/g, '/');
-        res.redirect(301, safepath + query);
-      } else {
-        next()
-      }
-}
+  if (req.path.endsWith("/") && req.path.length > 1) {
+    const query = req.url.slice(req.path.length);
+    const safepath = req.path.slice(0, -1).replace(/\/+/g, "/");
+    res.redirect(301, safepath + query);
+  } else {
+    next();
+  }
+};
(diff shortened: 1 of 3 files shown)


packages/pages/src/dev/server/server.ts
```diff
@@ -37,6 +38,9 @@ export const createServer = async (dynamicGenerateData: boolean) => {
     displayGenerateTestDataWarning = !(await generateTestData());
   }
 
+  // Redirect urls with a final slash to their canonical url without the slash
+  // app.use(finalSlashRedirect);
```
jimboHenris: uncomment ?

packages/pages/src/dev/server/middleware/finalSlashRedirect.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
mancusi, line 1: add tests

mancusi (changes requested):

packages/pages/src/dev/server/server.ts
mkilpatrick: I think you were looking at an old commit. This is not commented in the latest commit.

jimboHenris (approved):

mancusi (approved):

PR #151, opened 2022-08-05
feat(components): add AnalyticsProvider

benmcginnis:
Adds new AnalyticsProvider and AnalyticsScope components for sending events to the Yext Analytics system.

### DIFF
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -60,6 +60,7 @@
     "THIRD-PARTY-NOTICES"
   ],
   "dependencies": {
+    "@yext/analytics": "^0.2.0",
     "ansi-to-html": "^0.7.2",
     "chalk": "^5.0.1",
     "classnames": "2.3.1",
@@ -97,6 +98,7 @@
     "@storybook/react": "^6.5.10",
     "@storybook/testing-library": "^0.0.13",
     "@testing-library/react": "^12.1.5",
+    "@testing-library/user-event": "^14.4.2",
     "@types/escape-html": "^1.0.2",
     "@types/express": "^4.17.13",
     "@types/express-serve-static-core": "^4.17.30",
--- a/packages/pages/src/components/analytics/Analytics.ts
+++ b/packages/pages/src/components/analytics/Analytics.ts
@@ -1,0 +1,238 @@
+import { TemplateProps } from "../../common/src/template/types";
+import { getRuntime, isProduction } from "../../util";
+import { AnalyticsMethods } from "./interfaces";
+import {
+  ConversionDetails,
+  CookieManager,
+  EntityPage,
+  PagesAnalyticsService,
+  providePagesAnalytics,
+  StaticPage,
+  Visitor,
+} from "@yext/analytics";
+import { concatScopes, slugify } from "./helpers";
+
+/**
+ * The Analytics class creates a stateful facade in front of the \@yext/analytics
+ * Library's pagesAnalyticsProvider class. It takes in some data from the
+ * template configuration and uses it to provide configuration to the
+ * pagesAnalyticsProvider.
+ *
+ * Additionally, it provides handlers for controlling user opt-in for compliance
+ * requirements as well as for debugging, enabling conversion tracking, saving
+ * user identity information, and creating named analytics scopes for
+ * easy tagging.
+ *
+ * @public
+ */
+export class Analytics implements AnalyticsMethods {
+  private _optedIn: boolean;
+  private _conversionTrackingEnabled = false;
+  private _cookieManager: CookieManager | undefined;
+  private _analyticsReporter: PagesAnalyticsService | undefined;
+  private _pageViewFired = false;
+  private _enableDebugging = false;
+
+  /**
+   * Creates an Analytics instance, will fire a pageview event if requireOptin
+   * is false
+   *
+   * @param templateData - template data object from the pages system
+   * @param requireOptIn - boolean, set to true if you require user opt in before tracking analytics
+   */
+  constructor(
+    private templateData: TemplateProps,
+    requireOptIn?: boolean | undefined
+  ) {
+    this._optedIn = !requireOptIn;
+    this.makeReporter();
+    this.pageView();
+  }
+
+  private calculatePageType(): EntityPage | StaticPage {
+    const isStaticPage = !!this.templateData.document?.__?.staticPage;
+    const isEntityPage = !!this.templateData.document?.__?.entityPageSet;
+
+    let pageType: EntityPage | StaticPage;
+
+    if (isStaticPage) {
+      pageType = {
+        name: "static",
+        staticPageId: this.templateData.document.__.name as string,
+      };
+    } else if (isEntityPage) {
+      pageType = {
+        name: "entity",
+        pageSetId: this.templateData.document.__.name as string,
+        id: this.templateData.document.uid as number,
+      };
+    } else {
+      throw new Error("invalid document type");
+    }
+
+    return pageType;
+  }
+
+  private makeReporter() {
+    if (getRuntime().name !== "browser") {
+      return;
+    }
+    if (!this._optedIn) {
+      return;
+    }
+
+    const inProduction =
+      isProduction(this.templateData?.document?.siteInternalHostName) ||
+      isProduction(this.templateData?.document?.siteDomain);
+
+    this._analyticsReporter = providePagesAnalytics({
+      businessId: this.templateData.document.businessId as number,
+      pageType: this.calculatePageType(),
+      pageUrl: window.location.href,
+      production: inProduction,
+      referrer: document.referrer,
+      siteId: this.templateData.document.siteId as number,
+    });
+
+    this.setDebugEnabled(this._enableDebugging);
+  }
+
+  private canTrack(): boolean {
+    return (
+      getRuntime().name === "browser" &&
+      this._optedIn &&
+      !!this._analyticsReporter
+    );
+  }
+
+  private setupConversionTracking(): void {
+    this._cookieManager = new CookieManager();
+    this._analyticsReporter?.setConversionTrackingEnabled(
+      true,
+      this._cookieManager.setAndGetYextCookie()
+    );
+  }
+
+  /** {@inheritDoc AnalyticsMethods.enableConversionTracking} */
+  enableTrackingCookie(): void {
+    this._conversionTrackingEnabled = true;
+
+    if (this.canTrack()) {
+      this.setupConversionTracking();
+    }
+  }
+
+  /** {@inheritDoc AnalyticsMethods.identify} */
+  identify(visitor: Visitor): void {
+    if (this.canTrack()) {
+      this._analyticsReporter?.setVisitor(visitor);
+    }
+  }
+
+  /** {@inheritDoc AnalyticsMethods.async} */
+  async optIn(): Promise<void> {
+    this._optedIn = true;
+    this.makeReporter();
+    if (this._conversionTrackingEnabled && !this._cookieManager) {
+      this.setupConversionTracking();
+    }
+    if (!this._pageViewFired) {
+      await this.pageView();
+    }
+  }
+
+  /** {@inheritDoc AnalyticsMethods.async} */
+  async pageView(): Promise<void> {
+    if (!this.canTrack()) {
+      return Promise.resolve(undefined);
+    }
+    // TODO: if this successfully completes & conversion tracking is enabled
+    // and the user is opted in we should remove the y_source query parameter
+    // from the url if it is present to prevent double counting a listings click
+    // if the page is refreshed.
+    await this._analyticsReporter?.pageView();
+    this._pageViewFired = true;
+  }
+
+  /** {@inheritDoc AnalyticsMethods.track} */
+  async track(
+    eventName: string,
+    conversionData?: ConversionDetails
+  ): Promise<void> {
+    if (!this.canTrack()) {
+      return Promise.resolve();
+    }
+
+    await this._analyticsReporter?.track(
+      { eventType: slugify(eventName) },
+      conversionData
+    );
+  }
+
+  /** {@inheritDoc AnalyticsMethods.setDebugEnabled} */
+  setDebugEnabled(enabled: boolean): void {
+    this._enableDebugging = enabled;
+    this._analyticsReporter?.setDebugEnabled(enabled);
+  }
+
+  /** {@inheritDoc AnalyticsMethods.trackClick} */
+  trackClick(
+    eventName: string,
+    conversionData?: ConversionDetails
+  ): (e: MouseEvent) => Promise<void> {
+    return (e: MouseEvent) => {
+      if (!this.canTrack()) {
+        return Promise.resolve();
+      }
+
+      if (e.target === null || e.defaultPrevented) {
+        return this.track(eventName, conversionData);
+      }
+
+      const targetLink = e.target as HTMLAnchorElement;
+
+      if (targetLink.href === null || targetLink.href === undefined) {
+        return this.track(eventName, conversionData);
+      }
+
+      const linkUrl = new URL(targetLink.href);
+
+      if (
+        linkUrl.protocol === "mailto:" ||
+        linkUrl.protocol === "tel:" ||
+        linkUrl.protocol === "javascript:" ||
+        linkUrl.hostname === window.location.hostname
+      ) {
+        return this.track(eventName, conversionData);
+      }
+
+      const targetBlankOrSimilar =
+        (targetLink.target &&
+          !targetLink.target.match(/^_(self|parent|top)$/i)) ||
+        e.ctrlKey ||
+        e.shiftKey ||
+        e.metaKey;
+
+      if (targetBlankOrSimilar) {
+        return this.track(eventName, conversionData);
+      }
+
+      e.preventDefault();
+
+      const navigate = () => {
+        window.location.href = linkUrl.toString();
+      };
+
+      const awaitTimeout = new Promise<void>((resolve) => {
+        setTimeout(() => {
+          resolve();
+        }, 1000);
+      });
+
+      return Promise.race([
+        this.track(eventName, conversionData),
+        awaitTimeout,
+      ]).then(navigate);
+    };
+  }
+}
--- a/packages/pages/src/components/analytics/analytics.test.tsx
+++ b/packages/pages/src/components/analytics/analytics.test.tsx
@@ -1,0 +1,202 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from "react";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { TemplateProps } from "../../common/src/template/types";
+import { AnalyticsProvider } from "./provider";
+import { useAnalytics, useTrack } from "./hooks";
+import { AnalyticsScopeProvider } from "./scope";
+
+global.fetch = jest.fn().mockImplementation(
+  jest.fn(() => {
+    return Promise.resolve({ status: 200 });
+  }) as jest.Mock
+);
+
+jest.spyOn(global, "fetch");
+
+const baseProps: TemplateProps = {
+  document: {
+    __: {
+      entityPageSet: {},
+      name: "foo",
+    },
+    uid: 0,
+    businessId: 0,
+    siteId: 0,
+  },
+  __meta: { mode: "development" },
+};
+
+const currentProcess = global.process;
+
+beforeAll(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  global.process = undefined;
+});
+
+afterEach(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  global.fetch.mockClear();
+});
+
+afterAll(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  delete global.fetch;
+  global.process = currentProcess;
+});
+
+describe("Analytics", () => {
+  it("should fire a page view once", () => {
+    const App = () => {
+      return (
+        <AnalyticsProvider templateData={baseProps} requireOptIn={false} />
+      );
+    };
+    const { rerender } = render(<App />);
+    rerender(<App />);
+    expect(global.fetch).toHaveBeenCalledTimes(1);
+  });
+
+  it("should not fire a page view when opt in is required", () => {
+    render(<AnalyticsProvider templateData={baseProps} requireOptIn={true} />);
+    expect(global.fetch).toHaveBeenCalledTimes(0);
+  });
+
+  it("should track a click", () => {
+    const MyButton = () => {
+      const track = useTrack();
+      return <button onClick={async () => await track("foo click")} />;
+    };
+
+    render(
+      <AnalyticsProvider templateData={baseProps} requireOptIn={false}>
+        <MyButton />
+      </AnalyticsProvider>
+    );
+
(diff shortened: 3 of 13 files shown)


package.json
```diff
@@ -47,5 +47,8 @@
     "semver": "^7.3.7",
     "tsx": "^3.8.0"
   },
+  "dependencies": {
```
mkilpatrick, line 50: remove

packages/pages/src/components/analytics/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from "./interfaces";
```
mkilpatrick, line 1: Do you actually need to export all of these things or can just export the component?

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -0,0 +1,38 @@
+import { useContext } from "react";
+import { AnalyticsContext } from "./context";
+import { AnalyticsMethods } from "./interfaces";
+
+/**
+ * Returns a service that can be used to report analytics events
+ */
```
mkilpatrick, line 7: Anything that a user will use should have a line in the jsdoc that's `@public`

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+export function useAnalytics(): AnalyticsMethods|null {
+  return useContext(AnalyticsContext);
+}
+
+/**
+ * Returns the analytics track() method
```
mkilpatrick, line 13: Add periods to all of these. Also, they're not all that informative.

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,125 @@
...
+    return str
+      .toLowerCase()
+      .trim()
+      .replace(/[^\w\s-]/g, '')
+      .replace(/[\s_-]+/g, '')
+      .replace(/^-+|-+$/g, '');
+  }
+
+  async track(eventName: string, conversionData?: ConversionDetails): Promise<void> {
+    if(!this.canTrack()) return Promise.resolve(undefined);
```
mkilpatrick, line 119: Use braces, same throughout

packages/pages/src/index.ts
```diff
@@ -2,3 +2,5 @@ export * from "./dev/dev";
 export * from "./common/src/template/head";
 export * from "./common/src/template/paths";
 export * from "./common/src/template/types";
+export * from "./util";
```
mkilpatrick, line 5: These shouldn't be necessary

packages/pages/src/components/analytics/Analytics.ts
benmcginnis: probably should add that to the eslint config, if it's a style standard.

package.json
benmcginnis: are you just pointing out the file dependency? bc analytics is going to need to be bundled with the library is it not?

packages/pages/src/components/analytics/index.ts
benmcginnis: I'm not 100% sure on where we should make the package boundaries. I could see use cases for exposing these.  I was also somewhat following slapshot's conventions for testing but their tests are setup differently.

packages/pages/src/components/analytics/hooks.ts
benmcginnis: yeah i was going to do a documentation pass once things had settled a bit

packages/pages/src/index.ts
benmcginnis: I was having trouble using the runTime stuff without this.  Do you want people to import it from a sub-package?

packages/pages/package.json
```diff
@@ -82,9 +83,13 @@
...
   },
   "devDependencies": {
+    "@babel/preset-env": "^7.16.11",
+    "@babel/preset-typescript": "^7.16.7",
     "@microsoft/api-documenter": "^7.19.2",
     "@microsoft/api-extractor": "^7.29.0",
     "@testing-library/react": "^12.1.5",
+    "@testing-library/user-event": "^14.4.2",
+    "@types/babel__core": "^7.1.18",
     "@types/escape-html": "^1.0.2",
```
benmcginnis, line 93: I did not add these babel dependencies myself. They came in when I rebased. I assume I should remove @mkilpatrick ?

packages/pages/src/index.ts
benmcginnis: specifically when I was trying to use it directly in a test site.  Do you want the usage to be: 
``` js
import { getRuntime} from "@yext/pages/util"
```
as opposed to:
``` js
import { getRuntime} from "@yext/pages"
```

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,192 @@
...
+  private _analyticsReporter: PagesAnalyticsService | undefined;
+  private _pageViewFired = false;
+  private _enableDebugging = false;
+  constructor(
+    private templateData: TemplateProps,
+    requireOptIn?: boolean | undefined
+  ) {
+    this._optedIn = !requireOptIn;
+    this.makeReporter();
+    this.pageView(); // TODO: should I await this?
```
mkilpatrick, line 42: You can't. Not sure you'll be able to do this.

https://stackoverflow.com/questions/43431550/async-await-class-constructor

packages/pages/src/components/analytics/analytics.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+beforeAll(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  global.process = undefined;
+});
+
+beforeEach(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  global.fetch.mockClear();
```
mkilpatrick, line 44: I think you want to clear the mock after each instead of before, otherwise another test suite could be affected.

packages/pages/src/components/analytics/analytics.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+  delete global.fetch;
+  global.process = currentProcess;
+});
+
+describe("Analytics", () => {
+  it("should fire a page view", () => {
+    render(<AnalyticsProvider templateData={baseProps} requireOptIn={false} />);
+    expect(global.fetch).toHaveBeenCalledTimes(1);
+  });
+  it("should not fire a page view when opt in is required", () => {
```
mkilpatrick, line 59: Can you add empty lines between the tests?

packages/pages/src/components/analytics/analytics.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+      expect(generatedConversionUrl.searchParams.get("_yfpc")).toBeTruthy();
+      expect(generatedConversionUrl.searchParams.get("cid")).toBe(
+        expectedConversionData.cid
+      );
+      expect(generatedConversionUrl.searchParams.get("cv")).toBe(
+        expectedConversionData.cv
+      );
+    });
+  });
+  // TODO: figure out the right way to test window.location logic to credit listings with y_source param
```
mkilpatrick, line 149: This PR?

packages/pages/src/components/analytics/index.ts
mkilpatrick: I would only export the specific things you'd want a user to be able to import.

packages/pages/src/components/analytics/interfaces.ts
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * The track method will send a generic analytics event to Yext.
+   *
+   * @param eventName - the name of the event, will appear in Yext's Report Builder UI
+   * @param conversionData - optional details for tracking an event as a conversion
+   */
+  track(eventName: string, conversionData?: ConversionDetails): Promise<void>;
+
+  /**
+   * The identify method will allow you to tie analytics events to a specific user
```
mkilpatrick, line 20: period

packages/pages/src/components/analytics/provider.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from "react";
+import { PropsWithChildren } from "react";
+import { getRuntime } from "../../util";
+import { Analytics } from "./Analytics";
+import { AnalyticsProviderProps } from "./interfaces";
+import { AnalyticsContext } from "./context";
+
+/**
+ * The main Analytics component for you to use. Sets up the proper react context
+ * and bootstraps the Analytics reporter
```
mkilpatrick, line 10: period

packages/pages/src/components/analytics/provider.tsx
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @param props - A PropsWithChildren that implements AnalyticsProviderProps/
```
mkilpatrick, line 12: what's the slash for?

packages/pages/src/index.ts
mkilpatrick: The package.json types and exports defines how to use the import. This index.ts is the entrypoint to the pagesjs dev server.

packages/pages/CHANGELOG.md
```diff
@@ -1,25 +1,21 @@
 # 1.0.0-beta.1 (2022-08-05)
 
-
 ### Bug Fixes
 
-* **dev:** honor feature name casing in local dev url (fixes [#121](https://github.com/yext/sites-scripts/issues/121)) ([#146](https://github.com/yext/sites-scripts/issues/146)) ([e6fb9f2](https://github.com/yext/sites-scripts/commit/e6fb9f2f759d9e93101512e2270053a0157beba8))
-* improve errors related to undefined getPath (fixes [#130](https://github.com/yext/sites-scripts/issues/130)) ([#138](https://github.com/yext/sites-scripts/issues/138)) ([37b33d0](https://github.com/yext/sites-scripts/commit/37b33d00f75b8acb7e232d2935c173f98347cb21))
-* index page incorrectly saying there is no localData (fix [#135](https://github.com/yext/sites-scripts/issues/135)) ([#136](https://github.com/yext/sites-scripts/issues/136)) ([dd272d2](https://github.com/yext/sites-scripts/commit/dd272d2977ac76bf416965382b44fcc199c2c869))
-* update bundler plugin copying ([#111](https://github.com/yext/sites-scripts/issues/111)) ([c602104](https://github.com/yext/sites-scripts/commit/c602104d8f921b47842c2a245fa4c89b571caf80)), closes [#109](https://github.com/yext/sites-scripts/issues/109) [#109](https://github.com/yext/sites-scripts/issues/109)
-
```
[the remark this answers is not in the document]
benmcginnis: I believe so, yes. I did not change it myself.

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,192 @@
...
+  CookieManager,
+  EntityPage,
+  PagesAnalyticsService,
+  providePagesAnalytics,
+  StaticPage,
+  Visitor,
+} from "@yext/analytics";
+
+/**
+ * The Analytics class creates a stateful facade in front of the \@yext/analytics
```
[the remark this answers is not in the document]
benmcginnis: That actually needs to be there or the docs command will fail because it will try to treat it as a release tag

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,192 @@
...
+ * easy tagging.
+ *
+ * @public
+ */
+export class Analytics implements AnalyticsMethods {
+  private _optedIn: boolean;
+  private _conversionTrackingEnabled = false;
+  private _scopes: string[] = [];
+  private _cookieManager: CookieManager | undefined;
+  // TODO: create event queue for things to fire when optIn is called if we are not optedIn
```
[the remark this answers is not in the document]
benmcginnis: I don't want to block on doing this, I'm also not sure if that's strictly the correct thing to do. Need to ask product. It would lead to the time stamps being wrong, I guess. I've already set it up to track a pageView properly on opt in, not sure if the rest is necessary or not.

packages/pages/src/components/analytics/Analytics.ts
benmcginnis: Hmm I do love the builder pattern for constructing JS classes, I can look at that in the future. It could also significantly clean up some of the configuration mess with inspecting the stream document.

packages/pages/src/components/analytics/analytics.test.tsx
benmcginnis: I can adjust and see if that works properly. Most of my research on the topic suggested this way.

packages/pages/src/components/analytics/analytics.test.tsx
benmcginnis: Sure! can you add to eslint?

packages/pages/src/components/analytics/analytics.test.tsx
benmcginnis: I did a lot of research trying to figure out the correct way of doing this but I don't want to block on it. I'm pretty sure it's redundant bc it's already tested for in the analytics library itself. Just trying to be thorough.

I don't think it needs to be part of this PR.

packages/pages/src/components/analytics/interfaces.ts
benmcginnis: Eslint

packages/pages/src/components/analytics/provider.tsx
benmcginnis: To keep you on your toes

packages/pages/src/components/analytics/index.ts
benmcginnis: ok, i've removed the interfaces.

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -0,0 +1,102 @@
...
+  return useAnalytics().identify;
+};
+
+/**
+ * Simpler hook that just returns the analytics pushScope method
+ *
+ * @public
+ */
+export const useScope = () => {
+  return useAnalytics().pushScope;
```
bhainesva, line 101: maybe it makes more sense to have a `setScope` than `pushScope`? I think the imperative way pushscope works gets confusing in the react components. does it push every time the component rerenders?

packages/pages/package.json
```diff
@@ -58,6 +58,7 @@
     "THIRD-PARTY-NOTICES"
   ],
   "dependencies": {
+    "@yext/analytics": "^0.2.0",
```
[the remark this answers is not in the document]
benmcginnis: my intention was for bug fixes to apply automatically.  @mkilpatrick  @tmeyer2115  what do you think about pinning to a specific version or accepting bug fixes here?

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,254 @@
...
+    if (this.canTrack()) {
+      this._analyticsReporter?.setVisitor(visitor);
+    }
+  }
+
+  /** {@inheritDoc AnalyticsMethods.async} */
+  async optIn(): Promise<void> {
+    this._optedIn = true;
+    this.makeReporter();
+    this.setDebugEnabled(this._enableDebugging);
```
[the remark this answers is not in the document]
benmcginnis: I can move this into makeReporter

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,254 @@
...
+        return this.track(eventName, conversionData);
+      }
+
+      const targetBlankOrSimilar =
+        (targetLink.target &&
+          !targetLink.target.match(/^_(self|parent|top)$/i)) ||
+        e.ctrlKey ||
+        e.shiftKey ||
+        e.metaKey ||
+        e.which === 2 ||
```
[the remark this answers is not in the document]
benmcginnis: I pulled this from classic page's logic but it's here for legacy browser support for handling right clicks. I can remove since we don't support those browsers anymore and the whole site probably won't work on them

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,254 @@
...
+  private calculatePageType(): EntityPage | StaticPage {
+    const isStaticPage = !!this.templateData.document?.__?.staticPage;
+    const isEntityPage = !!this.templateData.document?.__?.entityPageSet;
+
+    let pageType: EntityPage | StaticPage;
+
+    if (isStaticPage) {
+      pageType = {
+        name: "static",
+        staticPageId: this.templateData.document.__.name as string,
```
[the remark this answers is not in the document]
benmcginnis: @mkilpatrick I didn't want to modify code elsewhere in the repo and it seems like this would be a fairly size-able change?

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -0,0 +1,102 @@
...
+  // @ts-ignore
+  if (!window.setAnalyticsOptIn) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    window.setAnalyticsOptIn = () => {
+      ctx.optIn();
+    };
+  }
+
+  // TODO: this is ugly, I imagine there is a more elegant way of doing this
```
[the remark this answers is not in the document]
benmcginnis: I'm sure it does. I was more thinking that there may be a better, more react best practices approach to this that we should research for a future version

packages/pages/src/components/analytics/hooks.ts
benmcginnis: that is a good question that i need to test

packages/pages/src/components/analytics/Analytics.ts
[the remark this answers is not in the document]
mkilpatrick: 👍

packages/pages/src/components/analytics/Analytics.ts
benmcginnis: technically we don't actually need to await it. i was just wondering if it would be better to

packages/pages/src/components/analytics/Analytics.ts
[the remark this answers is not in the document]
benmcginnis: I think for now we can leave this as is and work on specifying the types later? This seems like setting it up would end up changing a lot of the rest of the library

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -27,8 +27,8 @@ export function useAnalytics(): AnalyticsMethods {
 
   // TODO: is this the right way / place to expose a callback for use by a Cookie Management banner?
```
mkilpatrick, line 28: Do you want to remove this TODO now?

mkilpatrick (approved):

packages/pages/src/components/analytics/hooks.ts
benmcginnis: I mean I am still unclear if this is the best way. Product also mentioned potentially checking for a cookie as a future thing they may want to do. 

What is your philosophy on TODOs? I do not consider them merge blockers generally.  I can also create a corresponding issue in Github to keep track, rather than just leaving them in the code, if you'd prefer.

benmcginnis: Spoke offline, going to resolve this conversation and leave the TODO in place.

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,231 @@
...
+  private _analyticsReporter: PagesAnalyticsService | undefined;
+  private _pageViewFired = false;
+  private _enableDebugging = false;
+  constructor(
+    private templateData: TemplateProps,
+    requireOptIn?: boolean | undefined
+  ) {
+    this._optedIn = !requireOptIn;
+    this.makeReporter();
+    this.pageView();
```
bhainesva, line 42: Are there ever cases when someone might want to construct multiple Analytics instances on a page and the automatic pageview might cause problems?

packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,231 @@
...
+  private makeReporter() {
+    if (getRuntime().name !== "browser") {
+      return;
+    }
+    if (!this._optedIn) {
+      return;
+    }
+
+    const inProduction =
+      isProduction(this.templateData?.document?.siteInternalHostName) ||
```
bhainesva, line 78: Is the `siteInternalHostName` only populated in prod?

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  const scope = useScope();
+
+  // TODO: this is ugly, I imagine there is a more elegant way of doing this
+  return {
+    trackClick(
+      eventName: string,
+      conversionData?: ConversionDetails
+    ): (e: MouseEvent) => Promise<void> {
+      return ctx.trackClick(eventName, conversionData);
```
bhainesva, line 45: think we need to concatScopes here

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -0,0 +1,96 @@
...
+  };
+}
+
+/**
+ * Simpler hook that just returns the analytics track() method.
+ *
+ * @public
+ */
+export const useTrack = () => {
+  return useAnalytics().track;
```
bhainesva, line 77: need useScope here

packages/pages/src/components/analytics/analytics.test.tsx
```diff
@@ -0,0 +1,202 @@
...
+    siteId: 0,
+  },
+  __meta: { mode: "development" },
+};
+
+const currentProcess = global.process;
+
+beforeAll(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
bhainesva, line 37: do we need these if we declare them ahead of time?

bhainesva (approved):

packages/pages/src/components/analytics/hooks.ts
bhainesva: nvm

packages/pages/src/components/analytics/Analytics.ts
benmcginnis: the isProduction method returns true if window.location.href matches the domain name you pass in and you're in the browser environment

packages/pages/src/components/analytics/hooks.ts
benmcginnis: do we? trackClick calls track under the hood and the concatScopes call is there

benmcginnis: oh yeah you're right, nevermind

packages/pages/src/components/analytics/helpers.ts
```diff
@@ -0,0 +1,19 @@
...
+      .trim()
+      // strip special characters
+      .replace(/[^\w\s-]|[\s-]+/g, "")
+  );
+}
+
+export function concatScopes(a: string, b: string): string {
+  const aSlug = slugify(a);
+  const bSlug = slugify(b);
+  if (aSlug.length == 0) {
```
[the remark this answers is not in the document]
benmcginnis: ack

packages/pages/src/components/analytics/analytics.test.tsx
benmcginnis: typescript _hates_ mucking with the globals in a manner that's needed for this to work.  I can try to declare but then I get several other typescript errors. I personally don't think it's worth it to try to make typescript happy with this.

bhainesva (approved):

mkilpatrick (approved):

mkilpatrick (approved):

PR #158, opened 2022-08-09
feat(components): add Link component

bhainesva:
Summary of changes from the existing Consulting version of this component, mainly for @tatimblin: 
* I converted the LinkType to a const rather than an enum so that users can either import the value, or pass a raw string. Follows the pattern used by the `layout` prop on the Image, which came out of a slack discussion.
* The component defines its own CTA type, in the type I made `label` optional and changed its type from string to ReactNode so that it's easier to use things like icons. 
* I removed the global props for now, since I think that's probably worth a separate PR to make sure we feel good about how it works in a public repo
* I removed the phone number library. It was only used for trying to autodetect phone numbers, and I'm not sure it's worth the cost of the dependency size
* I changed the linkType so that it's only inferred if it isn't explicitly specified
* I removed the `parseLinkType` function. I thin kif we type the linkType, we don't need to worry about handling arbitrary strings.
* I removed `linkType` from the base LinkConfig. I got confused by it compared to the cta.linkType, and wasn't sure if there was a specific use case we had in mind for it.

### DIFF
--- a/packages/pages/src/components/analytics/hooks.ts
+++ b/packages/pages/src/components/analytics/hooks.ts
@@ -19,7 +19,7 @@ declare global {
  *
  * @public
  */
-export function useAnalytics(): AnalyticsMethods|null {
+export function useAnalytics(): AnalyticsMethods | null {
   const ctx = useContext(AnalyticsContext);
 
   if (!ctx) {
--- a/packages/pages/src/components/analytics/hooks.ts
+++ b/packages/pages/src/components/analytics/hooks.ts
@@ -1,5 +1,5 @@
 import { ConversionDetails, Visitor } from "@yext/analytics";
-import { useContext } from "react";
+import { MouseEvent, useContext } from "react";
 import { AnalyticsContext } from "./context";
 import { concatScopes } from "./helpers";
 import { AnalyticsMethods } from "./interfaces";
@@ -19,12 +19,11 @@ declare global {
  *
  * @public
  */
-export function useAnalytics(): AnalyticsMethods {
+export function useAnalytics(): AnalyticsMethods|null {
   const ctx = useContext(AnalyticsContext);
+
   if (!ctx) {
-    throw new Error(
-      "Attempted to call useAnalytics outside of an AnalyticsProvider"
-    );
+    return ctx;
   }
 
   // TODO: is this the right way / place to expose a callback for use by a Cookie Management banner?
@@ -41,7 +40,7 @@ export function useAnalytics(): AnalyticsMethods {
     trackClick(
       eventName: string,
       conversionData?: ConversionDetails
-    ): (e: MouseEvent) => Promise<void> {
+    ): (e: MouseEvent<HTMLAnchorElement>) => Promise<void> {
       return ctx.trackClick(concatScopes(scope, eventName), conversionData);
     },
     setDebugEnabled(enabled: boolean): void {
@@ -74,7 +73,7 @@ export function useAnalytics(): AnalyticsMethods {
  * @public
  */
 export const useTrack = () => {
-  return useAnalytics().track;
+  return useAnalytics()?.track;
 };
 
 /**
@@ -83,7 +82,7 @@ export const useTrack = () => {
  * @public
  */
 export const usePageView = () => {
-  return useAnalytics().pageView;
+  return useAnalytics()?.pageView;
 };
 
 /**
@@ -92,5 +91,5 @@ export const usePageView = () => {
  * @public
  */
 export const useIdentify = () => {
-  return useAnalytics().identify;
+  return useAnalytics()?.identify;
 };
--- a/packages/pages/src/components/link/link.test.tsx
+++ b/packages/pages/src/components/link/link.test.tsx
@@ -1,0 +1,35 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from "react";
+import { render } from "@testing-library/react";
+import { Link } from "./link";
+
+describe("Link", () => {
+  it("renders component when given href + children", () => {
+    render(<Link href="https://yext.com">Click Me</Link>);
+  });
+
+  it("renders component when given partial cta and children", () => {
+    render(<Link cta={{ link: "https://yext.com" }}>Click Me</Link>);
+  });
+
+  it("renders component when given full cta prop and no children", () => {
+    render(
+      <Link
+        cta={{ link: "https://yext.com", label: "Click Me", linkType: "URL" }}
+      />
+    );
+  });
+
+  it("renders component when given full cta prop and no children", () => {
+    render(<Link cta={{ link: "https://yext.com" }} />);
+  });
+});
+
+// TODO: Add tests
+// Check target="_blank" present on newTab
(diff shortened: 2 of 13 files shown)


packages/pages/src/components/link/link.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { useState } from 'react';
+import classNames from "classnames";
+import { getHref, isEmail, reverse } from './methods';
+import type { CTA } from "./types";
```
tatimblin, line 4: Is this going away?

packages/pages/src/components/link/link.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  it("renders component when given partial cta and children", () => {
+    render(<Link cta={{ link: "https://yext.com" }}>Click Me</Link>);
+  });
+
+  it("renders component when given full cta prop and no children", () => {
+    render(<Link cta={{ link: "https://yext.com", label: "Click Me" }} />);
+  });
+
+  it("renders component when given full cta prop and no children", () => {
```
mkilpatrick, line 18: Partial props?

packages/pages/src/components/link/link.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+    render(<Link cta={{ link: "https://yext.com" }} />);
+  });
+});
+
+// TODO: Add tests
```
mkilpatrick, line 23: Do you plan to add these in this PR?

packages/pages/src/components/link/link.tsx
```diff
@@ -0,0 +1,114 @@
+import React, { useState } from "react";
+import classNames from "classnames";
+import { getHref, isEmail, reverse } from "./methods";
+import type { CTA } from "./types";
+
+/**
+ * Configuration options available for any usages of the Link component.
```
mkilpatrick, line 7: Add `@public` to any docs that should be shown to users. It allows the docs to output in the doc generation.

packages/pages/src/components/link/link.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const { children, onClick, className, ...rest } = props;
+    const link: CTA = isHREFProps(props) ? { link: props.href } : props.cta;
+
+    const obfuscate =
+      props.obfuscate || (props.obfuscate !== false && isEmail(link.link));
+    const [humanInteraction, setHumanInteraction] = useState<boolean>(false);
+
+    const handleClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
+      setHumanInteraction(true);
+      if (onClick) onClick(e);
```
mkilpatrick, line 87: Can you use braces? We need to update our style linting.

packages/pages/src/components/link/link.tsx
```diff
@@ -0,0 +1,114 @@
...
+      direction: useLinkAsLabel && isObfuscate ? "rtl" : "ltr",
+    };
+
+    const renderedLink = isObfuscate ? reverse(link.link) : link.link;
+
+    return (
+      <a
+        className={classNames("Link", className)}
+        href={humanInteraction || !obfuscate ? getHref(link) : "obfuscate"}
+        onClick={(e) => handleClick(e)}
```
mkilpatrick, line 104: I believe you can just do handleClick here

packages/pages/src/components/link/link.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const obfuscatedStyle: React.CSSProperties = {
+      ...props.style,
+      unicodeBidi: "bidi-override",
+      direction: useLinkAsLabel && isObfuscate ? "rtl" : "ltr",
+    };
+
+    const renderedLink = isObfuscate ? reverse(link.link) : link.link;
+
+    return (
+      <a
```
mkilpatrick, line 101: Should analytics stuff work out of the box here?

bhainesva: It should, but the analytics stuff isn't implemented yet. Mcgin is planning on updating the link component to do it after he gets his provider pr merged.

packages/pages/src/components/link/link.test.tsx
benmcginnis: I think if we're going to delete the TODO we should add an issue to add these tests

benmcginnis: but after our discussion in #151 it seems like it's fine to leave the TODOs

packages/pages/src/components/link/link.tsx
benmcginnis: adding this in now. i created a PR against this branch #179 but I'm just going to close it and merge it into this branch for simplicity

packages/pages/src/components/link/link.test.tsx
benmcginnis: sounds like this is resolved?

benmcginnis: @mkilpatrick I'm going to resolve this unless you were talking about something else

benmcginnis (approved):

packages/pages/src/components/analytics/hooks.ts
```diff
@@ -19,12 +19,11 @@ declare global {
  *
  * @public
  */
-export function useAnalytics(): AnalyticsMethods {
+export function useAnalytics(): AnalyticsMethods | null {
   const ctx = useContext(AnalyticsContext);
+
   if (!ctx) {
-    throw new Error(
```
mkilpatrick, line 25: Why remove the error and return null? Seems a little weird you need to null check everywhere now.

packages/pages/src/components/link/methods.test.ts
```diff
@@ -0,0 +1,97 @@
+import { getHref, isEmail, reverse } from "./methods";
+
+enum LinkType {
```
mkilpatrick, line 3: Shouldn't this pull in from types instead of redefining?

packages/pages/src/components/analytics/hooks.ts
benmcginnis: So in adding analytics to the Link component we realized this was an anti-pattern because it means that any component where we to automatically add analytics (e.g. the Link component) would only ever be able to be called as a child of an `AnalyticsProvider` or it would throw an error.  You can't call a hook inside of a conditional so unless we want to make the Link component useless outside of the context of an AnalyticsProvider we have to do it this way.  

I think in the future we could refactor the Analytics class so that it doesn't keep track of it's own state and instead move the state into the useAnalytics hook so that useAnalytics never returns null, but you'd still have the problem of the track method and other methods not doing anything unless you've given the template data to Analytics at some point prior to calling the `track` method.

packages/pages/src/components/link/methods.test.ts
benmcginnis: @bhainesva ?

bhainesva: fixed

benmcginnis (approved):

mkilpatrick (approved):

PR #159, opened 2022-08-09
feat(components): add Address component

tatimblin:
This component is intended to easily format a localized address from Yext platform data. Built into the component is formatting instructions per country, but users can also pass in their own custom format.

Note: The [getDirections()](https://github.com/yext/sites-react-components/blob/master/src/components/Address/methods.ts#L18) methods will eventually be added to the address components scope. I wanted to split it up to be more manageable.

### DIFF
--- a/packages/pages/src/components/address/address.tsx
+++ b/packages/pages/src/components/address/address.tsx
@@ -5,7 +5,7 @@ import { getUnabbreviated } from "./methods";
 
 /**
  * Renders an HTML address based from the Yext Knowledge Graph. Example of using the component to render
- * a location entities address from Yext Knowledge Graph:
+ * a location entity's address from Yext Knowledge Graph:
  * ```
  * import { Address } from "@yext/pages/components";
  *
@@ -14,7 +14,7 @@ import { getUnabbreviated } from "./methods";
  *       Arlington, VA, 22201,
  *       US
  * const customAddress = (<Address address={document.address} lines={[['line1', 'city', 'region']]} />);
- *   --> 1101 Wilson Blvd., Arlington, Virginia
+ *   --> 1101 Wilson Blvd., Arlington, VA
  * ```
  *
  * @public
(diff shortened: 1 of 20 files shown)


packages/pages/src/components/address/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  line3?: string;
+  city: string;
+  region?: string;
+  postalCode: string;
+  countryCode: string;
+  sublocality?: string;
+  extraDescription?: string;
+  localizedRegionName?: string; // previously in derived data
+  localizedCountryName?: string; // previously in derived data
+  ","?: string;
```
tatimblin, line 18: I added this because we wanted separators (commas) to be explicitly declared in the formatter. Maybe it should be part of a new extended interface though?

```
export AddressFormatter extends AddressType {
  ',': string;
}
```

bhainesva: Yeah I don't think this should be part of the address type. 

Can we do
```
export type AddressLine = ((keyof AddressType) | ",")[];
```
instead?

packages/pages/src/components/address/address.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * Renders an HTML address based from the Yext Knowledge Graph. Example of using the component to render
+ * a location entities address from Yext Knowledge Graph:
+ * ```
+ * import { Address } from "@yext/pages/components";
+ *
+ * const address = (<Address address={document.address} />);
+ *   --> 1101 Wilson Blvd., Suite 2300,
+ *       Arlington, VA, 22201,
+ *       US
+ * const customAddress = (<Address address={document.address} lines={[['line1', 'city', 'region']]} />);
```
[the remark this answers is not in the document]
tatimblin: It doesn't, (confirmed and) fixed the comment

mkilpatrick (approved):
LGTM, can you just resolve all comments? I'll approve but assume you want ProServe eyes on this too before merging.

PR #161, opened 2022-08-11
fix: update getRuntime to work with the plugins system

jimboHenris:
We noticed a bug where our getRuntime function makes the
assumption that a global window object will be available.
This will not be true when running in the plugins system
and will cause page generation to fail.

TEST=auto,manual

Updated unit tests. Tested locally to make sure getRuntime
still returns the expected results for node, deno, and
browser.

### DIFF
--- a/packages/pages/CHANGELOG.md
+++ b/packages/pages/CHANGELOG.md
@@ -1,25 +1,21 @@
 # 1.0.0-beta.1 (2022-08-05)
 
-
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


nbramblett (approved):

mkilpatrick:
LG, but can you update the PR title prefix before merging:
`fix(util):`

mkilpatrick (approved):

mancusi (approved):

PR #162, opened 2022-08-12
docs(components): init storybook

tatimblin:
add story for Image component


packages/pages/package.json
```diff
@@ -109,7 +111,17 @@
...
+    "ts-jest": "^28.0.7",
+    "@storybook/react": "^6.5.10",
+    "@storybook/addon-links": "^6.5.10",
+    "@storybook/addon-essentials": "^6.5.10",
+    "@storybook/addon-actions": "^6.5.10",
+    "@storybook/addon-interactions": "^6.5.10",
+    "@storybook/testing-library": "^0.0.13",
+    "@storybook/builder-webpack4": "^6.5.10",
+    "@storybook/manager-webpack4": "^6.5.10",
+    "@babel/core": "^7.18.10",
```
mkilpatrick, line 123: Is babel necessary?

mkilpatrick (commented):
storybook isn't a valid scope so it'll fail the PR check. This seems like a doc thing to me, so I'd update the title to something like
`docs: add storybook

And then in the description mention adding a story for the Image component.

packages/pages/package.json
tatimblin: @mkilpatrick There's an [issue on Github](https://github.com/storybookjs/storybook/issues/12826#issuecomment-713478017) where it looks like you can use a pre-built fork, or disable passing the components through babel (though it would still be a dependency). It's only a dev dependency so we could leave it, but I understand wanting to limit that too.

What are your thoughts on this?

mkilpatrick: Nah it's fine. Didn't realize what it was needed for.

mkilpatrick (approved):

mkilpatrick (approved):

PR #169, opened 2022-08-16
fix(dev): prevent crashing on empty id

### LINKED ISSUE #168: Dev server crashes when visiting url without entityId
### Describe the bug

I'm serving a site with a `job` template. If I visit a url like `http://localhost:5173/job/<valid-id>` the page renders as expected. If I visit a url like `http://localhost:5173/job/<invalid-id>` I get a 500, also as expected. 

If I visit `http://localhost:5173/job` the dev server crashes with this error:
```
Generated 10 files for stream "job-stream"/Users/bhaines/repo/<REPONAME>/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

SyntaxError: Unexpected token { in JSON at position 485
    at JSON.parse (<anonymous>)
    at ChildProcess.<anonymous> (file:///Users/bhaines/repo/<REPONAME>/node_modules/@yext/pages/dist/dev/server/ssr/generateTestData.js:85:25)
    at ChildProcess.emit (node:events:520:28)
    at maybeClose (node:internal/child_process:1090:16)
    at Socket.<anonymous> (node:internal/child_process:449:11)
    at Socket.emit (node:events:520:28)
    at Pipe.<anonymous> (node:net:687:12)

Node.js v17.4.0
error Command failed with exit code 7.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Reproduction

_No response_

### System Info

```shell
System:
    OS: macOS 12.3
    Memory: 5.76 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.4.0 - ~/.nvm/versions/node/v17.4.0/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v17.4.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.79
    Firefox: 103.0.2
    Safari: 15.4
  npmPackages:
    @yext/pages: 1.0.0-beta.1 => 1.0.0-beta.1
```

### Used Package Manager

yarn

### Used Browser

Firefox

### Validations

nbramblett:
Fixes #168 by catching the json parse error and returning null, which prompts the server to return an error.

It will still 500, when a 404 would _probably_ be preferable, but at least the server will not crash now.

### DIFF
--- a/packages/pages/src/dev/server/ssr/generateTestData.ts
+++ b/packages/pages/src/dev/server/ssr/generateTestData.ts
@@ -50,7 +50,7 @@ export const generateTestDataForPage = async (
 
   const command = "yext";
   const args = [
-    "sites",
+    "pages",
     "generate-test-data",
     "--featureName",
     `'${featuresConfig.features[0]?.name}'`,
@@ -127,17 +127,27 @@ export const generateTestDataForPage = async (
     });
 
     childProcess.on("close", () => {
+      let parsedData: any;
       if (testData) {
-        testData = JSON.parse(testData.trim());
+        try {
+          parsedData = JSON.parse(testData.trim());
+        } catch (e) {
+          stdout.write(
+            `\nUnable to parse test data from command: \`${command}${args.join(
+              " "
+            )}\``
+          );
+          resolve(null);
+        }
       } else {
         stdout.write(
-          `Unable to generate test data from command: \`${command}${args.join(
+          `\nUnable to generate test data from command: \`${command}${args.join(
             " "
           )}\``
         );
       }
 
-      resolve(testData);
+      resolve(parsedData);
     });
   });
 };
--- a/packages/pages/src/dev/server/ssr/pageLoader.ts
+++ b/packages/pages/src/dev/server/ssr/pageLoader.ts
@@ -84,10 +84,6 @@ export const pageLoader = async ({
       locale,
       projectStructure
     );
-    if (document === null) {
-      // Fall back to localData
-      document = await getLocalDataForEntity(entityId, locale);
-    }
   } else {
     // Get the document from localData
     document = await getLocalDataForEntity(entityId, locale);
--- a/packages/pages/src/dev/server/ssr/pageLoader.ts
+++ b/packages/pages/src/dev/server/ssr/pageLoader.ts
@@ -84,6 +84,10 @@ export const pageLoader = async ({
       locale,
       projectStructure
     );
+    if (document === null) {
+      // Fall back to localData
+      document = await getLocalDataForEntity(entityId, locale);
+    }
   } else {
     // Get the document from localData
     document = await getLocalDataForEntity(entityId, locale);


packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -127,17 +127,27 @@ export const generateTestDataForPage = async (
...
-        testData = JSON.parse(testData.trim());
+        try {
+          parsedData = JSON.parse(testData.trim());
+        } catch (e) {
+          stdout.write(
+            `\nUnable to parse test data from command: \`${command}${args.join(
+              " "
+            )}\``
+          );
+          resolve(null);
```
mkilpatrick, line 140: Is this what's actually fixing the crash?

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -50,7 +50,7 @@ export const generateTestDataForPage = async (
 
   const command = "yext";
   const args = [
-    "sites",
+    "pages",
```
mkilpatrick, line 53: Nice update

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -127,17 +127,27 @@ export const generateTestDataForPage = async (
...
+            `\nUnable to parse test data from command: \`${command}${args.join(
+              " "
+            )}\``
+          );
+          resolve(null);
+        }
       } else {
         stdout.write(
-          `Unable to generate test data from command: \`${command}${args.join(
+          `\nUnable to generate test data from command: \`${command}${args.join(
```
mkilpatrick, line 144: 🔥 on the newline

packages/pages/src/dev/server/ssr/pageLoader.ts
```diff
@@ -84,6 +84,10 @@ export const pageLoader = async ({
       locale,
       projectStructure
     );
+    if (document === null) {
+      // Fall back to localData
+      document = await getLocalDataForEntity(entityId, locale);
```
mkilpatrick, line 89: Idk if we want to do this. Can you verify in the discuss-react channel what the functionality should be? It could be a little strange where you have local data for an entity that no longer exists and you're dev'ing against it only to find in prod that it's not there. Sort of and edge case and this fallback could be useful. Just not sure so let's see what Product wants.

packages/pages/src/dev/server/ssr/generateTestData.ts
nbramblett: the general try/catch is preventing it, but resolve(null) lets us actually indicate that we couldn't provide the data so that the caller can handle that case more elegantly.

packages/pages/src/dev/server/ssr/pageLoader.ts
nbramblett: Updating to not fall back and just let it bomb out

mkilpatrick (approved):

PR #170, opened 2022-08-18
feat: extract features.json generation into its own command

mkilpatrick:
Moves features.json generation (from the templates files) out of the plugin build step and into its own `pages generate features` command. Users have been confused why they needed to run a build before they could generate test data. This solves that.


packages/pages/src/bin/pages.ts
```diff
@@ -38,7 +34,41 @@ switch (command) {
...
+  case "generate": {
+    const generateTargets = ["features"];
+
+    if (args.length != 2) {
+      process.stdout.write(
+        `Missing a target to generate. Valid values are: ${generateTargets.join(
+          ", "
+        )}\n`
+      );
+      process.exit(1);
```
mancusi, line 49: return

packages/pages/src/common/src/template/internal/loader.ts
```diff
@@ -0,0 +1,73 @@
...
+      );
+
+    importedModules.push({
+      ...templateModuleInternal,
+      path: templateModulePath,
+    });
+  }
+
+  return importedModules.reduce((prev, module) => {
+    if (prev.has(module.config.name)) {
```
mancusi, line 60: at this point we've reconciled the config name right? like we've checked if a `name` was set on the config and otherwise use the filename?

packages/pages/src/common/src/template/internal/loader.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Loads all templates in the project.
+ * @param templateModulePaths the templates filepaths to load as modules
+ * @param transpile set to true if the templates need to be transpiled (such as when they are in tsx format)
+ * @param adjustForFingerprintedAsset removes the fingerprint portion (for server bundles)
+ * @returns Promise<{@link TemplateModuleCollection}>
+ */
+export const loadTemplateModules = async (
+  templateModulePaths: string[],
+  transpile: boolean,
```
mancusi, line 20: when would we not want to transpile or adjust for fingerprinted assets? 

I'd think we'd *always* want to do this.

packages/pages/tests/fixtures/template.js
```diff
@@ -0,0 +1,106 @@
+var __create = Object.create;
```
mancusi, line 1: this JS file is already transpiled so it'll load but there's no guarantee other js files will be written in a form accepted by default

mancusi (changes requested):
mostly questions -- not too many real comments

packages/pages/src/common/src/template/internal/loader.ts
mkilpatrick: This function is used by the generate features.json command, which loads the raw tsx file, so in that case we need to transpile. But this is also used by the plugin code (was trying to consolidate the loading of templates code), but at that point it's using the server assets which are already transpiled by this point. The server bundles also need to do the adjusting thing.

Do you think we shouldn't use in both places or do something else?

packages/pages/tests/fixtures/template.js
mkilpatrick: Are you saying to rerun it through the transpilation step?

packages/pages/src/common/src/template/internal/loader.ts
mkilpatrick: This ensures there aren't two features using the same name. This actually came from your code originally.

packages/pages/src/bin/pages.ts
mkilpatrick: If the process exits there's no need to return.

packages/pages/src/common/src/template/internal/loader.ts
```diff
@@ -0,0 +1,73 @@
...
+  adjustForFingerprintedAsset: boolean
+): Promise<TemplateModuleCollection> => {
+  const importedModules = [] as TemplateModuleInternal<any, any>[];
+  for (const templateModulePath of templateModulePaths) {
+    let templateModule = {} as TemplateModule<any, any>;
+    try {
+      if (transpile) {
```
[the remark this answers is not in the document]
mkilpatrick: What do you mean by _these_? Bundle is set to true when transpiling.

packages/pages/src/bin/pages.ts
mancusi: Yeah I know but seems safer to do an explicit return as well to prevent unintended behavior.

packages/pages/src/common/src/template/internal/loader.ts
mancusi: Your response doesn't answer my original question. 

Is the config name on the module at this point specifically what's set in `config.name` in a TemplateConfig or is it the reconciled value of: `config.name if set OR filename`

packages/pages/src/common/src/template/internal/loader.ts
mancusi: I see. Sounds good then.

packages/pages/tests/fixtures/template.js
mancusi: I was just saying that this file doesn't seem like a realistic fixture because the module appears to be already transpiled (unless that was the explicit intention of the test)

packages/pages/src/common/src/template/internal/loader.ts
mancusi: If it's just a copy of what already existed then I assume it worked though in which case you can disregard this.

packages/pages/src/common/src/template/internal/loader.ts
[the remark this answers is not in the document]
mancusi: similar to https://github.com/yext/pages/pull/170#discussion_r950382269

packages/pages/src/common/src/template/internal/loader.ts
mkilpatrick: `convertTemplateModuleToTemplateModuleInternal` (called above) handles determining the name to use; it prefers the `name` defined in TemplateConfig and if not set uses the filename.

packages/pages/src/bin/pages.ts
mkilpatrick: Can't `A 'return' statement can only be used within a function body.`

mancusi (approved):

PR #171, opened 2022-08-19
feat(components): add Map component

tatimblin:
render a Google or Mapbox map using the existing Yext locator build pattern methods

J=CR-2808


packages/pages/src/components/map/marker.tsx
```diff
@@ -0,0 +1,95 @@
...
+    , pinEl);
+  }
+
+  return null;
+}
+
+Marker.defaultProps = {
+  hideOffscreen: false,
+  icon: (
+    <svg width="30" height="38" viewBox="0 0 30 38" fill="none" xmlns="http://www.w3.org/2000/svg">
```
mkilpatrick, line 86: Should we make this an imported file?

packages/pages/src/components/map/map.tsx
```diff
@@ -0,0 +1,129 @@
...
+  className,
+  clientKey,
+  controls,
+  getMapOptions,
+  padding,
+  panStartHandler,
+  provider,
+  providerOptions,
+  singleZoom,
+  ...rest
```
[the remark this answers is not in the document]
tatimblin: rest is an object equivalent to all the other props not pulled out in the lines above

packages/pages/src/components/map/map.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { MapProps, MapContextType } from "./types";
+import "./map.css";
+
+export const MapContext = createContext<MapContextType | null>(null);
+
+export function useMapContext() {
+  const ctx = useContext(MapContext);
+
+  if (!ctx || ctx.map === undefined) {
+    throw new Error('Attempted to call useMapContext() outside of <Map>.');
```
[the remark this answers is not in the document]
Jared-Hood: Tests added for both failure and success

packages/pages/src/components/map/map.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  return ctx.map;
+}
+
+export const Map = ({
```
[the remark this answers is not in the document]
Jared-Hood: The internals of the map api wrapper is a singleton class so it will only load the specific provider code once. Then it just creates a new instance of the map in each wrapper. I created a storybook where two maps are loaded.

packages/pages/src/components/map/marker.tsx
Jared-Hood: I'm not opposed to making this an imported file, but I think it's also fine being inlined here.

packages/pages/src/components/map/map.tsx
```diff
@@ -1,17 +1,26 @@
...
-import { GoogleMaps, Map as MapType, MapOptions } from "@yext/components-tsx-maps";
+import {
+  GoogleMaps,
+  Map as MapType,
+  MapOptions,
+} from "@yext/components-tsx-maps";
 import { Coordinate, GeoBounds } from "@yext/components-tsx-geo";
-import { MapProps, MapContextType } from "./types";
-import css from "./map.module.css";
+import type { MapProps, MapContextType } from "./types";
```
mkilpatrick, line 15: The linter should have complained that this doesn't end in .js. Did it not do that on commit? Do you get errors on `pnpm lint`?

packages/pages/src/components/map/marker.tsx
```diff
@@ -65,31 +43,41 @@ export const Marker = ({
...
   return null;
-}
+};
 
 Marker.defaultProps = {
   hideOffscreen: false,
   icon: (
-    <svg width="30" height="38" viewBox="0 0 30 38" fill="none" xmlns="http://www.w3.org/2000/svg">
-      <path d="M30 15.0882C30 23.4212 23.3333 30.7353 15 38C7.22222 31.2941 0 23.4212 0 15.0882C0 6.75523 6.71573 0 15 0C23.2843 0 30 6.75523 30 15.0882Z" fill="red" />
+    <svg
```
mkilpatrick, line 67: Can you extract this to a named const outside of the component (aka the Marker const)?

packages/pages/src/components/map/map.tsx
Jared-Hood: No it's not picking up on that file, but I'll add them now.

Jared-Hood: pnpm lint did callout the marker.tsx file through the commit flow and running it manually just called out the index file, which I also missed. Neither though found any issue with the map.tsx file.

mkilpatrick (approved):
LGTM as long as ProServe is good with it too

PR #172, opened 2022-08-22
feat(components): add getDirections function

tatimblin:
power 'Get Directions' CTAs. This function follows a hierarchy of methods to derive an entities map url, starting with using the GMB id managed by Listings. There are two TODOs which would align us completely with classic pages behviour once we know more about how a GMB/streams data structure, this shouldn't be considered blocking.

J=CR-2816

### DIFF
--- a/packages/pages/src/components/address/methods.ts
+++ b/packages/pages/src/components/address/methods.ts
@@ -1,8 +1,11 @@
 import {
   AddressType,
+  GetDirectionsConfig,
   ListingPublisher,
+  ListingPublisherOption,
   ListingType,
   MapProvider,
+  MapProviderOption,
 } from "./types";
 
 /**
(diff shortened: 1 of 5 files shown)


packages/pages/src/components/address/methods.ts
```diff
@@ -22,3 +27,152 @@ export const getUnabbreviated = (
 
   return unabbreviatedField && address[unabbreviatedField];
 };
+
+/**
+ * Get a third-party maps url for a Yext location
+ *
+ * @param {any} profile - Partial of Yext Location entity profile
```
benmcginnis, line 34: I feel like we should specify the profile fields we require?

packages/pages/src/components/address/methods.ts
```diff
@@ -22,3 +27,152 @@ export const getUnabbreviated = (
...
+ * @param {Provider} provider - Google, Apple, Bing
```
mkilpatrick, line 35: MapProvider is the actual type

packages/pages/src/components/address/types.ts
```diff
@@ -40,3 +40,24 @@ export type AddressLineProps = {
...
+
+/**
+ * The type definition for a Listing
+ */
+export interface ListingType {
+  listingUrl: string;
+  publisher: ListingPublisher;
+}
+
+export enum MapProvider {
```
mkilpatrick, line 59: You should do the trick where we allow enum or string so that a user doesn't need to import the type. You can see other examples in other components, such as Image.

packages/pages/src/components/address/types.ts
```diff
@@ -40,3 +40,24 @@ export type AddressLineProps = {
   line: AddressLine;
   separator?: string;
 };
+
+/**
+ * The available listing publishers
+ */
+export enum ListingPublisher {
```
mkilpatrick, line 47: Same enum/string comment as below.

packages/pages/src/components/address/methods.ts
mkilpatrick: Agreed. This should take in the items it needs rather than the entire profile.

packages/pages/src/components/address/methods.ts
```diff
@@ -22,3 +27,152 @@ export const getUnabbreviated = (
...
+      query = encodeArray([
+        address?.line1,
+        address?.city,
+        address?.region,
+        address?.postalCode,
+      ]);
+      return getDirectionsBing(query, directions);
+      break;
+    default:
+      const gmbURL = getlistingUrl(
```
mkilpatrick, line 72: It's little weird you construct the gmbURL and then check if it exists. Can you determine if you should construct/return it beforehand? Like check if ref_listings then return gmb, else if googlePlaceId return getDirectionsGooglePlaceID(), else return getDirectionsGoogle()

packages/pages/src/components/address/methods.ts
```diff
@@ -22,3 +27,152 @@ export const getUnabbreviated = (
...
+ * Get a maps url from listings
+ *
+ * @param {Yext.ListingType[]} listings - Yext listings
+ * @param {ListingPublisher} publisher - provider to get listing for
+ * @returns {string} - url
+ */
+const getlistingUrl = (
+  listings: ListingType[] = [],
+  publisher: ListingPublisher
+): string | undefined => {
```
mkilpatrick, line 161: Based on my previous comment you shouldn't need this to return undefined.

packages/pages/src/components/address/methods.ts
tatimblin: I changed the method to just get the `ListingType` so it is more obvious that it has to see if a listings is available for that provider first. Also renamed to `getListingByProvider`

packages/pages/src/components/address/methods.ts
tatimblin: It isn't guaranteed that a client will have GMB setup. It's rare but you could buy pages without listings. Does this not assuming sound good?

packages/pages/src/components/address/methods.ts
```diff
@@ -31,74 +34,85 @@ export const getUnabbreviated = (
...
-      const gmbURL = getlistingUrl(
-        profile.ref_listings,
-        ListingPublisher.googlemybusiness
+      const gmb = getListingByProvider(
+        listings,
+        ListingPublisherOption.GOOGLEMYBUSINESS
       );
-      if (gmbURL) {
-        return gmbURL;
+      if (gmb) {
```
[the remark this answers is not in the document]
tatimblin: Right now we are mimicking what is in the .soy version of the component

packages/pages/src/components/address/methods.ts
```diff
@@ -107,61 +121,71 @@ const getDirectionsApple = (query: string, directions?: boolean): string => {
...
  * @returns {string} - url
  */
-const getlistingUrl = (
+const getListingByProvider = (
   listings: ListingType[] = [],
   publisher: ListingPublisher
-): string | undefined => {
+): ListingType | undefined => {
   // TODO: Extract cid from URL to use in directions search
   const gmb = listings.find((l) => l.publisher === publisher);
```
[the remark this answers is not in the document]
tatimblin: @mkilpatrick Switched it to a converting listings array to map, let me know what you think

packages/pages/src/components/address/methods.ts
```diff
@@ -31,74 +34,89 @@ export const getUnabbreviated = (
...
+      address.line2,
+      address.city,
+      address.region,
+      address.postalCode,
+      address.countryCode,
+    ]);
+
+  switch (config.provider) {
+    case MapProviderOption.APPLE:
+      // Apple Maps requires a query string
```
mkilpatrick, line 66: Do you want to log a console warning or something?

mkilpatrick (approved):

Issue #160, opened 2022-08-11
Windows Issues


lymarrie:
### Describe the bug

Hi,

The build seems to fail on Windows machines. I used node version 17.9.1.
I run into the following error after running `yext pages generate`

```
Warning: CI execution for this site will be in Linux - please be sure your build commands are 
compatible in that environment
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead. 
✖ [13ms] Nothing to clean
✖ [10ms] Failed to copy Yext plugin files
✔ [27ms] Successfully copied Yext plugin files
✔ [60ms] Generated 2 hydration entry-points
✔ [3ms] Injected template renderer.
✓ 15 modules transformed.
✔ [4ms] Validated template modules
✔ [5ms] Successfully wrote C:\Users\[name]\react\pages-starter-react-locations\sites-config
✖ [2ms] Failed to write manifest.json
ENOENT: no such file or directory, open 'C:\Users\[name]\react\pages-starter-react-locations\dist/plugin/manifest.json'
[vite]: Rollup failed to resolve import "....src        emplateslocation" from "dist/hydration_templates/location.tsx".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
error during build:
Error: [vite]: Rollup failed to resolve import "....src emplateslocation" from "dist/hydration_templates/location.tsx".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
    at onRollupWarning (file:///C:/Users/[name]/react/pages-starter-react-locations/node_modules/vite/dist/node/chunks/dep-71eb12cb.js:43609:19)
    at onwarn (file:///C:/Users/[name]/react/pages-starter-react-locations/node_modules/vite/dist/node/chunks/dep-71eb12cb.js:43399:13)
    at Object.onwarn (file:///C:/Users/[name]/react/pages-starter-react-locations/node_modules/rollup/dist/es/shared/rollup.js:23178:13)
    at ModuleLoader.handleResolveId (file:///C:/Users/[name]/react/pages-starter-react-locations/node_modules/rollup/dist/es/shared/rollup.js:22335:26)
    at file:///C:/Users/[name]/react/pages-starter-react-locations/node_modules/rollup/dist/es/shared/rollup.js:22296:26
```

### System Info

```shell
[...]

PR #173, opened 2022-08-22
fix(dev/plugin): address incompatibilities with windows

mancusi:
This CR handles a number of incompatibilities the PagesJS toolchain had with windows and updates it to make it functional. Specifically, I've updated entry-points in our code which touch the underlying operating system to be path separator aware using `path.join`. 

I've made the opinionated decision that we only want these entry-points to change while keeping the internals of the system using posix-style separators. This is because node requires posix-style separators for imports and mixing them internally seems prone to cause more unintentional bugs in the future. This notably means that even on windows, the paths generated by our `manifest.json` still use posix-style separators. Considering we use the manifest internally to import and match node imports this consistency is important.

Fixes: https://github.com/yext/pages/issues/160
J=SUMO-4729
TEST=auto,manual
 
Ran tests and tested the following on windows and unix:
- hydration 
- prod mode page generation
- dev mode page generation (dynamic)
- dev mode page generation (local)
- stateful updating


packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -62,11 +63,12 @@ const discoverInputs = async (
       const parsedPath = parse(template);
 
       if (parsedPath.ext === ".tsx" || parsedPath.ext === ".jsx") {
-        input[`hydrate/${parsedPath.name}`] =
-          `${hydrationOutputDir}/${template}`.replace("jsx", "tsx");
+        input[`hydrate/${parsedPath.name}`] = path
```
mkilpatrick, line 66: Do the keys need to be transformed?

mancusi:
waiting on https://github.com/yext/pages/pull/174 before I un-draft this PR

packages/pages/src/vite-plugin/build/buildStart/buildStart.ts
```diff
@@ -60,18 +61,26 @@ const copyPluginFiles = (fileEmitter: EmitFile) => {
     startLog: "Copying Yext plugin files",
   });
 
-  const currentPath = new URL(import.meta.url).pathname;
-  const pathToPluginsDir = path.resolve(currentPath, "../../../../plugin");
-  const pluginFiles = glob.sync(`${pathToPluginsDir}/*.ts`);
+  const currentPath = fileURLToPath(import.meta.url);
+  const pathToPluginsDir = path.resolve(
+    currentPath,
+    path.join("..", "..", "..", "..", "plugin")
```
mkilpatrick, line 67: Pretty sure this could be one string and it's replace the slashes automatically.

mancusi: that feels more confusing than just keeping it like this (bc it mutates the "/")

packages/pages/src/vite-plugin/build/build.ts
mancusi: the longer answer is that rollup handles these correctly already.

nbramblett:
> Ran tests and tested the following on windows and unix:
> ...
>     * dev mode page generation

Bit of a nitpick here but did you test dev both with and without the `--local` flag? When it's disabled, it does a file query on the localData directory so we want that to not break ideally.

mancusi:
> > Ran tests and tested the following on windows and unix:
> > ...
> > * dev mode page generation
> 
> Bit of a nitpick here but did you test dev both with and without the `--local` flag? When it's disabled, it does a file query on the localData directory so we want that to not break ideally.

yep this also works. I'll update the title.

packages/pages/src/common/src/project/structure.ts
```diff
@@ -1,5 +1,6 @@
 import pathLib from "path";
 import _ from "lodash";
+import path from "path";
```
nbramblett, line 3: isn't this redundant with pathLib?

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -151,3 +142,31 @@ export const generateTestDataForPage = async (
...
   });
 };
+
+const addCommonArgs = (
+  featuresConfig: FeaturesConfig,
+  featureName: string,
+  locale: string
+) => {
+  // We need to specially handle the JSON when running on windows due to an existing bug/behavior in Powershell where inner quotes are
+  // stripped from strings when passed to a third-party program. Read more: https://stackoverflow.com/questions/52822984/powershell-best-way-to-escape-double-quotes-in-string-passed-to-an-external-pro.
```
nbramblett, line 152: Is this equally a problem in the _output_? We run JSON.parse on the output of this command.

mkilpatrick (approved):

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -151,3 +142,31 @@ export const generateTestDataForPage = async (
...
+  let jsonConfig;
+  if (process.platform == "win32") {
+    jsonConfig = `"${JSON.stringify(featuresConfig).replaceAll(`"`, `\\"`)}"`;
```
nbramblett, line 155: just to double check, does this correctly handle cases where an inner string already has an escaped quote? So like if my json string looks like `{"msg": "Hello \"friend\""}` that will parse correctly? It presumably would become `\\\"` but I think it might need to be `\\\\"` because the first slash isn't being doubled (the double slash in general is odd to me though).

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -5,6 +5,7 @@ import { readdir } from "fs/promises";
 import { parse } from "path";
 import { InputOption } from "rollup";
 import { ProjectStructure } from "../../common/src/project/structure.js";
+import path from "path";
```
nbramblett, line 8: combine with above import?

`import path, {parse} from "path"` I think would work.

packages/pages/src/vite-plugin/build/buildStart/buildStart.ts
```diff
@@ -60,18 +61,26 @@ const copyPluginFiles = (fileEmitter: EmitFile) => {
...
+  );
+  // We must use path.resolve to reconcile filepaths on Windows as glob returns filepaths with forward slashes by default.
+  const pluginFiles = glob
+    .sync(`${pathToPluginsDir}/*.ts`)
```
nbramblett, line 71: shouldn't this be joined as well? Or is glob.sync posix compliant

packages/pages/src/dev/server/ssr/generateTestData.ts
mancusi: No.

packages/pages/src/vite-plugin/build/buildStart/buildStart.ts
mancusi: glob suggests users to use `/` even on windows

packages/pages/src/common/src/project/structure.ts
mancusi: yeah and it's unused, I'll clean it up.

packages/pages/src/dev/server/ssr/generateTestData.ts
mancusi: This doesn't work but it also doesn't work currently anyways: `Stream validation diagnostic for stream "my-stream\"-id-1": invalid name my-stream"-id-1 at position: /source`

mancusi: there's a change I can make to the regex to make it more robust which I'll add.

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -45,31 +45,24 @@ export const generateTestDataForPage = async (
 ): Promise<any> => {
   const siteStreamPath = path.resolve(
     process.cwd(),
-    projectStructure.sitesConfigRoot + "/" + projectStructure.siteStreamConfig
+    projectStructure.sitesConfigRoot.getAbsolutePath() +
+      "/" +
```
nbramblett, line 49: update?

nbramblett (approved):
Approving so that if my comment isn't an issue you can just merge after acking

mkilpatrick (approved):

PR #179, opened 2022-08-24
add Link analytics

benmcginnis:


packages/pages/src/components/analytics/Analytics.ts
```diff
@@ -179,7 +180,7 @@ export class Analytics implements AnalyticsMethods {
   trackClick(
     eventName: string,
     conversionData?: ConversionDetails
-  ): (e: MouseEvent) => Promise<void> {
+  ): (e: MouseEvent<HTMLAnchorElement>) => Promise<void> {
```
bhainesva, line 183: Should we also support button elements?

packages/pages/src/components/link/link.tsx
```diff
@@ -86,17 +91,28 @@ export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
     const { children, onClick, className, ...rest } = props;
     const link: CTA = isHREFProps(props) ? { link: props.href } : props.cta;
 
+    let trackEvent: string|null;
+    let analytics: AnalyticsMethods|null;
+
+    if (AnalyticsContext !== null) {
+      trackEvent = props.eventName || props.cta ? "cta" : "link";
+      analytics = useAnalytics();
```
bhainesva, line 99: hooks can only be called at the top level of the component: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

packages/pages/src/components/analytics/Analytics.ts
benmcginnis: so i was wondering about this, `buttons` should not navigate and trackClick won't actually work if you put it on a button as it needs the `href` to figure out what it's going to do. As it stands right now if there is now `href` it behaves just like `track()` so people should really only use this on an `a` element.

benmcginnis: so i guess i'm saying, No, a button should use `track` and not `trackClick`

packages/pages/src/components/link/link.tsx
benmcginnis: probably should add the hooks es lint rule

PR #187, opened 2022-08-26
fix(plugin): add features.json generation back to build plugin

mkilpatrick:
Prior work was done to extract features.json generation from the build plugin. Unfortunately this is causing problems to run as a separate `pages` command in the YextCI system due to the experimental node flag required (which is actually only needed for the dev command). In order to avoid the issue in the meantime so we can unblock people and cut new releases, add the features generation back to the build as it was beforehand. Users can still generate features.json themselves though.


mancusi (approved):
glad it's encapsulated well-enough that this is easy to add back 👍

PR #198, opened 2022-08-31
chore: auto-format and lint as pre-commit hook

mkilpatrick:
When a user commits, prettier and `eslint --fix` will now run as a pre-commit hook automatically. This also removes the format Github Action that was causing merge issues and actions to not re-run correctly.

For those that use an IDE for committing (like Sublime Merge), follow these steps - https://typicode.github.io/husky/#/?id=command-not-found


mancusi:
> For those that use an IDE for committing (like Sublime Merge), follow these steps - https://typicode.github.io/husky/#/?id=command-not-found

Can you add this to some more visible dev documentation in here?

mancusi (approved):

PR #199, opened 2022-09-02
fix(dev): local dev not working on linux distros

mkilpatrick:
The local dev server would fail to start because node flags are not properly passed along in all operating systems. The workaround is to execute a shell script that executes the script via node with the flags. See https://stackoverflow.com/a/68154775.

Fixes https://github.com/yext/pages/issues/132


packages/pages/src/bundler.js
```diff
@@ -63,11 +63,7 @@ if (pluginFiles.length == 0) {
 }
 mkdirSync(pluginOutputPath, { recursive: true });
 pluginFiles.map((filepath) =>
-  copyFile(
-    filepath,
-    `${pluginOutputPath}${path.basename(filepath)}`,
-    () => undefined
-  )
+  copyFileSync(filepath, `${pluginOutputPath}${path.basename(filepath)}`)
```
nbramblett, line 66: why change this?

packages/pages/src/bundler.js
```diff
@@ -79,3 +75,11 @@ try {
 } catch (e) {
   console.error(e);
 }
+
+/**
+ * Copy the sh binary.
+ */
+copyFileSync(
+  path.resolve("./src/bin/pages.sh"),
+  path.resolve("./dist/bin/pages.sh")
```
nbramblett, line 84: use pluginOutputPath?

packages/pages/src/bundler.js
mkilpatrick: I was using copyFileSync in another place so updated the rest. The interface is nicer too since you don't need to handle the errors yourself.

packages/pages/src/bundler.js
mkilpatrick: pluginOutputPath is a different path than the bin path

mancusi (changes requested):
some tests are failing

mancusi (approved):

PR #201, opened 2022-09-06
feat(plugin): add urlWriteback plugin

bhainesva:
Tested by deploying this site: https://github.com/big-ol-sandbox/bhaines-site-starter-react-basic/blob/full/functions/onUrlChange/urlWriteback.ts. 

And verifying the field gets updated when a deploy happens: https://sandbox.yext.com/s/3160439/yextsites/112010/branch/252/deploys/recent


packages/pages/src/plugins/urlWriteback/yext.ts
```diff
@@ -0,0 +1,81 @@
...
+    v: options?.v || "20220903",
+  }), {
+    method: "PUT",
+    body: JSON.stringify(body),
+    headers: { "Content-Type": "application/json; charset=utf-8" },
+  });
+  const response = await fetch(req);
+  if (response.status < 200 || response.status >= 300) {
+    const responseBody = await response.json() as ApiResponse<T>;
+    throw responseBody.meta.errors[0].message
```
tatimblin, line 77: Curious if this would be visible anywhere in the platform

bhainesva: Worst case it's visible in the snowflake logs. Bryan has told me it's also visible in the platform on the deploys but I haven't seen exactly where.

packages/pages/src/plugins/index.ts
```diff
@@ -0,0 +1,2 @@
+import urlWritebackPlugin from "./urlWriteback/index.ts";
```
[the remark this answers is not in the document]
bhainesva: cc: @mkilpatrick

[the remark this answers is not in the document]
mkilpatrick: If they're in their own workspace then it'll be a different npm package which we don't want.

packages/pages/src/plugins/index.ts
```diff
@@ -0,0 +1,2 @@
...
+export { urlWritebackPlugin }
```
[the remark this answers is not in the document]
mkilpatrick: I'd call it `urlWriteback` since the usage will be `import { urlWriteback } from "@yext/pages/plugins"`

packages/pages/src/plugins/urlWriteback/index.ts
```diff
@@ -0,0 +1,32 @@
+import { updateEntity } from "./yext.ts";
+
+interface UrlWritebackConfig {
```
mkilpatrick, line 3: I think this needs to be exported as it's the input to the func

packages/pages/src/plugins/urlWriteback/test.ts
```diff
@@ -0,0 +1,11 @@
+import { assertEquals } from "https://deno.land/std@0.154.0/testing/asserts.ts";
+import { buildApiUrl } from "./yext.ts";
+
+Deno.test("buildApiUrl", () => {
+	const url = buildApiUrl("https://www.yext.com", `entities/123`, {
```
mkilpatrick, line 5: Did the hook not auto format?

packages/pages/src/plugins/urlWriteback/yext.ts
mkilpatrick: [Example](https://www.yext.com/s/1089287/yextsites/37118/branch/914/deploy/6gr6dp3pqu/stream/01GBWRH21XYPG2KDW3G65VRMQ2) - see the console log for plugin invocation section

bhainesva: Do logs for all of the plugins (onUrlChange, OnPageGenerate, w/e) show up in that section?

packages/pages/src/plugins/index.ts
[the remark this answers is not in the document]
bhainesva: Does it make sense for this to be in an npm package at all given that it's intended to be run from Deno? I'm not sure what the use case for importing into a node environment would be.

packages/pages/src/plugins/urlWriteback/test.ts
bhainesva: I guess not 🤷

packages/pages/src/plugins/index.ts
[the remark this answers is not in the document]
bhainesva: It looks like it's breaking the tests as is because of the `Deno` reference and the different import styles

packages/pages/src/plugins/urlWriteback/yext.ts
mkilpatrick: Any invocations associated with an activity should show up.

packages/pages/src/plugins/index.ts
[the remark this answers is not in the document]
mkilpatrick: Oic. My other comments was wrong about the import now thinking about it. This will never be imported as an npm module. Given that, I think we can make it its own workspace. It still requires a package.json but it will only be published to deno, not npm.

Unclear if we'll want a whole deno release mechanism for it, but we can figure that out later.

packages/pages/src/plugins/urlWriteback/yext.ts
tatimblin: Very cool that's where I was hoping it would go

packages/pages/package.json
```diff
@@ -13,7 +13,7 @@
     ".": "./dist/index.js",
     "./vite-plugin": "./dist/vite-plugin/plugin.js",
     "./util": "./dist/util/index.js",
-    "./components": "./dist/components/index.js"
+    "./components": "./dist/components/index.js",
```
mkilpatrick, line 16: The final comma here is invalid json

packages/plugins/urlWriteback/README.md
```diff
@@ -0,0 +1,4 @@
+# urlWriteback Plugin
+
+This directory stores the deno function used as an onUrlChange event handler to write production
+urls back into the knowledge graph so that they can be used for listings.
```
mkilpatrick, line 4: Capitalize "knowledge graph" and "listings"

packages/plugins/urlWriteback/index.ts
```diff
@@ -0,0 +1,45 @@
+import { updateEntity } from "./yext.ts";
+
+/**
+ * The shape of the configuration options for the writeback plugin
```
mkilpatrick, line 4: Can you add periods to all of these comment sentences?

packages/plugins/urlWriteback/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  url: string;
+  entityId: string;
+  feature: string;
+  locale: string;
+}
+
+/**
+ * The urlWriteback function returns a configured function that can be used
+ * as a handler for the publishing system onUrlChange event in order to 
+ * save published urls back into the knowledge graph
```
mkilpatrick, line 26: Caps for KG again

packages/plugins/urlWriteback/index.ts
```diff
@@ -0,0 +1,45 @@
+import { updateEntity } from "./yext.ts";
+
+/**
+ * The shape of the configuration options for the writeback plugin
+ */
+export interface UrlWritebackConfig {
+  field: string
```
mkilpatrick, line 7: Colons. The formatter and linter should catch these things. You might need to rebase if you haven't in awhile.

mkilpatrick (approved):

PR #204, opened 2022-09-07
chore: swap changelog generation to generate-changelog

mkilpatrick:
This swaps from using `conventional-changelog-cli` to `generate-changelog` since the former had issues with rebases.


.eslintignore
```diff
@@ -0,0 +1 @@
+scripts/
```
mancusi, line 1: I think we should continue to lint all our JS/TS even in the scripts directory.

scripts/release.ts
```diff
@@ -81,16 +82,13 @@ step("\nUpdating package version...");
...
-  "-p",
-  "angular",
-  "-i",
-  "CHANGELOG.md",
-  "-s",
-  "--commit-path",
-  ".",
-];
+const latestTag = await getLatestTag(pkgName);
+if (!latestTag) process.exit();
```
mancusi, line 86: wrap with curly braces

scripts/release.ts
```diff
@@ -81,16 +82,13 @@ step("\nUpdating package version...");
...
+const sha = await run("git", ["rev-list", "-n", "1", latestTag], {
+  stdio: "pipe",
+}).then((res) => res.stdout.trim());
```
mancusi, line 89: if you're awaiting then you don't need to use a `.then`

mancusi (changes requested):

scripts/release.ts
mkilpatrick: Need to do this since the await returns ExecReturnValue<string>

scripts/releaseUtils.ts
```diff
@@ -100,7 +100,7 @@ export function getVersionChoices(currentVersion: string): VersionChoice[] {
   const isStable = !currentBeta && !currentAlpha;
 
   function inc(i: ReleaseType, tag = currentAlpha ? "alpha" : "beta") {
-    return semver.inc(currentVersion, i, tag)!;
+    return semver.inc(currentVersion, i, tag) || "0";
```
mancusi, line 103: what was the reason for this change?

We definitely want to throw an error if this lib doesn't return a result we'd expect.

mancusi (changes requested):

scripts/releaseUtils.ts
mkilpatrick: lint warning - updated to throw an error

mancusi (approved):

PR #210, opened 2022-09-11
feat(pages-plugins): add slug manager plugin

bhainesva:
TEST=manual
  tested in a demo account. confirmed that webhook and connector functions
  successfully updated primary and alternate language profiles

### DIFF
--- a/packages/plugins/index.ts
+++ b/packages/plugins/index.ts
@@ -1,2 +1,4 @@
 import urlWritebackPlugin from "./urlWriteback/index.ts";
-export { urlWritebackPlugin };
+import createSlugManager from "./slugManager/mod.ts";
+
+export { urlWritebackPlugin, createSlugManager };
--- a/packages/plugins/slugManager/manager.ts
+++ b/packages/plugins/slugManager/manager.ts
@@ -11,7 +11,13 @@ export interface InternalSlugManagerConfig {
   entityTypes?: string[];
   slugField?: string;
   slugFormatLocaleOverrides?: Record<string, string>;
-  api: Pick<IAPI, "savedSearchesIncludeEntity" | "updateField" | "listLanguageProfiles" | "listEntities">
+  api: Pick<
+    IAPI,
+    | "savedSearchesIncludeEntity"
+    | "updateField"
+    | "listLanguageProfiles"
+    | "listEntities"
+  >;
 }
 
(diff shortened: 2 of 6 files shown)


packages/plugins/index.ts
```diff
@@ -1,2 +1,4 @@
 import urlWritebackPlugin from "./urlWriteback/index.ts";
-export { urlWritebackPlugin };
+import createSlugManager from "./slugManager/mod.ts";
```
mkilpatrick, line 2: Why is the previous plugin's entrypoint index.ts and this one is mod.ts?

packages/plugins/slugManager/manager.ts
```diff
@@ -0,0 +1,158 @@
...
+    | "updateField"
+    | "listLanguageProfiles"
+    | "listEntities"
+  >;
+}
+
+export function createManager(config: InternalSlugManagerConfig) {
+  const {
+    slugFormatLocaleOverrides = {},
+    searchIds = [],
```
mkilpatrick, line 26: Will there be cases where people don't want to use a savedSearch but instead use the other scope types (entityType)?

packages/plugins/index.ts
bhainesva: I learned that `mod.ts` is the recommended thing: https://deno.land/manual@v1.25.2/contributing/style_guide#do-not-use-the-filename-indextsindexjs. 

I think we should update the other one too, was just going to make that a separate pr.

packages/plugins/slugManager/manager.ts
bhainesva: yes, the config also takes an `entityTypes` array: https://github.com/yext/pages/pull/210/files/d11d75dc9db5df837c0c8cd7377c4f6903034507#diff-9ebad0d98cc28c946d8f88e9fbc7b9b14aef410f3ea030940fa6e35e89fcb1cfR28

bhainesva: I didn't format them as `scopes` because we aren't using streams here, it's using kapi which doesn't talk about filters the same way.

packages/plugins/index.ts
mkilpatrick: Sounds good

mkilpatrick (approved):

PR #217, opened 2022-09-13
chore(components): upgrade storybook to use vite

tatimblin:
This PR updates storybook configuration to support two things:

1) Temporarily allow unsupported OpenSSL commands in node until Storybook version 7 is released and webpack version is upgraded there. This will allow us to run Storybook on the same node version as the rest of the package.
2) Switch underlying builder from webpack to vite. This will allow us to import our vite configuration into storybook, so Storybook can more closely mimic full builds


mkilpatrick (approved):

tatimblin:
@mkilpatrick readded a proxy loader that prevents non js imports from failing in jest environment, that I mistakenly thought was used for the same purpose in storybook.

mkilpatrick (approved):

Issue #218, opened 2022-09-14
default export should not be required in templates


tatimblin:
### Describe the bug

I am trying to add a robots.txt to a site following [this guide](https://hitchhikers.yext.com/docs/pages/robots.txt/). If I paste the sample into my project, build the site, and then try and serve I get an error that the template is missing a default export.

```
Error: Default export missing in template: /src/templates/robots.ts
```

~~I am, able to get around this issue by adding a default export like so~~ Doing this just uses the html wrappers and turns the should be txt file into an html file.

```
export default render;
```

For more context, I am not able to have the default export be any of the following i.e.

```
export default undefined;

export default {};

export default () => {};
```

It must be a function that returns a string, and that string will take precedence over the render function.

### Reproduction

https://github.com/yext-pages/Master-Ria

### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 681.34 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 18.4.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.0 - ~/repo/Master-Ria/node_modules/.bin/npm
  Browsers:
    Chrome: 105.0.5195.102
    Firefox: 100.0
    Safari: 15.6.1
  npmPackages:
    @yext/pages: 1.0.0-beta.6 => 1.0.0-beta.6
```

### Used Package Manager

pnpm

### Used Browser

Chrome

### Validations

- [X] Check that there isn't [already an issue](https://github.com/yext/pages/issues) that reports the same bug to avoid creating a duplicate.

bhainesva:
I'm still seeing this issue trying to add a robots.txt file on beta-9. For example here: https://github.com/yextconsulting/site-starter-react-consulting.

It works when running `yarn dev`, but `yarn build` gives me this:
```
'default' is not exported by src/templates/robots.tsx, imported by dist/hydration_templates/robots.tsx
file: /Users/bhaines/repo/site-starter-react-basic/dist/hydration_templates/robots.tsx:3:7
1: import * as React from "react";
2: import * as ReactDOM from "react-dom";
3: import Page from "../../src/templates/robots";
          ^
4: 
5: const data = (window as any).__INITIAL__DATA__;
error during build:
[...]

Issue #211, opened 2022-09-12
React root element id changes when site is deployed


bhainesva:
### Describe the bug

Hello. When I serve my site locally with `yarn dev` I see that the root element has an id of `root`:
<img width="1904" alt="image" src="https://user-images.githubusercontent.com/6308571/189656276-f2b396d7-9084-48ea-a122-b099fad374ee.png">

But when I deploy the site the root element's id is `reactele`:
<img width="1904" alt="image" src="https://user-images.githubusercontent.com/6308571/189656434-26f6e02a-0ee6-4919-a6bb-2191f4c9e7aa.png"> 

### Reproduction

https://sandbox.yext.com/s/3178323/yextsites/123470/branch/1661/deploys/recent?b=3178323

### System Info

```shell
System:
    OS: macOS 12.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 186.52 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.4.0 - ~/.nvm/versions/node/v17.4.0/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v17.4.0/bin/npm
  Browsers:
    Chrome: 105.0.5195.102
    Firefox: 104.0.2
    Safari: 15.4
  npmPackages:
    @yext/pages: 1.0.0-beta.6 => 1.0.0-beta.6
```

### Used Package Manager

yarn

### Used Browser

Firefox

### Validations

- [X] Check that there isn't [already an issue](https://github.com/yext/pages/issues) that reports the same bug to avoid creating a duplicate.

PR #219, opened 2022-09-14
feat(dev): allow render-only templates to work

mkilpatrick:
Fixes https://github.com/yext/pages/issues/218 https://github.com/yext/pages/issues/211


packages/pages/src/dev/client/entry-client.tsx
```diff
@@ -36,15 +36,18 @@ const hydrate = async () => {
...
+    console.error(
+      "Either a default export or render function is required in template: " +
+        template.path
+    );
     return;
   }
 
   ReactDOM.hydrate(
-    <Component {...(window as any)._RSS_PROPS_} />,
+    render ? render() : <Component {...(window as any)._RSS_PROPS_} />,
```
mancusi, line 50: render functions expect TemplateRenderProps ([src](https://github.com/yext/pages/blob/main/packages/pages/src/common/src/template/types.ts#L24))  same as Component functions exported from template modules.

packages/pages/src/dev/client/entry-client.tsx
```diff
@@ -36,15 +36,18 @@ const hydrate = async () => {
...
-    console.error("Default export missing in template: " + template.path);
+  if (!Component && !render) {
+    console.error(
+      "Either a default export or render function is required in template: " +
+        template.path
+    );
     return;
   }
 
   ReactDOM.hydrate(
```
mancusi, line 49: `render` doesn't return a React.Element. It doesn't need to be hydrated. It is supposed to represent the whole page.

mancusi: I don't think anything needs to be rendered client-side

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -57,15 +57,16 @@ export const serverRenderRoute =
...
+        });
 
       // render the component to its html
       // Since we are on the server using plain TS, and outside
       // of Vite, we are not using JSX here
-      const appHtml = await ReactDOMServer.renderToString(
-        React.createElement(Component, props)
-      );
+      const appHtml = render
+        ? render(props)
```
mancusi, line 66: similar comment to entry but the expected return of the render function is the whole file -- not just the html body like the default export

packages/pages/src/dev/server/public/templateBase.html
```diff
@@ -7,7 +7,7 @@
       import ReactDOM from "react-dom";
       import React from "react";
       const hydrate = async () => {
-        const templates = import.meta.glob("/src/templates/*.(jsx|tsx)");
+        const templates = import.meta.glob("/src/templates/*.(jsx|tsx|ts|js)");
```
mancusi, line 10: similar comments as above for the changes here.

mancusi (changes requested):

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -57,15 +57,19 @@ export const serverRenderRoute =
           featuresConfig,
           dynamicGenerateData,
           projectStructure,
-        }
-      );
+        });
+
+      if (render) {
+        res.status(200).set({ "Content-Type": "text/html" }).end(render(props));
```
mancusi, line 63: the content type isn't necessarily text/html.

In the issue, they were trying to render robots.txt

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -57,15 +57,19 @@ export const serverRenderRoute =
...
+        return;
+      }
 
       // render the component to its html
       // Since we are on the server using plain TS, and outside
       // of Vite, we are not using JSX here
-      const appHtml = await ReactDOMServer.renderToString(
-        React.createElement(Component, props)
-      );
+      const appHtml = Component
```
mancusi, line 70: before this do:
```
if (!Component) {
 // throw error
}
```

then you don't need to switch on component.

packages/pages/src/dev/server/ssr/pageLoader.ts
```diff
@@ -117,5 +108,5 @@ export const pageLoader = async ({
     relativePrefixToRoot: getRelativePrefixToRootFromPath(path),
   };
 
-  return { template, Component, props: templateRenderProps };
+  return { template, Component, render, props: templateRenderProps };
```
mancusi, line 111: would it be better to just return the `TemplateModuleInternal` rather than pieces of it?

packages/pages/src/dev/server/public/templateBase.html
```diff
@@ -26,16 +26,22 @@
             throw new Error("Function not implemented.");
           },
         };
-        const { default: Component } = await template.getComponent();
-        if (!Component) {
-          console.error("Default export missing in template: " + template.path);
+        const { default: Component, render } = await template.getComponent();
+        if (!Component && !render) {
```
mancusi, line 30: you can revert this as it's no longer needed.

packages/pages/src/dev/client/entry-client.tsx
```diff
@@ -36,16 +36,23 @@ const hydrate = async () => {
     },
   };
 
-  const { default: Component } = await template.getComponent();
+  const { default: Component, render } = await template.getComponent();
 
-  if (!Component) {
-    console.error("Default export missing in template: " + template.path);
+  if (!Component && !render) {
```
mancusi, line 41: you can revert this as it's no longer needed

mancusi (changes requested):

packages/pages/src/dev/server/ssr/pageLoader.ts
mkilpatrick: It's already defined in the caller so I can just use it from there and not return as many things.

packages/pages/src/dev/client/entry-client.tsx
```diff
@@ -36,16 +36,23 @@ const hydrate = async () => {
     },
   };
 
-  const { default: Component } = await template.getComponent();
+  const { default: Component, render } = await template.getComponent();
```
mancusi, line 39: this whole function is called "hydrate" Why do we even need to check the render and optionally return?

I'd hope this code path wouldn't get hit at all when just render is defined.

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -47,24 +48,36 @@ export const serverRenderRoute =
...
+        locale,
+        featuresConfig,
+        dynamicGenerateData,
+        projectStructure,
+      });
+
+      if (templateModuleInternal.render) {
+        res
+          .status(200)
+          .set({ "Content-Type": "text/html" })
```
mancusi, line 65: maybe we use https://expressjs.com/en/api.html#res.type and get the extension from the path?

Falling back to text/plain otherwise.

packages/pages/src/dev/server/public/templateBase.html
```diff
@@ -26,16 +26,22 @@
             throw new Error("Function not implemented.");
           },
         };
-        const { default: Component } = await template.getComponent();
+        const { default: Component, render } = await template.getComponent();
```
mancusi, line 29: getComponent renders the template module. Why not just have it return the default export if it exists and throw an error otherwise?

You shouldn't call get commit unless you're confident the template has a component.

mancusi (changes requested):

mancusi (approved):

tatimblin:
@mkilpatrick Can you release a new version with this. I'd like to setup robots.txt for a site ready to launch soon

PR #224, opened 2022-09-20
chore(playground): add playground and initial playwright tests

mkilpatrick:


playground/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
tatimblin, line 1: Can this be [inherited](https://www.typescriptlang.org/tsconfig#extends) from a base config up a directory?

tatimblin (commented):
The config and starter site look fine. For naming 'playground' sounds like we can freely edit without worrying, but I think that would make tests fail? and 'react' means that there would be other non react test sites.

playground/tsconfig.json
mkilpatrick: Probably

mkilpatrick:
> The config and starter site look fine. For naming 'playground' sounds like we can freely edit without worrying, but I think that would make tests fail? and 'react' means that there would be other non react test sites.

Might want to rename to pages-starter-react-locations

oshi97:
Closing this PR since we've merged in a playwright tests in a different PR, will keep the branch around though

PR #230, opened 2022-09-28
fix: features command when using top-level await

mkilpatrick:
When using a top-level await to dynamically import a module in a template, `npm run features` was failing with cjs errors. This was due to [module-from-string](https://www.npmjs.com/package/module-from-string) requiring `--experimental-vm-modules` in order to work properly.


mancusi:
Noting that this relies on a flag which could be altered or removed in a non-semantic way. The surface area of a node change that would prevent this fix from working is small so I think it's alright to send out.

mancusi (approved):

PR #235, opened 2022-10-03
perf(plugin): dedupe the css file list to avoid unnecessary imports

mkilpatrick:
Before this change, the list of css files was an array, resulting in **many** of the same file being imported over and over. This changes to use a set so that each file is only imported once.


packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
```diff
@@ -64,12 +63,12 @@ const getCssTags = (
   const [file, info] = entry;
 
   seen.add(file);
-  const cssFiles = info.css || [];
+  const cssFiles = new Set(info.css) || [];
```
mancusi, line 66: this line currently returns a set or an array. make this ways be a set.

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
```diff
@@ -64,12 +63,12 @@ const getCssTags = (
...
   (info.imports || [])
     .flatMap((f) => getCssTags(f, manifest, seen))
-    .forEach((f) => cssFiles.push(f));
+    .forEach((f) => cssFiles.add(f));
 
-  return cssFiles;
+  return Array.from(cssFiles);
```
mancusi, line 71: just continue to return a set and do a `cssFiles.forEach` to append them to the head.

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
```diff
@@ -56,18 +57,20 @@ const getCssTags = (
...
   }
   const [file, info] = entry;
 
   seen.add(file);
-  const cssFiles = info.css || [];
+  const cssFiles = new Set(info.css);
   (info.imports || [])
     .flatMap((f) => getCssTags(f, manifest, seen))
-    .forEach((f) => cssFiles.push(f));
+    .map((f) => f.forEach((file) => cssFiles.add(file)));
```
nbramblett, line 73: this nested map seems unnecessary since the above line is flat-mapping? If it is necessary then the outer call should also be .forEach, not map.

mkilpatrick: flatmap doesn't work on sets. changed it to be cleaner

nbramblett (approved):

mancusi (approved):

PR #238, opened 2022-10-05
docs(components): add link stories

### LINKED ISSUE #222: Storybook stories missing from some components
### Describe the bug

At the time of writing, the following components are missing Storybook stories:

- address
- analytics
- link

### Reproduction

_No response_

### System Info

```shell
N/A
```

### Used Package Manager

pnpm

### Used Browser

Chrome

### Validations

- [x] Check that there isn't [already an issue](https://github.com/yext/pages/issues) that reports the same bug to avoid creating a duplicate.

tatimblin:
fix #222 
add stories for various states of the Link component like an Email, Phone number, CTA, or URL

J=SLAP-2399
TEST=manual

logged the contexts to the actions pane


mkilpatrick:
Can you also mention the Link component in the PR title?

Also, is this good without the analytics console logging we talked about?

tatimblin:
I updated the description to include more information. I did not include the HTTP request parts of the component as it would require more work to the component. I think it would eventually be nice to have to improve visibility in how data looks (in a UI), but the jest tests McGin wrote cover everything already.

packages/pages/src/components/link/link.tsx
```diff
@@ -23,15 +23,15 @@ interface LinkConfig
 /**
  * The shape of the data passed to {@link Link} when directly passing an HREF to the Link component.
  */
-interface HREFLinkProps extends LinkConfig {
+export interface HREFLinkProps extends LinkConfig {
```
mkilpatrick, line 26: Why do these need to be exported? I don't see them used in the story.

packages/pages/src/util/runtime.ts
```diff
@@ -11,7 +11,7 @@ class Runtime {
   constructor() {
     if (typeof process !== "undefined") {
       this.name = "node";
-      this.version = process.versions.node;
+      this.version = process.versions?.node;
```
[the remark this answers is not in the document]
tatimblin: It did not, but I no longer need getRuntime in storybook

packages/pages/src/components/link/link.tsx
```diff
@@ -87,7 +87,7 @@ function isHREFProps(props: LinkProps): props is HREFLinkProps {
  */
 export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
   function Link(props, ref) {
-    const { children, onClick, className, eventName, ...rest } = props;
+    const { children, onClick, className, eventName, cta, ...rest } = props;
```
nmanu1, line 90: Can the uses of `props.cta` below be replaced with `cta`?

packages/pages/src/components/link/link.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+HREF.args = {
+  href: "https://yext.com",
+  children: "Learn More",
+};
+
+// CTA
+export const CTA = Template.bind({});
+
+CTA.args = {
```
nmanu1, line 23: For my own understanding, when specifying args like this, would we still be able to override these props via controls in the UI?

packages/pages/src/components/link/link.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Link } from "./link.js";
+
+export default {
+  title: "components/Link",
+  component: Link,
+} as ComponentMeta<typeof Link>;
+
+const Template: ComponentStory<typeof Link> = (args) => <Link {...args} />;
+
+// HREF
```
nmanu1, line 12: nit: are these inline comments necessary? they seem redundant since they each match their story name

packages/pages/src/components/link/link.tsx
tatimblin: This is an [issue in storybook](https://github.com/storybookjs/storybook/issues/12064). I could also duplicate the type in the storybook file. [This](https://stackoverflow.com/a/63138461/5627731) provides some more info into the error I believe, removing declarations in tsconfig would also resolve the error

packages/pages/src/components/link/methods.ts
```diff
@@ -20,27 +20,6 @@ const getHref = (cta: CTA): string => {
...
-/**
- * CTA constructor
- *
- * @param {string} link
- * @param {LinkType} linkType
- * @param {string} label
- *
- * @returns {boolean}
- */
-export function cta(
```
nmanu1, line 32: is this constructor not needed anymore?

tatimblin: @nmanu1 I replaced it with `makeCTA` inside the component. One so the name doesn't conflict and two it now takes the LinkProps so theres no real value for it existing outside the react component

nmanu1: I see. but this wasn't used before the refactor either from what I can see. was it just not being used anywhere? and it's only available internally (not publicly exported), right?

packages/pages/src/components/link/link.tsx
```diff
@@ -87,10 +87,13 @@ function isHREFProps(props: LinkProps): props is HREFLinkProps {
  */
 export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(
   function Link(props, ref) {
-    const { children, onClick, className, eventName, ...rest } = props;
-    const link: CTA = isHREFProps(props) ? { link: props.href } : props.cta;
+    const makeCTA = (props: LinkProps): CTA =>
```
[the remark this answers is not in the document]
tatimblin: Done. I have to set the cta as `CTA` because it could optionally be `Omit<CTA, "label">`. (`label` is optional in `CTA`)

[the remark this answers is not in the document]
nmanu1: I think if a type is defined for `link` (i.e. `const link: CTA = ...`), the assertion won't be needed anymore? generally, I'd prefer not to have assertions if there's another way to get the typing resolved

packages/pages/src/components/link/methods.ts
tatimblin: It was exported to be used outside of the component but now it is done in the component. I also didn't like the name of the function being the same as the type

packages/pages/src/components/link/link.tsx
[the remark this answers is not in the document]
tatimblin: yep updated

packages/pages/src/components/analytics/analytics.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { action } from "@storybook/addon-actions";
+import {
+  AnalyticsProvider,
+  AnalyticsScopeProvider,
+  AnalyticsContext,
+  ScopeContext,
+} from "./index.js";
+
+export default {
```
yen-tt, line 11: instead of doing a const assertion, can this be updated to have `const meta: ComponentMeta<typeof AnalyticsProvider>` and `export default meta` instead? It provides better typing this way

packages/pages/src/components/analytics/analytics.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+            mode: "development",
+          },
+        }}
+        enableDebugging={true}
+      >
+        <Story />
+      </AnalyticsProvider>
+    ),
+  ],
+} as ComponentMeta<typeof AnalyticsProvider>;
```
yen-tt, line 32: instead of doing a const assertion, can this be updated to have `const meta: ComponentMeta<typeof AnalyticsProvider>` and `export default meta` instead? It provides better typing this way

packages/pages/src/components/analytics/analytics.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * See the context object on the analytics wrapper component
+ * @returns
```
yen-tt, line 36: nit: clarify what is being returned or remove @returns here and below

packages/pages/src/components/link/link.tsx
yen-tt, line 26: should these new export interfaces be marked with `@public` as well? and perhaps moved to `types.ts` in link/ so they can be exported through `index.ts`

packages/pages/src/components/analytics/analytics.stories.tsx
tatimblin: removed analytics stories per @tmeyer2115 suggestion

packages/pages/src/components/analytics/analytics.stories.tsx
tatimblin: removed analytics stories per @tmeyer2115 suggestion

packages/pages/src/components/analytics/scope.tsx
```diff
@@ -2,7 +2,7 @@ import React, { PropsWithChildren } from "react";
 import { concatScopes } from "./helpers.js";
 import { AnalyticsScopeProps } from "./interfaces.js";
 
-const ScopeContext = React.createContext({ name: "" });
+export const ScopeContext = React.createContext({ name: "" });
```
nmanu1, line 5: does this need to be exported anymore?

packages/pages/src/components/link/types.ts
```diff
@@ -9,16 +13,77 @@ export const LinkType = {
...
+  eventName?: string;
+  conversionDetails?: ConversionDetails | undefined;
+}
+
+/**
+ * The shape of the data passed to {@link Link} when directly passing an HREF to the Link component.
+ *
+ * @public
+ */
+export interface HREFLinkProps extends LinkConfig {
```
nmanu1, line 52: if we're exporting these interfaces publicly, we probably also want to export `LinkConfig` so people can get the complete type, right?

packages/pages/src/components/link/types.ts
```diff
@@ -9,16 +13,77 @@ export const LinkType = {
...
+  > {
+  obfuscate?: boolean;
+  eventName?: string;
+  conversionDetails?: ConversionDetails | undefined;
+}
+
+/**
+ * The shape of the data passed to {@link Link} when directly passing an HREF to the Link component.
+ *
+ * @public
```
nmanu1, line 50: do we generate docs from these, e.g. via API extractor?

tatimblin: I believe it's in preparation. The extractor is set-up

nmanu1: do we need to run it then to extract the docs from these new imports?

tatimblin: no we don't have to do anything

packages/pages/src/components/link/types.ts
nmanu1: I meant in terms of documentation purposes, it would be more complete. if we were generating doc pages from these with API extractor, it would complain about `LinkConfig` not being exported. but, if we're not doing that then I guess it's not needed for now

nmanu1 (approved):

PR #239, opened 2022-10-06
feat(components): support pageDomain param

### LINKED ISSUE #229: Analytics: pageDomain param support
### Description

The pixel request can be passed a `pageDomain` query param. This param can be used in cases when the site isn't associated with a domain yet. Without it, it's not possible to fire analytics events before a real domain is created for the site. It would be nice if the provider had support for using this param.  

### Suggested solution

`providePagesAnalytics` could take a pageDomain option. We'd probably also need support from `yext/analytics`. 

### Alternative

Could manually make pixel requests, could create a real domain. 

### Additional context

https://yext.slack.com/archives/C02F556S09Y/p1664308802364519

Here's the place I was trying to record analytics events and ran into this issue of them not working: https://github.com/yext-pages/Guru/blob/f87e7f4baa9dba95f1ba3ed36ad740af0d83fed7/src/templates/index.tsx#L78. 

### Validations

- [X] Check that there isn't already an issue that requests the same feature to avoid creating a duplicate.

nmanu1:
Resolves #229. Bumps the version of `@yext/analytics` and adds support for passing in the `pageDomain` as a prop to the `AnalyticsProvider`, which is passed to `providePagesAnalytics`.

J=SLAP-2402
TEST=manual

Tested locally to make sure the `pageDomain` could be set from the `AnalyticsProvider` props and was only sent as a query param when formatted correctly (included the scheme). Otherwise, a console warning was logged and it was not sent in the analytics API requests.

### DIFF
--- a/packages/pages/src/components/analytics/provider.tsx
+++ b/packages/pages/src/components/analytics/provider.tsx
@@ -22,12 +22,17 @@ export function AnalyticsProvider(
     enableTrackingCookie,
     enableDebugging,
     templateData,
+    pageDomain,
   } = props;
 
   const analyticsRef = useRef<AnalyticsMethods | null>(null);
 
   if (analyticsRef.current === null) {
-    analyticsRef.current = new Analytics(templateData, requireOptIn);
+    analyticsRef.current = new Analytics(
+      templateData,
+      requireOptIn,
+      pageDomain
+    );
   }
 
   const analytics = analyticsRef.current;
@@ -38,7 +43,8 @@ export function AnalyticsProvider(
 
   let enableDebuggingDefault = debuggingParamDetected();
   if (typeof process !== "undefined") {
-    enableDebuggingDefault = enableDebuggingDefault || process.env?.NODE_ENV === "development";
+    enableDebuggingDefault =
+      enableDebuggingDefault || process.env?.NODE_ENV === "development";
   }
   analytics.setDebugEnabled(enableDebugging ?? enableDebuggingDefault);
 
--- a/packages/pages/THIRD-PARTY-NOTICES
+++ b/packages/pages/THIRD-PARTY-NOTICES
@@ -1,6 +1,6 @@
 The following NPM package may be included in this product:
 
- - @yext/analytics@0.2.2
+ - @yext/analytics@0.3.0
 
 This package contains the following license and notice below:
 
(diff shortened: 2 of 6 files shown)


packages/pages/src/components/analytics/provider.tsx
```diff
@@ -22,12 +22,17 @@ export function AnalyticsProvider(
...
   } = props;
 
   const analyticsRef = useRef<AnalyticsMethods | null>(null);
 
   if (analyticsRef.current === null) {
-    analyticsRef.current = new Analytics(templateData, requireOptIn);
+    analyticsRef.current = new Analytics(
+      templateData,
+      requireOptIn,
+      pageDomain
```
mkilpatrick, line 34: Per your comment
> If none is specified, the hostname for the site ID is used

does that mean the analytics library itself handles the logic and we're just passing this prop through to it? If so, do you know if it handles empty string in a particular way? I could see that leading to something unexpected.

packages/pages/src/components/analytics/provider.tsx
```diff
@@ -38,7 +43,8 @@ export function AnalyticsProvider(
 
   let enableDebuggingDefault = debuggingParamDetected();
   if (typeof process !== "undefined") {
-    enableDebuggingDefault = enableDebuggingDefault || process.env?.NODE_ENV === "development";
+    enableDebuggingDefault =
```
mkilpatrick, line 46: Wondering why this got reformatted. There's a commit hook to run prettier. So either it didn't work for you or it didn't work for Tom, which is weird.

packages/pages/src/components/analytics/provider.tsx
nmanu1: This is actually handled by the backend (Fusion). The request is still valid regardless of the string that's passed in (even invalid URL or empty string). But, it seems like the event would be logged as invalid if the `pageDomain` doesn't have a scheme (for `product=sites` only) and there is no hostname for the site ID.

Here we're passing straight to the analytics library and the analytics library checks that the string has a scheme before it sends it as a query param

packages/pages/src/components/analytics/provider.tsx
nmanu1: I actually ran prettier manually before committing, and it made this change

PR #245, opened 2022-10-18
fix(dev): parse multiple stream documents

### LINKED ISSUE #228: alternateLanguageFields data not available in dynamic mode
### Describe the bug

I have added
```
alternateLanguageFields: [
  "slug",
],
```
to my stream and regenerated my features.json, following the steps outlined [here](https://www.notion.so/Advanced-Multi-Lang-Support-93db0ef3a01f4536ab6f5a4378b59561). Next I re-served my site and expected to be able to interact with the new data in the document object
```
"alternateLanguageFields": {
	"es_US": {
		"slug": "es_US/123-main-st"
	}
},
```

This data however was not available when serving in dynamic mode, I was able to serve with local data as expected though.

### Reproduction

https://github.com/yext-pages/Master-Ria/tree/dev/trunk checkout dev/trunk and then add the first code block to the stream

### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 177.05 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 18.4.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.12.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 105.0.5195.125
    Firefox: 104.0.2
    Safari: 15.6.1
  npmPackages:
    @yext/pages: 1.0.0-beta.6 => 1.0.0-beta.6
```

### Used Package Manager

pnpm

### Used Browser

Chrome

### Validations

tatimblin:
The `alternateLanguageFields` feature requires multiple documents to be queried, so the locale cannot be included when using that feature. Additionally since the `alternateLanguageFields` feature uses information from multiple documents (languages), multiple documents will be returned. `parseConcatenatedDocuments()` will transform the output of multiple consecutive documents into an array so that it can be searched by locale for the proper document.

fix #228

J=SLAP-2401
TEST=auto,manual

successfully served multi-lang site with alternateLanguageFields feature enabled and disabled

### DIFF
--- a/packages/pages/src/dev/server/ssr/generateTestData.ts
+++ b/packages/pages/src/dev/server/ssr/generateTestData.ts
@@ -128,11 +128,11 @@ export const generateTestDataForPage = async (
       let parsedData: any;
       if (testData) {
         try {
-          const documents = splitUnsafeDocuments(testData.trim());
+          const documents = parseConcatenatedDocuments(testData.trim());
           if (documents.length === 1) {
             parsedData = documents[0];
           } else {
-            parsedData = getDocumentByLocale(parsedData, locale);
+            parsedData = getDocumentByLocale(documents, locale);
           }
         } catch (e) {
           stdout.write(
@@ -181,9 +181,9 @@ const prepareJsonForCmd = (json: any) => {
 };
 
 // There is a chance multiple documents will be returned consecutively, without proper containment. Look for them and parse as an array
-const splitUnsafeDocuments = (documentResponse: string): any[] => {
-  const DOCUMENT_SEPARATOR = "}\n{";
-  const DOCUMENT_JOINER = "},\n{";
+const parseConcatenatedDocuments = (documentResponse: string): any[] => {
+  const DOCUMENT_SEPARATOR = `}\n{\n  "__`;
+  const DOCUMENT_JOINER = `},\n{\n  "__`;
   const objects = documentResponse
     .split(DOCUMENT_SEPARATOR)
     .join(DOCUMENT_JOINER);
--- a/packages/pages/src/dev/server/ssr/generateTestData.ts
+++ b/packages/pages/src/dev/server/ssr/generateTestData.ts
@@ -180,7 +180,8 @@ const prepareJsonForCmd = (json: any) => {
   return jsonString;
 };
 
-// There is a chance multiple documents will be returned consecutively, without proper containment. Look for them and parse as an array
+// When using a CLI version <= 0.1_299 there is a chance multiple documents will be returned consecutively, without proper containment.
+// Look for them and parse as an array. This can be removed once the CLI handles multiple documents as an array https://yexttest.atlassian.net/browse/SPR-4712
 const parseConcatenatedDocuments = (documentResponse: string): any[] => {
   const DOCUMENT_SEPARATOR = `}\n{\n  "__`;
   const DOCUMENT_JOINER = `},\n{\n  "__`;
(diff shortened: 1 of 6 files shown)


packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -175,3 +179,17 @@ const prepareJsonForCmd = (json: any) => {
   }
   return jsonString;
 };
+
+// There is a chance multiple documents will be returned consecutively, without proper containment. Look for them and parse as an array
+const splitUnsafeDocuments = (documentResponse: string): any[] => {
+  const DOCUMENT_SEPARATOR = "}\n{";
```
tatimblin, line 185: This could incorrectly be triggered by a multi-line field value that contains this string

tatimblin: Here's what the start/end of a document looks like
```
	"yextRoutableCoordinate": {
		"latitude": 33.20214104581977,
		"longitude": -117.36811858975076
	}
}              <----- need to find this point
{
	"__": {
		"name": "robots_txt",
		"staticPage": {
			"plugin": {}
		},
		"templateType": "JS"
	},
	"_site": {
```
Because the `__` property is default, we can check for that, and since a field starting with three underscores is technically possible (though not created by the user via custom field). I should be able to check for just the opening of the underscores `"}\n{\n  "__`. This would greatly reduce the already small chance for an incorrect match, basically down to someone storing a stream response with the alternateLanguageFields feature in a multi-line text field.

yen-tt: It seems a somewhat odd that the data returned by invoking `yext sites generate-test-data [other flags]` would not include commas to separate multiple objects for a proper json format. Instead of having workarounds here trying to split the document by some specific regex and reconstruct the json object, is it possible to update `yext sites generate-test-data` to add commas between documents? -- I assume in `gocode/src/yext/cli/sites/generatetestdata/generatetestdata.go` somewhere, it's appending the documents together to send in one go

tatimblin: @yen-tt some further discussion on slack

yen-tt: So the backend will provide as a json array instead of the missing comma situation here. Definitely make this easier without worrying about edge cases. Let's update this pr after that work is done.

mancusi: good idea!

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -175,3 +179,17 @@ const prepareJsonForCmd = (json: any) => {
   }
   return jsonString;
 };
+
+// There is a chance multiple documents will be returned consecutively, without proper containment. Look for them and parse as an array
```
mancusi, line 183: can you make a note that versions of the CLI <= XX return documents without the containment?

Also just an improvement for the future but pinning certain versions of pagesjs to the CLI will be pretty necessary. Unfortunately the CLI isn't semantically versioned yet.

mancusi (dismissed):
minor comment otherwise lgtm

tatimblin:
Changed to merge into feature branch so this can be closed, and then updating with Spruce's fix to the CLI can be added before merging to main.

yen-tt (approved):

PR #246, opened 2022-10-20
feat(pages): multibrand setup support in dev and vite-plugin build

yen-tt:
This PR adds support for multibrand setup in pages dev and vite-plugin build by exposing a new optional field `domain` in 
`ProjectStructureConfig` to specify the domain of the site, aka the subfolder name inside template root folder and sites config root folder.

As outlined in the spec, this PR updated dev and build logic such that any template files defined in a matching domain subdirectory will be used instead of those with the same name found at the root. Any other additional template files in the template root folder will be shared across the domain specific subdirectory. `features.json` will be generated in root sites-config folder by default, or the domain specific sites-config folder if a domain is provided (will update the feature command in another item).

J=SLAP-2411
TEST=auto&manual

new jest tests for `getTemplatesFilepath` passed

created a [starter repo](https://github.com/yen-tt/pages-starter/tree/dev/test-multibrand) with multibrand setup (`locations.brand1.com` and `locations.brand2.com` folder in site-config and src/templates):
- run `pnpm pack` for yext/pages package with `domain: "locations.brand1.com",` in defaultConfig of `structure.ts`. ran `npm run dev:multibrand` in the repo, see from landing page that location and static page template inside brand1 folder is used AND random and robots template in root folder is used. All pages successfully loaded. ran `npm run build:multibrand` successfully and see the same result.
- run `pnpm pack` for yext/pages package without `domain` in defaultConfig of `structure.ts`. ran `npm run dev` in the repo, see that landing page display links to templates from root folder only, without templates in brand1 or brand2. All pages successfully loaded. ran `npm run build` successfully and see the same result.


packages/pages/src/vite-plugin/build/buildStart/buildStart.ts
```diff
@@ -6,6 +6,7 @@ import { fileURLToPath } from "url";
 import { PluginContext, EmitFile } from "rollup";
 import { generateHydrationEntryPoints } from "./hydration.js";
 import { ProjectStructure } from "../../../common/src/project/structure.js";
+import { getTemplatesFilepath } from "../../../common/src/template/internal/getTemplatesFilepath.js";
```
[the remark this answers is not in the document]
tatimblin: https://github.com/yext/pages/pull/199

packages/pages/src/dev/server/public/templateBase.html
```diff
@@ -4,24 +4,23 @@
   <body>
     <div id="reactele"><!--app-html--></div>
     <script type="module">
```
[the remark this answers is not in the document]
yen-tt: Yup, it's a copy paste! Noted in the pr description: changes in templateBase.html is copied over from running `pnpm pack` and use the content of the `compiled entry-client.js`. @mkilpatrick informed me we can't import entry-client.tsx directly because it doesn't work correctly for hmr. He opened a GH [issue](https://github.com/vitejs/vite/issues/9054) with Vite, which is not resolve yet. ([slack thread)](https://yext.slack.com/archives/C02LLE9BW2K/p1666193518715529?thread_ts=1666192140.834989&cid=C02LLE9BW2K)

packages/pages/src/dev/server/public/templateBase.html
```diff
@@ -4,24 +4,23 @@
...
-      import ReactDOM from "react-dom";
-      import React from "react";
+      import * as ReactDOM from "react-dom";
```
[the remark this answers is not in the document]
yen-tt: `allowSyntheticDefaultImports` is set to true to it should allow the old approach.. but this change was what generated for `entry-client.js` in dist.

packages/pages/src/common/src/template/internal/getTemplatesFilepath.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param templatesRoot  The folder path where all template files live
+ * @param templatesDomain The folder path where the template files of a specific domain live
+ * @returns a list of template filepaths
+ */
+export const getTemplatesFilepath = (
+  templatesRoot: Path,
+  templatesDomain?: Path
+): string[] => {
+  const templatesRootFilepath: string[] = glob.sync(
+    `${templatesRoot.getAbsolutePath()}/**/*.{tsx,jsx,js,ts}`
```
tatimblin, line 20: Would this get files in other domain folders?

I think this would only look in the current (root) directory
```suggestion
    `${templatesRoot.getAbsolutePath()}/*.{tsx,jsx,js,ts}`
```

packages/pages/src/common/src/template/internal/getTemplatesFilepath.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Path } from "../../project/structure.js";
+import { getTemplatesFilepath } from "./getTemplatesFilepath.js";
+
+const domainPath = "src/templates/some.domain.com";
+const rootPath = "src/templates";
+jest.mock("glob", () => {
+  return {
+    sync: (path: string) => {
+      return path.indexOf(domainPath) != -1
+        ? [`${domainPath}/brand.tsx`, `${domainPath}/test.tsx`]
```
tatimblin, line 11: Can you add a test with a second brand

yen-tt: I updated the mocked glob to include additional filepaths from another domain folder and use `minimatch` as a dev util package to convert the glob expression to regex to simulate the filepath matching process (to avoid having to manually construct the folder structure and files for the test).
This should help test that it shouldn't retrieve files from domain 2 if we specify domain 1. As such, there's no need for adding another test as it will be the same as the second test we have here.

packages/pages/src/dev/server/public/templateBase.html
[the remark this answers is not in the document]
oshi97: oh wait are we copy pasting the file from dist? (prob dumb question but what is the `dist` folder??)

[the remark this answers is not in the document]
yen-tt: all good, the `dist` folder is generated from command `pnpm run build`. That's how I get the compiled `entry-client.js` file from `packages/pages/src/dev/client/entry-client.tsx`. I don't see other command to specifically compile entry-client.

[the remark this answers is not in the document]
oshi97: Oh I see, I thought we were manually copy pasting it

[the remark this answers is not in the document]
yen-tt: oh we are, I ran the build command, then copy the compiled `entry-client.js` into `templateBase.html`. From this previous [pr comment](https://github.com/yext/pages/pull/18#discussion_r859813812), there's a sumo item to make automate it but it's not yet implemented.

packages/pages/src/common/src/template/internal/getTemplatesFilepath.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        `${domain2Path}/brand2.tsx`,
+        `${domain2Path}/test.tsx`,
+        `${rootPath}/share.tsx`,
+        `${rootPath}/test.tsx`,
+      ];
+      return filepaths.filter((f) => minimatch(path.resolve(f), glob));
+    },
+  };
+});
+afterAll(() => jest.unmock("glob"));
```
oshi97, line 24: Is this unmock necessary so it doesn't bleed into other tests? I thought that jest.mock would only last for the current test file but could be wrong

packages/pages/src/common/src/template/internal/getTemplatesFilepath.ts
```diff
@@ -0,0 +1,37 @@
+import glob from "glob";
+import path from "path";
+import { Path } from "../../project/structure.js";
```
oshi97, line 3: nit but would it make sense to move Path to its own file?

packages/pages/src/dev/server/ssr/featureNameToTemplateModuleInternal.ts
```diff
@@ -1,21 +1,24 @@
...
 } from "../../../common/src/template/internal/types.js";
+import { Path } from "../../../common/src/project/structure.js";
+import { getTemplatesFilepath } from "../../../common/src/template/internal/getTemplatesFilepath.js";
 
 // Determines the template module to load from a given feature name (from the exported config)
 export const featureNameToTemplateModuleInternal = async (
   devserver: ViteDevServer,
-  featureName: string
+  featureName: string,
+  templatesRoot: Path,
```
oshi97, line 14: would it makes things easier to unit test if this took in templatesFilepath as a parameter instead of templatesRoot + templatesDomain and then computing templatesFilepath?

packages/pages/src/dev/server/ssr/featureNameToTemplateModuleInternal.ts
```diff
@@ -1,21 +1,24 @@
...
+  templatesDomain?: Path
 ): Promise<TemplateModuleInternal<any, any> | null> => {
-  const directoryFilenames = await readdir(`./${TEMPLATE_PATH}`);
-
-  for (const filename of directoryFilenames) {
-    const templateFilepath = `${TEMPLATE_PATH}/${filename}`;
+  const templatesFilepath = getTemplatesFilepath(
```
oshi97, line 17: maybe `templateFilepaths`, it's multiple file paths right?

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -14,9 +14,12 @@ import { ProjectStructure } from "../../../common/src/project/structure.js";
...
+export const generateTestData = async (domain?: string): Promise<boolean> => {
   const command = "yext";
-  const args = ["sites", "generate-test-data"];
+  let args = ["sites", "generate-test-data"];
+  if (domain) {
+    args = args.concat("--hostname", domain);
+  }
 
   async function generate() {
     const childProcess = spawn(command, args);
```
oshi97, line 25: nit and clearly this was here before, but below I think it can just return `!!exitCode`. Also instead of 
```ts
return new Promise((resolve) => {
    resolve(generate());
  });
```
I think just  `return generate()` should be fine?

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -43,11 +46,12 @@ export const generateTestDataForPage = async (
...
 ): Promise<any> => {
+  const siteConfigPath =
+    projectStructure.sitesConfigDomain?.getAbsolutePath() ??
+    projectStructure.sitesConfigRoot.getAbsolutePath();
   const siteStreamPath = path.resolve(
     process.cwd(),
-    projectStructure.sitesConfigRoot.getAbsolutePath() +
-      "/" +
-      projectStructure.siteStreamConfig
+    siteConfigPath + "/" + projectStructure.siteStreamConfig
```
oshi97, line 54: do we want to use path.join here instead? sometimes other os (windows) are weird

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -45,31 +46,46 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
...
+  templateRootDir: string,
+  templatedomainDir: string | undefined,
   hydrationOutputDir: string
 ): Promise<InputOption> => {
-  return (await readdir(templateDir)).reduce(
-    (input: Record<any, any>, template) => {
-      const parsedPath = parse(template);
+  const entryPoints: Record<string, string> = {};
+  const updateInput = async (dir: string) =>
+    (await readdir(dir, { withFileTypes: true }))
```
oshi97, line 67: I want to say this await is unnecessary since readdir returns a promise?

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -45,31 +46,46 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
...
+        }
+        return input;
+      }, entryPoints);
 
-      input[`server/${parsedPath.name}`] = path.join(templateDir, template);
-      return input;
-    },
-    {}
-  );
+  if (templatedomainDir) {
```
oshi97, line 86: Would it be better to use Promise.all here since we're sometimes waiting on two separate promises?

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -45,31 +46,46 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
...
 ): Promise<InputOption> => {
-  return (await readdir(templateDir)).reduce(
-    (input: Record<any, any>, template) => {
-      const parsedPath = parse(template);
+  const entryPoints: Record<string, string> = {};
+  const updateInput = async (dir: string) =>
+    (await readdir(dir, { withFileTypes: true }))
+      .filter((dirent) => !dirent.isDirectory())
+      .map((file) => file.name)
+      .reduce((input, template) => {
```
oshi97, line 70: it feels weird to use a reduce here when we're not using the return value directly and are actually modifying the entryPoints object. Maybe a forEach if we want to do it this way?

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -45,31 +46,46 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
...
-  templateDir: string,
+  templateRootDir: string,
+  templatedomainDir: string | undefined,
   hydrationOutputDir: string
 ): Promise<InputOption> => {
-  return (await readdir(templateDir)).reduce(
-    (input: Record<any, any>, template) => {
-      const parsedPath = parse(template);
+  const entryPoints: Record<string, string> = {};
+  const updateInput = async (dir: string) =>
```
oshi97, line 66: what is the `Input` this is updating?

packages/pages/src/common/src/project/structure.ts
```diff
@@ -117,6 +127,18 @@ export class ProjectStructure {
       this.#config.filepathsConfig.sitesConfigRoot
     );
     this.templatesRoot = new Path(this.#config.filepathsConfig.templatesRoot);
+
+    const domain = this.#config.filepathsConfig.domain;
+    if (domain) {
+      this.domain = domain;
+      this.sitesConfigDomain = new Path(
```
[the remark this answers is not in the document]
yen-tt: it's just a wrapper class to provide useful methods to operate on the provided path, like `getRelativePath` and `getAbsolutePath`

packages/pages/src/common/src/template/internal/getTemplatesFilepath.test.ts
yen-tt: This was copied from `loader.test.ts`. I thought you would need `resetModules` / `resetMocks` in the jest config to have it auto unmock but it looks like it's fine without it! I cloned this test fie, removed the mock glob code, and it failed so this mock didn't bleed into other test. I will remove

packages/pages/src/common/src/template/internal/getTemplatesFilepath.ts
yen-tt: sure

packages/pages/src/vite-plugin/build/build.ts
yen-tt: sure

packages/pages/src/vite-plugin/build/build.ts
yen-tt: renamed to `updateEntryPoints` so it match with the variable name. Invoking this would update `entryPoints`, adding new filepaths or "input"(s) as rollup bundle's entry point(s)

packages/pages/src/common/src/project/path.ts
```diff
@@ -0,0 +1,22 @@
+import pathLib from "path";
+
+/**
+ * Provides useful methods to operate on a specific property of {@link ProjectFilepathsConfig}.
```
tmeyer2115, line 4: Is `ProjectFilepathsConfig` a type or interface? I can't seem to find it defined anywhere in the repo.

packages/pages/src/common/src/project/structure.ts
```diff
@@ -17,6 +18,12 @@ export interface ProjectFilepaths {
   hydrationBundleOutputRoot: string;
   /** The folder path where the compiled server bundles should go */
   serverBundleOutputRoot: string;
+  /**
+   * The domain of the site, which also serve as the subfolder name
+   * inside {@link templatesRoot} and {@link sitesConfigRoot}.
+   * This is used for the case of multibrand setup within a single repo.
```
tmeyer2115, line 24: I would put this sentence first. By having it first, people can just skip the rest of the description if they're not doing multi-brand.

packages/pages/src/common/src/project/structure.ts
```diff
@@ -102,6 +109,10 @@ export class ProjectStructure {
 
   sitesConfigRoot: Path;
   templatesRoot: Path;
+  domain?: string;
+  sitesConfigDomain?: Path;
```
tmeyer2115, line 113: I think we should spend more time thinking about the names `sitesConfigDomain` and `templatesDomain`. They feel a bit un-intuitive.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -0,0 +1,37 @@
...
+ * provided template folder paths, only the file in domain template
+ * folder path is collected.
+ *
+ * @param templatesRoot  The folder path where all template files live
+ * @param templatesDomain The folder path where the template files of a specific domain live
+ * @returns a list of template filepaths
+ */
+export const getTemplateFilepaths = (
+  templatesRoot: Path,
+  templatesDomain?: Path
```
tmeyer2115, line 17: Small nit: I think we can get rid of the `templates` prefix for both variables. It's implied by the method name.

packages/pages/src/common/src/project/path.ts
yen-tt: I assume `ProjectFilepathsConfig ` got rename to `ProjectFilepaths` interface and this comment was meant to point to that. I can update

mancusi: defined here https://github.com/yext/pages/blob/e58a8ec42258b56688b6ce39373f0d22dbe60c4c/packages/pages/src/common/src/project/structure.ts#L68

packages/pages/src/common/src/project/structure.ts
```diff
@@ -17,6 +18,12 @@ export interface ProjectFilepaths {
   hydrationBundleOutputRoot: string;
   /** The folder path where the compiled server bundles should go */
   serverBundleOutputRoot: string;
+  /**
+   * The domain of the site, which also serve as the subfolder name
+   * inside {@link templatesRoot} and {@link sitesConfigRoot}.
+   * This is used for the case of multibrand setup within a single repo.
+   */
+  domain?: string;
```
mancusi, line 27: Not sure exactly what the spec said here but I don't think we should use the word "domain" here. "scope" makes more sense to me as you're scoping a build to a subset of templates. What do you all think?

packages/pages/src/common/src/project/structure.ts
mancusi: adding onto this, can't sitesConfigDomain and templateDomain be resolved from the domain? if that's so then I think we should expose functions which abstract away how to handle the root + domain logic

yen-tt: the naming here is to follow the existing variable names `sitesConfigRoot` and `templatesRoot`. Alternatively, if we are fine with deviating from that, maybe `domainSitesConfigPath`?

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Get all the template files in the provided template folder path(s).
+ * If there are two files that share the same name between the two
+ * provided template folder paths, only the file in domain template
+ * folder path is collected.
+ *
+ * @param templatesRoot  The folder path where all template files live
+ * @param templatesDomain The folder path where the template files of a specific domain live
+ * @returns a list of template filepaths
+ */
+export const getTemplateFilepaths = (
```
mancusi: I don't think that getTemplateFilepaths should know about this scoping logic -- that should be relatively contained. What if it just took a list of paths, iterated through them in-order, and only kept the last seen templates?

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
```diff
@@ -46,7 +49,7 @@ export default (projectStructure: ProjectStructure) => {
 
     finisher = logger.timedLog({ startLog: "Writing manifest.json" });
     try {
-      await generateManifestFile(templateModules, projectStructure);
```
oshi97, line 49: Asking mostly out of curiosity but I thought typescript would have complained about this before?

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -14,26 +14,22 @@ import { ProjectStructure } from "../../../common/src/project/structure.js";
  * generateTestData will run yext sites generate-test-data and return true in
  * the event of a successful run and false in the event of a failure.
  */
-export const generateTestData = async (): Promise<boolean> => {
+export const generateTestData = async (domain?: string): Promise<boolean> => {
```
mancusi, line 17: add documentation for this new parameter

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
yen-tt: I didn't get an error, but I saw a white dotted underline with a message `'await' has no effect on the type of this expression.ts(80007)`.

yen-tt: just saw your edit, yea adding an eslint rule would be good (maybe this one: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/await-thenable.md)

mancusi (changes requested):
did you test deno generation as well? I see you updated the rendering package which is used during `yext pages generate` and not hit during just an `npm run build`

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -0,0 +1,34 @@
...
+ * If there are two files that share the same name between the two
+ * provided template folder paths, only the file in domain template
+ * folder path is collected.
+ *
+ * @param root  The folder path where all template files live
+ * @param domain The folder path where the template files of a specific domain live
+ * @returns a list of template filepaths
+ */
+export const getTemplateFilepaths = (root: Path, domain?: Path): string[] => {
+  const templatesRootFilepaths: string[] = glob.sync(
```
oshi97, line 16: would this function be easier to test if it didn't call glob internally but took `templatesRootFilepaths` and `templatesDomainFilepaths` as params or something? Then we wouldn't have to do write our own mock glob

oshi97: maybe that's a little tricky since domain can be undefined though...?

packages/pages/src/dev/client/entry-client.tsx
```diff
@@ -9,7 +9,7 @@ const hydrate = async () => {
   };
 
   // Can't use string interpolation here so src/templates is hardcoded
-  const templates = import.meta.glob("/src/templates/*.(jsx|tsx)");
+  const templates = import.meta.glob("/src/templates/**/*.(jsx|tsx)");
 
   const routes: Route[] = Object.keys(templates).map((path) => {
```
oshi97, line 14: this routes variable is a little weird, it looks like we only need the single template variable we get in the `routes.find` below? but maybe we can address this in the cleanup epic

yen-tt:
> did you test deno generation as well? I see you updated the rendering package which is used during `yext pages generate` and not hit during just an `npm run build`

Not sure if I understand the full process of testing deno generation, but I did `yext pages build && yext pages generate-test-data --hostname locations.brand1.com && yext pages render --hostname locations.brand1.com` and ran `yext pages serve`. That seems to work fine. Let me know if I should be testing another way or if I'm missing something!

mancusi:
> > did you test deno generation as well? I see you updated the rendering package which is used during `yext pages generate` and not hit during just an `npm run build`
> 
> Not sure if I understand the full process of testing deno generation, but I did `yext pages build && yext pages generate-test-data --hostname locations.brand1.com && yext pages render --hostname locations.brand1.com` and ran `yext pages serve`. That seems to work fine. Let me know if I should be testing another way or if I'm missing something!

yep that's exactly what I meant!

packages/pages/src/common/src/project/structure.ts
yen-tt: the two new variables can be derived from joining the root variables with domain, but since they can potentially be use in multiple places, I thought it would be good to construct the paths and store it without the need to construct it each time.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
yen-tt: Hmm, not sure how I feel about having the "order matter" logic when processing the list of paths.. could be a little confusing but that should be fine with documentation.
Just to get a better understanding, what is the reason you think it shouldn't know about about domain / root? Since it's a concept we are introducing, I think this help makes it more clear that domain files take precedent over root files, and any unique root files are share across domains.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
yen-tt: I kinda like that we are testing the glob pattern as well currently, to avoid mistakes with accidentally including subfolder/files or other non-template extensions?

packages/pages/src/dev/client/entry-client.tsx
yen-tt: yea I took a close look at the logic there, most of the code here is not really needed. `import.meta.glob("/src/templates/*.(jsx|tsx)");` would give use a `templates` object like this ([Vite doc](https://vitejs.dev/guide/features.html#glob-import)):
```
// code produced by vite
const modules = {
  '/src/templates//foo.tsx': () => import('./dir/foo.js'),
  '/src/templates//bar.tsx': () => import('./dir/bar.js')
}
```
since we already have the template path, we can just directly import that.

packages/pages/src/dev/client/entry-client.tsx
```diff
@@ -21,16 +21,14 @@ const hydrate = async () => {
...
-
-  const template = routes.find(
-    (route) => route.name === templateFilenameWithoutSuffix
+  const templatePath = (window as any)._RSS_TEMPLATE_PATH_;
+  const template = routes.find((route) =>
+    templatePath.endsWith(route.path)
   ) || {
-    name: templateFilename?.split(".")[0],
-    path: `/src/templates/${templateFilename}`,
+    name: templatePath.split("/").at(-1).split(".")[0],
```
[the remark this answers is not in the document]
yen-tt: updated the file so this is no longer needed

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
yen-tt: added to the doc for capturing pagesjs repo changes!

packages/pages/src/dev/client/entry-client.tsx
yen-tt: mm actually, I see that there's a limitation with vite and dynamic import:
Screen Shot 2022-10-21 at 1 54 51 PM
I will revert the changes, and perhaps we can address this in the cleanup epic

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
mancusi: My main goal is to limit how much other components needs to handle specific scoping logic. Ideally, it'd only be contained within the ProjectStructure.

I don't see why `getTemplateFilepaths` needs to have custom logic on whether a domain was set or not. Instead it can just accept an arbitrary number of paths with rules that match the business rules of how to determine which template is which.

packages/pages/src/dev/client/entry-client.tsx
mancusi: a little out of scope it'd be nice to get that cleaned up. right now the code is assuming `src/templates` is the root of where templates will be which prevents us from allowing that to be configurable.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -0,0 +1,34 @@
...
+    `${root.getAbsolutePath()}/*.{tsx,jsx,js,ts}`
+  );
+  if (!domain) {
+    return templatesRootFilepaths;
+  }
+  const templatesDomainFilepaths: string[] = glob.sync(
+    `${domain.getAbsolutePath()}/*.{tsx,jsx,js,ts}`
+  );
+  const templatesDomainFilenames = templatesDomainFilepaths.map((t) =>
```
oshi97, line 25: nit but does `templatesDomainFilepaths.map(path.basename)` work?

packages/pages/src/common/src/project/structure.ts
yen-tt: Hmm, we are not fully scoped to just the specified subset since we do use other files in root folder (e.g. a file with name exist only in root will be share across all brand folders, or something specified in a domain features.json but doesn't exist in that template subfolder would fallback to root)

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
yen-tt: nope, there's a conflict between the expected function signature of `.map` param and `path.basename`

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
yen-tt: fair point, I will update!

packages/pages/src/dev/client/entry-client.tsx
yen-tt: for sure! As this item is blocking other ones in the epic, I think we can make a separate item to address this existing bug.

packages/pages/src/common/src/project/structure.ts
yen-tt: updated the name to `scope`

packages/pages/src/common/src/project/structure.ts
yen-tt: updated the name to `scopedTemplatesPath` and `scopedsitesConfigPath`

packages/pages/src/common/src/project/structure.ts
```diff
@@ -110,9 +110,9 @@ export class ProjectStructure {
 
   sitesConfigRoot: Path;
   templatesRoot: Path;
-  domain?: string;
-  sitesConfigDomain?: Path;
-  templatesDomain?: Path;
+  scope?: string;
+  scopedTemplatesPath?: Path;
+  scopedsitesConfigPath?: Path;
```
mancusi, line 115:
```suggestion
  scopedSitesConfigPath?: Path;
```

mancusi (changes requested):
minor comment

mancusi (commented):
lgtm but will hold off  on approval so @mkilpatrick can review as well

packages/pages/src/common/src/template/internal/getTemplateFilepaths.test.ts
```diff
@@ -0,0 +1,48 @@
...
+        `${rootPath}/share.tsx`,
+        `${rootPath}/test.tsx`,
+      ];
+      return filepaths.filter((f) => minimatch(path.resolve(f), glob));
+    },
+  };
+});
+
+describe("getTemplateFilepaths", () => {
+  it("collects all template files from root folder path", () => {
```
mkilpatrick, line 26: I'm not sure what the spec defines, but are we now treating subfolders as scopes? This will break existing usages if someone is using subfolders and they're no longer globbed.

mkilpatrick: Oh interesting. I thought we were globbing everything in src/templates but it sounds like we're only globbing files at the root. If that's the case then your change isn't breaking.

yen-tt: the logic in main is globbing everything inside src/templates with subfolders files too. But that also include folder paths (vs file paths) which is currently not being handle properly.

packages/pages/src/dev/client/entry-client.tsx
yen-tt: https://yexttest.atlassian.net/browse/SLAP-2426

packages/pages/src/common/src/template/internal/getTemplateFilepaths.test.ts
yen-tt: [Confirmed with PM](https://yext.slack.com/archives/C02LLE9BW2K/p1666624017742579?thread_ts=1666623883.329669&cid=C02LLE9BW2K): we only use templates at the root level. The public documentation currently states to not put sub directories into the templates directory, so we are sticking with that paradigm for now

mkilpatrick: Sounds good

mkilpatrick (approved):

PR #248, opened 2022-10-25
feat(dev): add flags to features command for multibrand setup

yen-tt:
This PR updates pages `features` command to accept new argument `scope` to  use for multibrand setup. Also update the structure of features files to follow the `yargs` setup from previous pr.

J=SLAP-2424
TEST=manual

tested the following commands:
- `pages generate` - showed the sub commands
- `pages generate features` - generate features.json for `./src/templates` root level to `./sites-config` folder
- `pages generate features --scope locations.brand1.com` - generate features.json for `./src/templates/locations.brand1.com` to `./sites-config/locations.brand1.com` folder


packages/pages/src/generate/features/features.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+  await createFeaturesJson(templateModules, featuresFilepath);
+};
+
+export const featureCommandModule: CommandModule<unknown, FeaturesArgs> = {
+  command: "features",
+  describe: "Generates features.json file",
+  builder: (yargs) => {
+    return yargs
+      .option("templatesRoot", {
```
mkilpatrick, line 47: Are the two flags that aren't scope necessary? It could be nice but we're adding functionality we probably don't need.

yen-tt: I was thinking we do expose the config option `templatesRoot` and `sitesConfigRoot` in our vite-plugin, so it may be possible that people can configure their root file paths differently from the defaults, which would affect the filepaths in how we scope the build (in which case I will update `dev` command to include these two options too). If you think that's not super necessary still, I can remove it here

mkilpatrick: Makes sense, but in reality we should unexpose those in the plugin as it won't actually work. It's because you can't import templates with a variable prefix since Vite statically replaces the import.meta call.

So I think we should remove these options for now here and in the plugin. Better yet, don't allow any options in the plugin.

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
```diff
@@ -9,7 +9,7 @@ import {
   loadTemplateModules,
   TemplateModuleCollection,
 } from "../../../common/src/template/internal/loader.js";
-import { createFeaturesJson } from "../../../generate/features.js";
+import { createFeaturesJson } from "../../../generate/features/createfeaturesjson.js";
```
oshi97, line 12: could we make this file camelcase? git can struggle changing filename casing so you might need to do two commits or do like `git rm --cached` or something

packages/pages/src/generate/features/features.ts
```diff
@@ -0,0 +1,66 @@
...
+        describe: "The folder path where all template files live",
+        type: "string",
+        default: "src/templates",
+        demandOption: false,
```
oshi97, line 51: is there a config file where somebody can specify their settings for this stuff?

packages/pages/src/generate/features/features.ts
yen-tt: Okay, I will remove the options for features in this pr. We can probably address the plugin options in the next PR I put up for build command

packages/pages/src/generate/features/features.ts
yen-tt: using pages command line, there's currently no way for user to configure these options beside the new arguments we are adding. From vite-plugin, they can technically pass in a `ProjectStructureConfig` object containing whatever settings they want (e.g `yextSSG({ //user-config })`). But as Duval[ mentioned above](https://github.com/yext/pages/pull/248#discussion_r1004799768), we may want to restrict what user can pass in.

packages/pages/src/generate/features/features.ts
```diff
@@ -4,18 +4,13 @@ import { Path } from "../../common/src/project/path.js";
...
->;
+type FeaturesArgs = Pick<ProjectFilepaths, "scope">;
 
-const handler = async ({
-  templatesRoot,
-  sitesConfigRoot,
-  scope,
-}: FeaturesArgs): Promise<void> => {
+const handler = async ({ scope }: FeaturesArgs): Promise<void> => {
+  const templatesRoot = "src/templates";
```
mkilpatrick, line 12: You might want to still get this from the default ProjectStructure

mkilpatrick (approved):

oshi97 (approved):

PR #249, opened 2022-10-25
feat(dev/plugin): add build command with scope flag

yen-tt:
Vite CLI does not allow user to pass non-vite arguments (github discussions: [[1](https://github.com/vitejs/vite/issues/7065)][[2](https://github.com/vitejs/vite/pull/1188)][[3](https://github.com/vitejs/vite/issues/2891)]) -- it will throw an error for unorganized options. The recommended approach in these github discussions is to use environment variable to pass flags or info to custom vite plugins.

This PR added a new pages command: `build`, which will accepts a `scope` option for multibrand setup. Internally, it will set  an env var `YEXT_PAGES_SCOPE` and executes `vite build`. In our pages vite plugin construction, the `scope` value store in env var will be use to construct the `ProjectStructure` object.

J=SLAP-2413
TEST=manual

tested the following commands in a [test repo](https://github.com/yen-tt/pages-starter/tree/dev/test-multibrand):
- `pages build` - successfully build a `dist` folder with expected server and static files based on templates at root level
- `pages build --scope locations.brand2.com` - successfully build a `dist` folder with expected server and static files based on templates from locations.brand2.com folder

- ran `npm run build` for root level temmplates and `npm run build:multibrand` for locations.brand1.com templates, and `npm run serve`. See that both builds contain the expected pages.

### DIFF
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -6,10 +6,12 @@ type BuildArgs = Pick<ProjectFilepaths, "scope">;
 
 const handler = async ({ scope }: BuildArgs) => {
   // Pass CLI arguments as env variables to use in vite-plugin
-  if (scope) {
-    process.env.YEXT_PAGES_SCOPE = scope;
-  }
-  shell.exec("vite build");
+  shell.exec("vite build", {
+    env: {
+      ...process.env,
+      YEXT_PAGES_SCOPE: scope,
+    },
+  });
 };
 
 export const buildCommandModule: CommandModule<unknown, BuildArgs> = {
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -1,0 +1,26 @@
+import shell from "shelljs";
+import { CommandModule } from "yargs";
+import { ProjectFilepaths } from "../common/src/project/structure.js";
+
+type BuildArgs = Pick<ProjectFilepaths, "scope">;
+
+const handler = async ({ scope }: BuildArgs) => {
+  // Pass CLI arguments as env variables to use in vite-plugin
+  if (scope) {
+    process.env.YEXT_PAGES_SCOPE = scope;
+  }
+  shell.exec("vite build");
+};
+
+export const buildCommandModule: CommandModule<unknown, BuildArgs> = {
+  command: "build",
+  describe: "Build site using Vite",
+  builder: (yargs) => {
+    return yargs.option("scope", {
+      describe: "The subfolder to scope the served templates from",
+      type: "string",
+      demandOption: false,
+    });
+  },
+  handler,
+};
--- a/packages/pages/src/vite-plugin/plugin.ts
+++ b/packages/pages/src/vite-plugin/plugin.ts
@@ -5,6 +5,7 @@ import {
   ProjectStructureConfig,
 } from "../common/src/project/structure.js";
 import { build } from "./build/build.js";
+import _ from "lodash";
 
 /**
  * Options to configure functionality of the plugin.
@@ -16,7 +17,16 @@ export type Options = {
 };
 
 const plugin = (opts: Options = {}): PluginOption[] => {
-  const projectStructure = new ProjectStructure(opts.projectStructureConfig);
+  const projectConfigFromBuildArgs: Optional<ProjectStructureConfig> = {
+    filepathsConfig: {
+      scope: process.env.YEXT_PAGES_SCOPE,
+    },
+  };
+  const userProjectStructureConfig = _.merge(
+    opts.projectStructureConfig,
+    projectConfigFromBuildArgs
+  );
+  const projectStructure = new ProjectStructure(userProjectStructureConfig);
 
   return [build(projectStructure)];
 };
--- a/packages/pages/src/vite-plugin/plugin.ts
+++ b/packages/pages/src/vite-plugin/plugin.ts
@@ -5,7 +5,7 @@ import {
   ProjectStructureConfig,
 } from "../common/src/project/structure.js";
 import { build } from "./build/build.js";
-import _ from "lodash";
+import merge from "lodash/merge.js";
 
 /**
  * Options to configure functionality of the plugin.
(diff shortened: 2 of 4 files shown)


mkilpatrick (approved):
Does this work in CI correctly?

yen-tt:
> Does this work in CI correctly?

I tested this by updating the `ci.json` file in root folder and in the brand folder, specifically `buildArtifacts.buidCmd` to run `pages build` and `pages build --scope locations.brand2.com`. Then I ran `yext pages build && yext pages generate-test-data && yext pages render` for root level, and same command chain with `--hostname locations.brand2.com` in all the yext pages commands for scope level. Both worked fine as I can see the correct pages when serving the rendered output folder. Are there other ways you would like me to test or if I'm missing something?

mkilpatrick:
> > Does this work in CI correctly?
> 
> I tested this by updating the `ci.json` file in root folder and in the brand folder, specifically `buildArtifacts.buidCmd` to run `pages build` and `pages build --scope locations.brand2.com`. Then I ran `yext pages build && yext pages generate-test-data && yext pages render` for root level, and same command chain with `--hostname locations.brand2.com` in all the yext pages commands for scope level. Both worked fine as I can see the correct pages when serving the rendered output folder. Are there other ways you would like me to test or if I'm missing something?

I meant trying it in true YextCI which would require pnpm packing pagesjs and hosting it somewhere public to try it out through Pages. It think it'll work though.

packages/pages/src/vite-plugin/plugin.ts
```diff
@@ -5,6 +5,7 @@ import {
   ProjectStructureConfig,
 } from "../common/src/project/structure.js";
 import { build } from "./build/build.js";
+import _ from "lodash";
```
oshi97, line 8: nit but probably better just to import the one method instead of all of lodash `{ merge } from "lodash"`

packages/pages/src/build/build.ts
```diff
@@ -0,0 +1,26 @@
...
+import { ProjectFilepaths } from "../common/src/project/structure.js";
+
+type BuildArgs = Pick<ProjectFilepaths, "scope">;
+
+const handler = async ({ scope }: BuildArgs) => {
+  // Pass CLI arguments as env variables to use in vite-plugin
+  if (scope) {
+    process.env.YEXT_PAGES_SCOPE = scope;
+  }
+  shell.exec("vite build");
```
oshi97, line 12: child_process.spawn is used in a few other places, would we want to use instead there as well? If we did I was thinking we could add it to the doc

oshi97, line 12: does this require vite to be globally installed?

packages/pages/src/build/build.ts
```diff
@@ -0,0 +1,26 @@
+import shell from "shelljs";
+import { CommandModule } from "yargs";
+import { ProjectFilepaths } from "../common/src/project/structure.js";
+
+type BuildArgs = Pick<ProjectFilepaths, "scope">;
+
+const handler = async ({ scope }: BuildArgs) => {
+  // Pass CLI arguments as env variables to use in vite-plugin
+  if (scope) {
+    process.env.YEXT_PAGES_SCOPE = scope;
```
oshi97, line 10: Directly setting a process.env wouldn't have side effects would it? Like it wouldn't bleed into other processes or the terminal? If possible I think I prefer setting them in shell.exec's second parameter

packages/pages/src/build/build.ts
yen-tt: no, `build` ran without vite globally installed on my machine

yen-tt: sure! What do you mean add it to the doc?

packages/pages/src/vite-plugin/plugin.ts
yen-tt: there's an error if we do named export like that:
Screen Shot 2022-10-26 at 12 18 13 PM
I will keep it as is for now

packages/pages/src/build/build.ts
yen-tt: sure! what do you mean by add it to the doc?

packages/pages/src/vite-plugin/plugin.ts
oshi97: oof
does `import merge from "lodash/merge"` work? if not that's fine

packages/pages/src/build/build.ts
oshi97: oh the pagesjs doc with the assorted small cleanup things we'd want

packages/pages/src/build/build.ts
yen-tt: directly setting it doesn't get into terminal or other processes, only that specific process and its subprocesses. Sure, I can set it in shell.exec second param.

yen-tt:
> > > Does this work in CI correctly?
> > 
> > 
> > I tested this by updating the `ci.json` file in root folder and in the brand folder, specifically `buildArtifacts.buidCmd` to run `pages build` and `pages build --scope locations.brand2.com`. Then I ran `yext pages build && yext pages generate-test-data && yext pages render` for root level, and same command chain with `--hostname locations.brand2.com` in all the yext pages commands for scope level. Both worked fine as I can see the correct pages when serving the rendered output folder. Are there other ways you would like me to test or if I'm missing something?
> 
> I meant trying it in true YextCI which would require pnpm packing pagesjs and hosting it somewhere public to try it out through Pages. It think it'll work though.

I tested this in Sites using `https://github.com/yext-pages/multibrand-slapshot` to deploy domain `sunglasses.rayban.com` after setting the domain field in site settings. Worked as expected.

packages/pages/src/vite-plugin/plugin.ts
yen-tt: `import merge from "lodash/merge.js"` works, will update

oshi97 (approved):

mkilpatrick (approved):

PR #250, opened 2022-10-28
fix(common): remove templatesRoot configuration

tatimblin:
Remove functionality allowing `src/templates` to be customized. Currently the templates directory is configurable [here](https://github.com/yext/pages/blob/6de548c5450a34795ce6855929805b72597f53be/packages/pages/src/common/src/project/structure.ts#L112), which should let you customize `src/templates`. This currently doesn’t work because the [glob import that gets all the templates has a requirement of being a string](https://github.com/vitejs/vite/issues/5478), I have a little [demo](https://stackblitz.com/edit/vitejs-vite-aw7cqj?file=index.html) where you can see this. Per [slack discussion](https://yext.slack.com/archives/C02LLE9BW2K/p1666894181264749) this was decided to be removed

J=SLAP-2426
TEST=auto,manual

pnpm pack-ed project and installed in starter site was able to serve & build


packages/pages/src/common/src/project/structure.ts
```diff
@@ -128,7 +125,7 @@ export class ProjectStructure {
     this.sitesConfigRoot = new Path(
       this.#config.filepathsConfig.sitesConfigRoot
     );
-    this.templatesRoot = new Path(this.#config.filepathsConfig.templatesRoot);
+    this.templatesRoot = new Path("src/templates");
```
nmanu1, line 128: is there a single file (like a constants file or something) where we can define `src/templates` as the root path for templates so it can be referred to everywhere? that way, if we made an update to the path, we wouldn't need to remember to update each place individually

tatimblin: The `ProjectStructure` type is used as the config as well as internally in some of the functions so to remove it for the config meant to hardcode it in the function. I considered including it in the type as 
```
interface ProjectStructure {
  templatesRoot: 'src/templates'
  somthingElse: string
  ...
}
```
but after seeing there are several type declarations for `ProjectStructure`, one of which where `templatesRoot` is of type `Path` and each hardcoding is in its own sub-repo(ish) directory, I decided against it.

What are your thoughts regarding this?

nmanu1: actually, it looks like there is a `pages/packages/pages/src/dev/server/ssr/constants.ts` file. it already has the template path hardcoded. what do you think of moving that file a few folders up so it can be shared, or making a similar constants file in `pages/packages/pages` or `pages/packages/pages/src` with the template path?

tatimblin: I made a new constants file

packages/pages/src/common/src/project/structure.ts
```diff
@@ -74,7 +73,6 @@ export interface ProjectStructureConfig {
 
 export const defaultProjectStructureConfig: ProjectStructureConfig = {
   filepathsConfig: {
-    templatesRoot: "src/templates",
```
mkilpatrick, line 77: Why remove this from the default?

packages/pages/src/common/src/template/types.ts
```diff
@@ -142,8 +142,6 @@ export type Manifest = {
   };
   /** A map of project roots to their paths */
   projectFilepaths: {
-    /** The folder path where all template files live */
-    templatesRoot: string;
```
mkilpatrick, line 146: I think we should leave ProjectStructure the same and just not have it be an arg to the vite plugin for now. Pretty much none of the other fields are going to work correctly if you change them either. Maybe they will? I haven't tested. But there are definitely quite a few hardcoded things. If you check the plugin's actual code I think it hardcodes the hydration folder name too.

packages/pages/src/common/constants.ts
```diff
@@ -0,0 +1 @@
+export const TEMPLATES_ROOT = "src/templates";
```
nmanu1, line 1: can this value be used in the other constants file so it's not defined in both places?

packages/pages/src/common/src/project/structure.ts
tatimblin: I removed this so that I could remove `templatesRoot` from the type definition so people wouldn't be able to set in the plugins config

packages/pages/src/common/src/template/types.ts
tatimblin: Ok I'll revert and remove the ability to pass in your own project structure config

packages/pages/src/common/constants.ts
tatimblin: This is outdated due to duval's suggestion

packages/pages/src/common/src/project/structure.ts
tatimblin: outdated

packages/pages/src/vite-plugin/plugin.ts
```diff
@@ -5,28 +5,14 @@ import {
...
 import { build } from "./build/build.js";
-import merge from "lodash/merge.js";
 
-/**
- * Options to configure functionality of the plugin.
- *
- * @public
- */
-export type Options = {
-  projectStructureConfig?: Optional<ProjectStructureConfig>;
```
nmanu1, line 16: just double checking, none of the options in this config have any effect if the user passes it into the plugin?

nmanu1: I see, so it's unlikely that this change would break something for a user? if so, should there be another item made to look into this further if these are configurations that are expected to be available and work?

mkilpatrick (approved):

PR #252, opened 2022-11-01
fix(components): add loading eager to image

### LINKED ISSUE #223: image: allow default browser loading behaviour
### Describe the bug

I am loading a Yext image at the top of a page "above the hold" with the `<Image>` component. Because the image is above the fold it should not be lazy loaded, but the component currently hardcodes this state.

I would like to be able to set `loading="eager"` on the component.

[Further reading](https://web.dev/browser-level-image-lazy-loading/#avoid-lazy-loading-images-that-are-in-the-first-visible-viewport)
[Further reading 2](https://web.dev/lcp-lazy-loading)

### Reproduction

https://github.com/yext-pages/Master-Ria

### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 275.99 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 18.4.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.0 - ~/repo/Master-Ria/node_modules/.bin/npm
  Browsers:
    Chrome: 105.0.5195.125
    Firefox: 100.0
    Safari: 15.6.1
  npmPackages:
    @yext/pages: 1.0.0-beta.6 => 1.0.0-beta.6
```

### Used Package Manager

pnpm

### Used Browser

Chrome

### Validations

- [x] Check that there isn't [already an issue](https://github.com/yext/pages/issues) that reports the same bug to avoid creating a duplicate.

tatimblin:
fix #223

J=SLAP-2436
TEST=auto

### DIFF
--- a/packages/pages/src/components/image/image.tsx
+++ b/packages/pages/src/components/image/image.tsx
@@ -27,6 +27,7 @@ export const Image = ({
   placeholder,
   imgOverrides,
   style = {},
+  loading = "lazy",
 }: ImageProps) => {
   const imgRef = useRef<HTMLImageElement>(null);
   const [isImageLoaded, setIsImageLoaded] = useState(false);
@@ -108,7 +109,7 @@ export const Image = ({
         height={absHeight}
         srcSet={srcSet}
         sizes={sizes}
-        loading={"lazy"}
+        loading={loading}
         alt={imageData.alternateText || ""}
         {...imgOverrides}
       />
--- a/packages/pages/src/components/image/types.ts
+++ b/packages/pages/src/components/image/types.ts
@@ -71,25 +71,6 @@ export type ImageLayout =
   typeof ImageLayoutOption[keyof typeof ImageLayoutOption];
 
 /**
- * Loading option on the Image component
- *
- * @public
- */
-export const LoadingOption = {
-  /** The default loading state, will wait for image to enter viewport before loading */
-  LAZY: "lazy",
-  /** Force image to load immediately */
-  EAGER: "eager",
-} as const;
-
-/**
- * The type definition for image loading.
- *
- * @public
- */
-export type Loading = typeof LoadingOption[keyof typeof LoadingOption];
-
-/**
  * The shape of the data passed to {@link Image}.
  */
 interface BaseImageProps {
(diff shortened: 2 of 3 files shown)


packages/pages/src/components/image/image.tsx
```diff
@@ -108,7 +109,7 @@ export const Image = ({
         height={absHeight}
         srcSet={srcSet}
         sizes={sizes}
-        loading={"lazy"}
+        loading={loading}
```
nmanu1, line 112: do we want to add tests for this?

yen-tt (approved):

packages/pages/src/components/image/types.ts
```diff
@@ -70,6 +70,25 @@ export const ImageLayoutOption = {
 export type ImageLayout =
   typeof ImageLayoutOption[keyof typeof ImageLayoutOption];
 
+/**
+ * Loading option on the Image component
+ *
+ * @public
+ */
+export const LoadingOption = {
```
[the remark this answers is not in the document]
tatimblin: I thought so but wanted to match the [ImageLayoutOption](https://github.com/yext/pages/pull/252/files#diff-6beae3c822d7f350f9bc40753156055b412ff2f8e565a979af3a2283f81e1125R46) above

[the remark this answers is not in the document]
oshi97: I might even just use `HTMLImageElement["loading"]` as the type? The only two options are `"lazy" | "eager"` and I think doing so would remove some boilerplate

packages/pages/src/components/image/types.ts
```diff
@@ -70,6 +70,25 @@ export const ImageLayoutOption = {
...
+  /** Force image to load immediately */
+  EAGER: "eager",
+} as const;
+
+/**
+ * The type definition for image loading.
+ *
+ * @public
+ */
+export type Loading = typeof LoadingOption[keyof typeof LoadingOption];
```
oshi97, line 90: I don't think we're using the LoadingOption const for anything except creating this type, so it could probably be simplified to just `loading?: "eager" | "lazy"` down below and we wouldn't need the new type

oshi97: that would match
image

tatimblin: Yeah I'll switch it back to this

oshi97 (approved):

PR #253, opened 2022-11-03
feat(components): add Breadcrumbs component

cblair28:
This change adds the breadcrumb component, which takes in a 2d array of type [string, string][] of crumb names and crumb urls, and renders them on the page

J=CR-2793
TEST=unit

### DIFF
--- a/packages/pages/src/components/breadcrumbs/breadcrumbs.test.tsx
+++ b/packages/pages/src/components/breadcrumbs/breadcrumbs.test.tsx
@@ -1,0 +1,48 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import Breadcrumbs from "./breadcrumbs.js";
+
+const Breadcrumbs_data: { slug: string; name: string }[] = [
+  { name: "VA", slug: "va.html" },
+  { name: "Vienna", slug: "va/vienna.html" },
+  { name: "123 Main St", slug: "va/vienna/123-Main-St.html" },
+];
+
+describe("Breadcrumbs", () => {
+  it("Renders component", () => {
+    render(<Breadcrumbs breadcrumbs={Breadcrumbs_data} />);
+  });
+
+  it("Renders component with a single crumb", () => {
+    render(<Breadcrumbs breadcrumbs={[{ name: "VA", slug: "va.html" }]} />);
+  });
+
+  it("Renders single crumb with no url", () => {
+    render(<Breadcrumbs breadcrumbs={[{ name: "VA", slug: "" }]} />);
+  });
+
+  it("Renders first crumb an anchor tag not a span", () => {
+    render(<Breadcrumbs breadcrumbs={Breadcrumbs_data} />);
+    const firstCrumb = screen.getByText("VA");
+    const firstCrumbParent = firstCrumb.parentNode;
+    expect(firstCrumbParent?.nodeName === "A").toBeTruthy();
+  });
+
+  it("Does not render last crumb as an anchor tag", () => {
+    render(<Breadcrumbs breadcrumbs={Breadcrumbs_data} />);
+    const lastCrumb = screen.getByText("123 Main St");
+    const lastCrumbParent = lastCrumb.parentNode;
+    expect(lastCrumbParent?.nodeName === "LI").toBeTruthy();
+  });
+
+  it("Renders component with custom separator", () => {
+    render(<Breadcrumbs breadcrumbs={Breadcrumbs_data} separator=">" />);
+    const separators = screen.getAllByText(">");
+    for (const separator of separators) {
+      expect(separator.className === "Breadcrumbs-separator").toBeTruthy();
+    }
+  });
+});
--- a/packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
+++ b/packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
@@ -1,0 +1,74 @@
+import React from "react";
+import classNames from "classnames";
+import { BreadcrumbsProps } from "./types.js";
+import { Link } from "../link/link.js";
+import { getHref } from "../link/methods.js";
+import "./breadcrumbs.css";
+
+export interface BreadcrumbsPropsDefault extends BreadcrumbsProps {
+  breadcrumbs: Array<{ slug: string; name: string }>;
+}
+
+/*
+ * The 'Breadcrumbs' component renders a breadcrumbs style navigation bar
+ *
+ * @example
+ * ```
+ * <Breadcrumbs breadcrumbs={defaultTransformer(streamBreadcrumbs)} />
+ * ```
+ *
+ * @param {[{name: string, slug: string}]} breadcrumbs
+ *          List of {crumbName, crumbUrl}
+ * @param {ReactElement | string} separator
+ *          between each breadcrumb - defaults to '/', also accepts any HTML element (<svg>, <img>, etc)
+ */
+const Breadcrumbs = (props: BreadcrumbsPropsDefault) => {
+  const { breadcrumbs, className, separator = "/" } = props;
+
+  return (
+    <>
+    {breadcrumbs?.length && (
+      <nav
+        className={classNames("Breadcrumbs", className)}
+        aria-label="Breadcrumb"
+      >
+        <ol className="Breadcrumbs-list">
+          {breadcrumbs.map(({ name, slug }, idx) => {
+            const isLast = idx === breadcrumbs.length - 1;
+
+            return (
+              <li className="Breadcrumbs-item" key={idx}>
+                <Breadcrumb name={name} slug={isLast ? "" : slug} {...props} />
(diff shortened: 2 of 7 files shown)


packages/pages/src/components/breadcrumbs/breadcrumbs.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  it("Renders component with a single crumb", () => {
+    render(<Breadcrumbs breadcrumbs={[["VA", "va.html"]]} />);
+  });
+
+  it("Renders single crumb with no url", () => {
+    render(<Breadcrumbs breadcrumbs={[["VA", ""]]} />);
+  });
+
+  it("Renders first crumb a link", () => {
```
mkilpatrick, line 27: as a link?

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+import classNames from "classnames";
+import { BreadcrumbsProps } from "./types.js";
+import { Link } from "../link/link.js";
+import { cta } from "../link/methods.js";
+import "./breadcrumbs.css";
+
+export interface BreadcrumbsPropsDefault extends BreadcrumbsProps {
+  breadcrumbs: Array<[string, string]>;
```
mkilpatrick, line 9: Why transform this to a more generic [string, string] instead of having them strongly typed by name?

packages/pages/src/components/breadcrumbs/yextBreadcrumbs.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface BreadcrumbsPropsStreams extends BreadcrumbsProps {
+  streamsBreadcrumbs: StreamsBreadcrumbs;
+}
+
+/*
+ * Wrapper for the Breadcrumbs component which automatically processes Yext Streams data
+ */
+const YextBreadcrumbs = (props: BreadcrumbsPropsStreams) => {
```
mkilpatrick, line 25: What's the purpose of this instead of just using the Breadcrumbs directly?

packages/pages/src/components/breadcrumbs/types.ts
```diff
@@ -0,0 +1,8 @@
+import { ReactNode } from "react";
+
+export interface BreadcrumbsProps {
+  separator?: ReactNode;
+  className?: string;
+  yaTrack?: string;
```
tatimblin, line 6: we don't need ya track/scope McGin made an analytics component that handles sending those events differently

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+import classNames from "classnames";
+import { BreadcrumbsProps } from "./types.js";
+import { Link } from "../link/link.js";
+import { cta } from "../link/methods.js";
+import "./breadcrumbs.css";
+
+export interface BreadcrumbsPropsDefault extends BreadcrumbsProps {
```
tatimblin, line 8: What's the value of `BreadcrumbsPropsDefault`? Also can you extend an `HTMLDivElement` and pass the props to the wrapping element like in [address](https://github.com/yext/pages/blob/main/packages/pages/src/components/address/address.tsx#L39) (...props)

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * The 'Breadcrumbs' component renders a breadcrumbs style navigation bar
+ *
+ * @example
+ * ```
+ * <Breadcrumbs breadcrumbs={defaultTransformer(streamBreadcrumbs)} />
+ * ```
+ *
+ * @param {[[string, string]]} breadcrumbs
+ *          List of [crumbName, crumbUrl]
+ * @param {ReactElement | string} separator
```
tatimblin, line 22: On `BreadcrumbsProps` this is only a `ReactElement`

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
...
+      data-ya-scope={yaScope || "breadcrumbs"}
+      aria-label="Breadcrumb"
+    >
+      <ol className="Breadcrumbs-list">
+        {breadcrumbs.map(([label, link], idx) => {
+          const isLast = idx === breadcrumbs.length - 1;
+
+          return (
+            <li className="Breadcrumbs-item" key={idx}>
+              <Breadcrumb label={label} link={isLast ? "" : link} {...props} />
```
tatimblin, line 40: nit: I think this would work

```suggestion
              <Breadcrumb label={label} link={!isLast && link} {...props} />
```

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * @param {[[string, string]]} breadcrumbs
+ *          List of [crumbName, crumbUrl]
+ * @param {ReactElement | string} separator
+ *          between each breadcrumb - defaults to '/', also accepts any HTML element (<svg>, <img>, etc)
+ */
+const Breadcrumbs = (props: BreadcrumbsPropsDefault) => {
+  const { breadcrumbs, yaScope, className, separator = "/" } = props;
+
+  return (
+    <nav
```
tatimblin, line 29: Can we confirm links would be present before rendering the `<nav>`

packages/pages/src/components/breadcrumbs/methods.test.ts
```diff
@@ -0,0 +1,32 @@
+import { defaultTransformer } from "./methods.js";
+import { StreamsBreadcrumbs } from "./yextBreadcrumbs.js";
+
+function doTestDefaultTransformer() {
+  const testDetails: Array<[StreamsBreadcrumbs, Array<[string, string]>]> = [
+    [
+      {
+        dm_directoryParents: [
+          { slug: "index.html", name: "All locations" },
```
tatimblin, line 9: This raw data seems like it would be fine to use without transforming into the tuple

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
tatimblin: HTMLNavElement

packages/pages/src/components/breadcrumbs/yextBreadcrumbs.tsx
Ceseura: This was supposed to be an adapter template to unmarshall whatever the Streams breadcrumbs data format was into the generic `[string, string]` format

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
Ceseura: The implementation of this component involved two ways of using it:
1. Using `BreadcrumbsPropsDefault` and passing an array of [string, string] to manually set the breadcrumbs
2. Using `BreadcrumbsPropsStreams` and passing the profile field (?) and it would automatically unwrap the breadcrumbs from the correct directoryManager custom field

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+interface BreadcrumbProps {
+  label: string;
+  link?: string;
+}
+
+const Breadcrumb = (props: BreadcrumbProps) => {
+  const { label, link } = props;
+
+  let breadcrumbEl = <span className="Breadcrumbs-label">{label}</span>;
```
oshi97, line 60: nit but I think it's a little clearer if you just return the jsx directly?
```tsx
if (link) {
  return <Link/>
}

return <span/>
```
is probably simpler

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
cblair28: When I do that I get this type error: `Type 'string | false' is not assignable to type 'string | undefined'.
  Type 'boolean' is not assignable to type 'string'.`

packages/pages/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,74 @@
...
+  name: string;
+  slug?: string;
+}
+
+const Breadcrumb = (props: BreadcrumbProps) => {
+  const { name, slug } = props;
+
+  if (slug) {
+    return (
+      <Link className="Breadcrumbs-link Link--breadcrumbs" href={slug}>
```
Jared-Hood, line 65: This is missing a way to set the `eventName` property in order to add analytics to the component. Something like adding a boolean prop (`addAnalytics`) to the `Breadcrumbs` component and then if true setting the `eventName` here to `breadcrumb_${idx}` would work.

PR #254, opened 2022-11-09
feat(components): make check for provider in getDirections check be case insensitive

Jeffrey-Rhoads17:
J=CR-2961
TEST=manual

### DIFF
--- a/packages/pages/src/components/address/methods.test.ts
+++ b/packages/pages/src/components/address/methods.test.ts
@@ -31,12 +31,15 @@ const sampleListings: ListingType[] = [
   },
 ];
 
-const sampleListingsCaseInsensitive: ListingType[] = [
+const sampleListingsLowerCase: ListingType[] = [
   {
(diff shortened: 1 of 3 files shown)


bhainesva:
Change LGTM, but could we add  unit tests to the methods.test.ts file for the lowercase and uppercase cases?

bhainesva (approved):

packages/pages/src/components/address/methods.test.ts
```diff
@@ -85,6 +98,12 @@ describe("getDirections()", () => {
     );
   });
 
+  it("returns URL to Google Maps GMB listing without being case sensitive", () => {
+    expect(getDirections(undefined, sampleListingsCaseInsensitive)).toEqual(
```
bhainesva, line 102: I don't think this is exactly what we want. Our getDirections function could always be using the lowercase version, but since we pass in both we wouldn't know that based on this test case. 

I think we want an input where it's only lowercase, and one where it's only uppercase, and show that in both situations we get the listingUrl out.

PR #255, opened 2022-11-10
feat(components): make check for publisher in getDirections be case-insensitive

Jeffrey-Rhoads17:
…ase insensitive (#254)

J=CR-2961
TEST=manual

### DIFF
--- a/packages/pages/src/components/address/methods.test.ts
+++ b/packages/pages/src/components/address/methods.test.ts
@@ -31,6 +31,22 @@ const sampleListings: ListingType[] = [
   },
 ];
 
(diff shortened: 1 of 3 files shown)


JRonkin (dismissed):
LGTM

tmeyer2115:
Small nit: can we clean up the commit title? Something like "feat(components): make check for publisher in getDirections case-insensitive". Right now the term "check" is used twice, which is a bit confusing, and we use the `provider` terminology. I think `publisher` would be more correct as the Provider is likely KG.

tmeyer2115 (approved):

PR #256, opened 2022-11-11
feat(dev): document.slug URLs in dev mode

oshi97:
This PR changes URLs in dev mode from `/[template-name]/[entity-id]` to the form `/[document.slug]`,
where `document.slug` is the slug field in the entity and corresponding test data.

In order to render a template, we need two things, the template file itself (for example `src/templates/location.tsx`),
and the data for the entity we want to render using the template, i.e. the test data.

With the original URLs, the template to use is just the [template-name] in the URL, and the test-data is fetched
via the Yext CLI using the [entity-id] in the URL combined with the `config` export that could be loaded from the template.
The `config` would then be passed to the CLI inline.

With the new URLs, the order we get this information has flipped around. With prod URLs, we first get the test data
by hitting the CLI using the new --slug parameter added to generate-test-data in https://gerrit.yext.com/c/alpha/+/192814.
However, we STILL need a FeaturesConfig to send to the CLI. At least I think we do, when I try to call the CLI without it I get an error.

Since we have no way of knowing which template we need yet, and can't just get the `config` export like the original URLs, I decided to loop through every template and generate a FeaturesConfig with all templates that are for entity pages. I'm not sure if this is the right way to do this but it seems to work? This seems to let us hit the CLI properly, and gets us the test data we want. The CLI should error out if it encounters 0 or multiple entities for a single slug. 

(I believe this approach will not work if multiple templates share an entity, but I don't think that's a case we can address without more information, because which template would we even render to the page?)

After getting the test data we can get the corresponding template using `document.__.name`, and follow the same flow as the original URLs.

Also fixes a bug where static pages in --local mode would sometimes grab the wrong test data document (they would always grab the first one that did not have an entityId)

Adds an error message to the 404 pages for cases like when no document with the slug is found.

J=SLAP-2447
TEST=manual

when --no-prod-url is set, the original URL scheme will still be used. tested with --local and also dynamically generated test data. index page URLs look correct for both static and entity pages

tested that with prod urls and --local set the page renders as expected, and the index page has the right links.
tested on https://github.com/yext/pages-starter-react-locations, and also added an additional page in order to test with multiple entity types
tested that alternativeLanguageFields still works and will provide a document for the right locale with alternativeLanguageFields 
tested that index page will render (relatively) gracefully even when localData does not have slugs


packages/pages/src/dev/dev.ts
```diff
@@ -6,10 +6,11 @@ import { ProjectFilepaths } from "../common/src/project/structure.js";
 
 interface DevArgs extends Pick<ProjectFilepaths, "scope"> {
   local?: boolean;
+  "prod-url"?: boolean;
```
mkilpatrick, line 9: remove quotes

packages/pages/src/dev/dev.ts
```diff
@@ -6,10 +6,11 @@ import { ProjectFilepaths } from "../common/src/project/structure.js";
...
 }
 
-const handler = async ({ scope, local }: DevArgs) => {
-  await createServer(!local, scope);
+const handler = async ({ scope, local, "prod-url": useProdURLs }: DevArgs) => {
```
mkilpatrick, line 12: remove quotes

mkilpatrick:
I think this should still allow rendering static files. You can still do that by looking at the slug. You can tell that it's a static page by the fact that the config has no stream. We use this assumption when creating the features.json. If you can check that to identify the template type you can then short circuit and return the template without making a call to the CLI.

packages/pages/src/dev/dev.ts
oshi97: you need quotes here because of the dash!

oshi97:
going to add handling for static pages, but it shouldn't affect the code currently in the PR much

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -159,18 +180,30 @@ export const generateTestDataForPage = async (
...
+
+const getCommonArgs = (
+  featuresConfig: FeaturesConfig,
+  projectStructure: ProjectStructure
+) => {
+  const args = ["pages", "generate-test-data", "--printDocuments"];
+
+  args.push("--featuresConfig", prepareJsonForCmd(featuresConfig));
+
+  const sitesConfigPath =
```
tatimblin, line 193: Can this be done elsewhere so `siteStream` can be passed in instead of `projectStructure`, which is only used to get `siteStream`.

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -36,42 +39,60 @@ export const generateTestData = async (hostname?: string): Promise<boolean> => {
...
+  locale: string,
+  projectStructure: ProjectStructure
+): Promise<any> => {
+  const templateModules = await loadTemplateModules(
+    getTemplateFilePathsFromProjectStructure(projectStructure),
+    true,
+    false
+  );
+  const featuresConfig = await getFeaturesConfig(templateModules);
+  const featuresConfigForEntityPages: FeaturesConfig = {
```
yen-tt, line 54: nit: can we rename this to indicate this only contain non-static / entity related features?

packages/pages/src/dev/server/server.ts
```diff
@@ -8,9 +8,11 @@ import { indexPage } from "./middleware/indexPage.js";
 import { generateTestData } from "./ssr/generateTestData.js";
 import { ProjectStructure } from "../../common/src/project/structure.js";
 import { finalSlashRedirect } from "./middleware/finalSlashRedirect.js";
+import { serverRenderSlugRoute } from "./middleware/serverRenderSlugRoute.js";
 
 export const createServer = async (
   dynamicGenerateData: boolean,
```
yen-tt, line 14: Currently, if `dynamicGenerateData = true`, I believe this would call generateTestData(..), and then proceed to call generateTestDataForSlug / generateTestDataForPage on each new url path. I think the logic should be updated such that we either call `generateTestData` once on createServer OR call generateTestDataForSlug / generateTestDataForPage on each new route request?

packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -89,8 +84,31 @@ export const getLocalDataForEntity = async (
...
+      `No localData files match entityId and locale: ${entityId} ${locale}`
+    );
+  }
+  return localData;
+};
 
-  throw new Error(
-    `No localData files match entityId and locale: ${entityId} ${locale}`
-  );
+export const getLocalDataForSlug = async (slug: string, locale: string) => {
```
yen-tt, line 104: if there's more than one localData file generated for this slug field (coming from different streams), we should throw an error or inform user of that. I think currently this would just pick whatever file it gets to first?

mkilpatrick:
Some comments trying this out myself using the [starter](https://github.com/yext/pages-starter-react-locations) pointing locally to a packed version of your PR:

1. The urls on the index page don't work or are wrong.
- For instance, the static pages lead to a non-working page (presumably since you dropped static page support until your next commit.
```
500 TypeError: Cannot read properties of undefined (reading '__')
at file:///Users/mkilpatrick/Desktop/pages-starter-react-locations/node_modules/@yext/pages/dist/dev/server/middleware/serverRenderSlugRoute.js:20:30
at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```
- entity pages give urls like `http://localhost:5173/undefined` or `http://localhost:5173/loc10-bucktown`, but which don't work. It's not clear to me if the first example was because I using a site that **does not** have a published deploy and only a staging one. The second example I **was** using a site with a published deploy.

2. Given bullet two above, prod urls obviously only make sense when you have a live site. What should we do if there is no live site yet? Most dev work occurs _before_ a site first goes live. It would be nice to somehow fall back to the previous way, though this is a Product call.

oshi97:
> Some comments trying this out myself using the [starter](https://github.com/yext/pages-starter-react-locations) pointing locally to a packed version of your PR:
> 
> 1. The urls on the index page don't work or are wrong.
> 
> * For instance, the static pages lead to a non-working page (presumably since you dropped static page support until your next commit.
> 
> ```
> 500 TypeError: Cannot read properties of undefined (reading '__')
> at file:///Users/mkilpatrick/Desktop/pages-starter-react-locations/node_modules/@yext/pages/dist/dev/server/middleware/serverRenderSlugRoute.js:20:30
> at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
> ```
> 
> * entity pages give urls like `http://localhost:5173/undefined` or `http://localhost:5173/loc10-bucktown`, but which don't work. It's not clear to me if the first example was because I using a site that **does not** have a published deploy and only a staging one. The second example I **was** using a site with a published deploy.
> 
> 2. Given bullet two above, prod urls obviously only make sense when you have a live site. What should we do if there is no live site yet? Most dev work occurs _before_ a site first goes live. It would be nice to somehow fall back to the previous way, though this is a Product call.

the URLs likely weren't working because the CLI version supporting --slug wasn't finished deploying yet. I'm not sure about the /undefined url but my guess is that was due to the entity not having a document.slug set

packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -127,3 +124,31 @@ export const indexPage =
...
+    )}/${entityId}`;
+  };
+
+  const entities = localDataManifest.entity.get(templateName) || [];
+  return entities.reduce((entityAccumulator, { entityId, slug }) => {
+    return (
+      entityAccumulator +
+      `<li>
+      <a href="${formatLink(entityId, slug)}">
+        ${entityId}
```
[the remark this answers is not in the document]
oshi97: sure updated!

packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -4,21 +4,19 @@ import { readdir } from "fs/promises";
...
 
-class LocalDataManifest {
-  static: Array<string>;
-  entity: Map<string, Array<string>>;
-
-  constructor() {
-    this.static = [];
-    this.entity = new Map<string, Array<string>>();
-  }
+export interface LocalDataManifest {
```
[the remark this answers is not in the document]
oshi97: Sure!

packages/pages/src/dev/server/ssr/generateTestData.ts
oshi97: I tried to do that with `featuresConfigForEntityPages` but definitely open to suggestions!

packages/pages/src/dev/server/server.ts
oshi97: I believe we can't call generateTestData once on createServer because we don't know what URLs they'll want to test yet, and therefore don't know what slugs to call generateTestData for.

I think calling generateTestDataForPage once per new route request makes sense! You're essentially asking for caching, right? I imagine we'd want some way for user's to "bust" the cache and some time limit for cached generateTestData calls, say if a user wants to add new data to their Knowledge Graph and then test it in Pages. I think this is out of scope here though.

packages/pages/src/dev/server/ssr/getLocalData.ts
oshi97: I'm not sure exactly where this kind of validation should live but I don't think it should hurt to add it here for now!

oshi97:
I did some rough benchmarking since I noticed the first few requests were a little slow (maybe there is some warmup time or something?) 
image

packages/pages/src/dev/server/server.ts
yen-tt: I'm not sure if I follow, but we currently do [call `generateTestData()` in `createServer`](https://github.com/yext/pages/blob/dev/prod-urls-in-dev/packages/pages/src/dev/server/server.ts#L50) if `dynamicGenerateData` is true.

What I meant is, in this PR, if `dynamicGenerateData` is true (by default) and if user use `prod-url` flag, then we would generate 10 test data files randomly with `generateTestData()`, AND call `generateTestDataForSlug` for each time we go to an entity page. I think we should update the logic to NOT run that generate 10 test data files when `prod-url` is true, since that's not necessary. Similarly with `generateTestDataForPage`, if that is going to be executed, then we should need to call `generateTestData()`.

packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -138,8 +154,18 @@ export const getLocalDataForSlug = async ({
   locale: string;
   slug: string;
 }) => {
-  const localData = await getLocalData((data) => {
-    return data.slug?.toString() === slug && data.locale === locale;
+  let localData: Record<string, any> | null = null;
+  await getLocalData((data) => {
+    if (data.slug && data.slug === slug && data.locale === locale) {
+      if (!localData) {
+        localData = data;
```
yen-tt, line 161: this setup seems a little weird, this criterion function is meant to return a boolean for `getLocalData` but we are directly extracting the data here instead of what is returning from the actual function `getLocalData`. Is there any other we can do this to handle the error case of multiple localData files?

oshi97: ah yeah I wanted to be clever and avoid making another method (since we'd still want the other uses of getLocalData to short circuit) but I can just write another method

packages/pages/src/dev/server/server.ts
oshi97: I'm not sure why generateTestData gets called in create server only if dynamicGenerateData is true, I asked Duval on Slack

oshi97: Duval says this is to populate the index page

oshi97: if the user specifies --local we assume they have localData already

packages/pages/src/dev/server/ssr/getLocalData.ts
oshi97: The more I think about it, does it make sense to validate it here? Doesn't the CLI already throw an error if there are multiple entities with the same slug?

packages/pages/src/dev/server/server.ts
oshi97: gonna leave it like this for now if that's ok! we can definitely revisit

packages/pages/src/dev/server/ssr/getLocalData.ts
oshi97: I reverted back to the original getLocalData usage, but lmk if you feel there should be changes

yen-tt: yes, the CLI throws an error if there are multiple entities with the same slug. However, it doesn't handle the case of multiple localData files for a single entity but from different stream definitions (e.g. if there are "stream-1" and "stream-2" both have `entityType` as location, and a single entity with slug `abington-location-123`, then the command would generate two files for each stream definition).
I believe the original getLocalData would just get whichever file it comes across first. We either need to check if the stream name in the localData filename match with what is expected for the feature page, OR throw an error if there are multiple non-static data files found.

tatimblin:
My comments look good 👍

packages/pages/src/dev/server/ssr/getLocalData.ts
oshi97: Ah ok that makes sense, I can update! I think the dynamic mode generate-test-data call might also need a tweak

yen-tt:
error in code/snyk?

oshi97:
> error in code/snyk?

none of these "Unsanitized input from an HTTP header flows into end" issues are new, they're only popping up again because I refactored them into new methods

yen-tt (dismissed):

packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -125,12 +127,30 @@ export const indexPage =
...
+      )}">
+          ${staticURL}
+        </a>
+      </li>
+    </ul>`,
+    ""
+  );
+};
+
+const createPageListItems = (
```
mkilpatrick, line 148: createEntityPageListItems?

packages/pages/src/dev/server/middleware/sendStaticPage.ts
```diff
@@ -0,0 +1,27 @@
+import { TemplateModuleInternal } from "../../../common/src/template/internal/types.js";
+import { TemplateRenderProps } from "../../../common/src/template/types.js";
+import { getLocalDataForEntityOrStaticPage } from "../ssr/getLocalData.js";
+import { propsLoader } from "../ssr/propsLoader.js";
+import sendAppHTML from "./sendAppHTML.js";
+import { Response } from "express-serve-static-core";
+import { ViteDevServer } from "vite";
+
+export default async function sendStaticPage(
```
mkilpatrick, line 9: Add some TSDoc

packages/pages/src/dev/server/middleware/serverRenderSlugRoute.ts
```diff
@@ -19,32 +21,45 @@ type Props = {
...
+        return;
+      }
+
       const feature = document.__.name;
       const entityId = document.id;
-      const templateModuleInternal = await featureNameToTemplateModuleInternal(
+      const templateModuleInternal = await findTemplateModuleInternal(
         vite,
-        feature,
+        (t) => feature === t.config.name,
```
mkilpatrick, line 54: I think this is fine, but double check that the templateModuleInternal's config name is properly set to the either the config.name in the template OR the name of the file if it's not present. And that that same thing is being set in the document.__.name.

packages/pages/src/dev/server/server.ts
```diff
@@ -45,6 +45,8 @@ export const createServer = async (
   app.use(finalSlashRedirect);
 
   let displayGenerateTestDataWarning = false;
+  // call generateTestData to ensure we have data to populate the index page
```
mkilpatrick, line 48: Use caps/periods. It makes it easier to read. I originally read this and below together and had to reread to understand that they were different sentences.

packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -1,18 +1,40 @@
...
+    // The return value of the template's getPath()
+    staticURL: string;
+  }[];
+  entity: Map<
+    string,
+    {
+      // The entity's document.id
+      entityId: string;
+      // The entity's document.slug
+      slug: string | undefined;
```
mkilpatrick, line 26: When is slug undefined?

packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
```diff
@@ -0,0 +1,20 @@
...
+  slug: string,
+  templateFilepaths: string[]
+) {
+  return findTemplateModuleInternal(
+    vite,
+    (t) => {
+      if (!isStaticTemplateConfig(t.config)) {
+        return false;
+      }
+      return slug === t.getPath({});
```
mkilpatrick, line 16: I wonder if this could be an issue. Obviously for static files we want a static url, but I wonder if it's possible to pass in props in the static template and that would mess things up. Hopefully that's handled or caught earlier and a user can't do that. Probably not an issue with your change here though.

packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -42,30 +62,44 @@ export const getLocalDataManifest = async (): Promise<LocalDataManifest> => {
...
+    const entityId = data.id?.toString();
+    const slug = data.slug?.toString();
+    if (entityId) {
+      localDataManifest.entity.set(featureName, [
+        ...(localDataManifest.entity.get(featureName) || []),
+        { entityId, slug },
+      ]);
+    } else {
+      // The lack of an entityId signifies that this is a static template.
+      const templateModuleInternal = await findTemplateModuleInternal(
```
mkilpatrick, line 79: Why do we need to do this now when we didn't before? It doesn't seem to be related to slugs this that's handled above.

packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -1,24 +1,44 @@
 import path from "path";
 import fs from "fs";
 import { readdir } from "fs/promises";
+import { findTemplateModuleInternal } from "./findTemplateModuleInternal.js";
+import { ViteDevServer } from "vite";
 
 const LOCAL_DATA_PATH = "localData";
```
mkilpatrick, line 7: Not your change, but maybe this should be part of the projectStructure? Could be a followup PR if we want to do that.

packages/pages/src/dev/server/ssr/getLocalData.ts
oshi97: when you don't have a slug defined in your entity

packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
oshi97: I was thinking for static pages it might be nice to have the module export a static string instead of a function like
```
export const path = 'static-url.html';
```
but I'm not sure I like having multiple ways to specify the path so unsure for now

packages/pages/src/dev/server/ssr/getLocalData.ts
oshi97: This is to generate the URLs for static pages. Previously, the static page URLs would be the name of the feature, however the template name is typically not the URL that would be used in prod.

For instance in the pages react starter's `robots.tsx`, the feature name is `'turtlehead-tacos'`, however the getPath returns `'index.html'`. Previously, the corresponding URL in the index page would be `'turtlehead-tacos'`, but now it is `'index.html'`

packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
oshi97: Originally I was also worried about somebody's getPath erroring out, but my thought was that if you have a static page and do anything except directly return a string, then you're not following the contract of static templates.

packages/pages/src/dev/server/ssr/getLocalData.ts
oshi97: I can add this to our cleanup doc. I'm not too familiar with ProjectStructure and this string has been here for 7 months

packages/pages/src/dev/server/ssr/getLocalData.ts
mkilpatrick: Makes sense.

packages/pages/src/dev/server/ssr/getLocalData.ts
mkilpatrick: Sounds good

mkilpatrick (approved):

PR #257, opened 2022-11-14
fix(components): add empty string fallback for when publisher is null

Jeffrey-Rhoads17:
…esent

J=CR-2961
TEST=manual

### DIFF
--- a/packages/pages/src/components/address/methods.ts
+++ b/packages/pages/src/components/address/methods.ts
@@ -94,16 +94,16 @@ export const getDirections = (
       return getDirectionsBing(query, config.route);
     }
     default: {
-      const listingsMap: { [key in ListingPublisher]?: string } | undefined =
-        listings?.reduce(
-          (obj, listing) => ({
-            ...obj,
-            [(listing.publisher || "").toUpperCase()]: listing.listingUrl,
-          }),
-          {}
-        );
-      if (listingsMap[ListingPublisherOption.GOOGLEMYBUSINESS]) {
-        return listingsMap[ListingPublisherOption.GOOGLEMYBUSINESS];
+      const gmbListing = listings.find(
+        (listing) =>
+          listing &&
+          listing.publisher &&
+          listing.publisher.toUpperCase() ===
+            ListingPublisherOption.GOOGLEMYBUSINESS
+      );
+
+      if (gmbListing && gmbListing.listingUrl) {
+        return gmbListing.listingUrl;
       }
 
       if (googlePlaceId) {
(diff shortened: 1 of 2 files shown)


mancusi:
can you update the ListingType to specify that the publisher may be null?

Also under what circumstances is the publisher null on a listing?

packages/pages/src/components/address/methods.ts
```diff
@@ -98,7 +98,7 @@ export const getDirections = (
         listings?.reduce(
           (obj, listing) => ({
             ...obj,
-            [listing.publisher.toUpperCase()]: listing.listingUrl,
+            [(listing.publisher || "").toUpperCase()]: listing.listingUrl,
```
tmeyer2115, line 101: It feels strange to have an `object` that can have `''` as a key. How do we handle if multiple listings are without a `publisher`? The same entry in the `object` will get clobbered. 

Actually, do we even care about the case where there is no `publisher`? It seems like below we're only looking for one key in the map: `ListingPublisherOption.GOOGLEMYBUSINESS`. Could we get rid of the `listingsMap` concept entirely and just do a `filter`?

mancusi (approved):

packages/pages/src/components/address/methods.ts
```diff
@@ -95,13 +95,15 @@ export const getDirections = (
...
-            [listing.publisher.toUpperCase()]: listing.listingUrl,
-          }),
-          {}
-        );
+        listings
+          ?.filter((listing) => listing && listing.publisher != null)
+          ?.reduce(
+            (obj, listing) => ({
+              ...obj,
+              [(listing.publisher || "").toUpperCase()]: listing.listingUrl,
```
tmeyer2115, line 103: Since we have the `filter` above, do we need the `|| ""` here? I'm still not sure we actually need to construct a full `listingsMap` either. It seems like `listingsMap` is only used on lines 107 and 108. That is, we just check if GMB is in `listingsMap` and use it if it is. 

Instead of creating `listingsMap`, could you just do:
```
const gmbListing = listings.find(listing => 
    listing && listing.publisher === listingsMap[ListingPublisherOption.GOOGLEMYBUSINESS);

gmbListing && return gmbListing;
...
[Rest of default case]
```

Jeffrey-Rhoads17: I believe we have the `listingsMap` in place for if/when we're going to add more listing publisher options (right now it only checks for Google, but that may change down the line)

tmeyer2115 (approved):

PR #259, opened 2022-11-17
fix(dev): add back favicon support in dev mode

oshi97:
It appears we can't use loadTemplateModules() in dev mode. If we try to do so we lose support for importing .ico files (and I imagine other kinds of loaders too).

Instead, calling loadTemplateModule() (which is not used in loadTemplateModules()) in a loop works since it relies on vite to import modules.

J=[SlackThread](https://yext.slack.com/archives/C016ZKY42CF/p1668696423830079?thread_ts=1668627719.720419&cid=C016ZKY42CF)
TEST=manual

tested that I can import .ico favicons in the pages react starter, both dev mode and prod mode (yext pages generate -> yext pages serve) work

### DIFF
--- a/packages/pages/src/dev/server/ssr/generateTestData.ts
+++ b/packages/pages/src/dev/server/ssr/generateTestData.ts
@@ -70,10 +70,6 @@ export const generateTestDataForSlug = async (
   return getDocumentByLocale(parsedData, locale);
 };
 
-/**
- * In dev mode, we cannot use the loadTemplateModules that uses esbuild.
- * Instead we have to rely on vite to import templates.
- */
 const loadTemplateModuleCollectionUsingVite = async (
   vite: ViteDevServer,
   templateFilepaths: string[]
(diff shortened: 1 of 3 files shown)


packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -62,6 +70,31 @@ export const generateTestDataForSlug = async (
   return getDocumentByLocale(parsedData, locale);
 };
 
+/**
+ * In dev mode, we cannot use the loadTemplateModules that uses esbuild.
+ * Trying to do so results in losing loader support for .ico files and possibly
+ * other file types.
+ * Instead we have to rely on vite to import templates.
+ */
+const loadTemplateModuleCollectionUsingVite = async (
```
mkilpatrick, line 79: I'm confused why ico files are relevant to generateTestData. The only thing we're trying to do is load the modules in order to get the config off of them to pass to generateTestData.

oshi97: I feel like this is fine for now since this matches what we were doing before - if we want to update serverRenderSlugRoute to use ts-morph or loadTemplateModules we'd want serverRenderRoute to also match which I think is out of scope here

mkilpatrick: Fair enough

mkilpatrick (approved):

PR #264, opened 2022-11-29
feat(dev): use new available port if default dev server port is occupied

yen-tt:
Update pages dev command to use new port if the default port 5173 is already occupied.

J=SLAP-2461
TEST=manual

- run `pages dev` and see that the default port is used as normal.
- run `npx serve -p 5173` to occupy the default port. Then run `pages dev` and see that a random available port is use (instead of erroring out and exit like before), and the info text also display the new port.

### DIFF
--- a/packages/pages/src/dev/server/middleware/constants.ts
+++ b/packages/pages/src/dev/server/middleware/constants.ts
@@ -1,12 +1,17 @@
-export const defaultDevServerPort = 5173;
+import getPort, { portNumbers } from "get-port";
+
+// Will use any available port from 5173 to 6000, otherwise fall back to a random port
+export const devServerPort = await getPort({
+  port: portNumbers(5173, 6000),
+});
 
 export const dynamicModeInfoText = `Dynamic mode enabled. Below are sample URLs generated based on the contents 
 of the localData folder, but entity data will be re-fetched on each page load and reflect updates in real time. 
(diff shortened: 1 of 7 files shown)


packages/pages/src/dev/server/middleware/constants.ts
```diff
@@ -1,12 +1,12 @@
-export const viteDevServerPort = 5173;
+export const defaultDevServerPort = 5173;
```
oshi97, line 1: Would it be simpler to set devServerPort as a constant here like `export const devServerPort = getPort(`? Then we wouldn't have to pass it as a param to multiple places

packages/pages/src/dev/server/server.ts
```diff
@@ -54,6 +55,11 @@ export const createServer = async (
     ));
   }
 
+  // Will use defaultDevServerPort if available, otherwise fall back to a random available port
```
oshi97, line 58: Is it possible to choose the next sequential port? For instance if 5173 is occupied we pick 5174 and so on. I looked through the docs for get-port and didn't see a flag for this, if it's annoying it's not super important.

yen-tt: updated! will use the next sequential port from 5173 to 6000 (have to pick the last port of the range.), otherwise, it will fallback to a random port.

oshi97 (approved):
🇵🇹

PR #266, opened 2022-12-01
docs: link changelog from readme

mancusi:


README.md
```diff
@@ -29,6 +29,9 @@ PagesJS is a collection of tools that make it easy to develop on [Yext Pages](ht
 | [getRuntime()](https://github.com/yext/pages/blob/main/packages/pages/src/util/README.md#getRuntime)     |
 | [isProduction()](https://github.com/yext/pages/blob/main/packages/pages/src/util/README.md#isProduction) |
 
+## Changelog
+See [Changelog](https://github.com/yext/pages/blob/main/packages/pages/CHANGELOG.md).
```
yen-tt, line 33: Nice! I noticed an inconsistency in the headers for the recent beta releases, they are much smaller compare to `1.0.0-beta.5 (2022-08-30)` and the older versions. Could you update the recent headers to be bigger like [here](https://github.com/yext/pages/blob/main/packages/pages/CHANGELOG.md#100-beta5-2022-08-30) in changelog file? (e.g. `#### 1.0.0-beta.6 (2022-09-08)` to `# 1.0.0-beta.6 (2022-09-08`)

mancusi: this was done as part of https://github.com/yext/pages/pull/204 and isn't relevant to this review.

@mkilpatrick do you know what the rebase issue was?

mkilpatrick: I changed it because the other changelog package didn't work right. It pulled in every PR into every release. I found this latter package actually works, but agree that the formatting of it isn't as nice.

mkilpatrick: Btw, the changelog is already linked on line 15

PR #272, opened 2022-12-05
fix(utils): improve runtime environment checking

### LINKED ISSUE #244: Runtime constructor errors under certain conditions
### Describe the bug

I am calling the util `getRuntime` function. I expect no errors to be thrown but I see an error `Uncaught TypeError: process.versions is undefined`. 

This is happening because of this line: 
https://github.com/yext/pages/blob/main/packages/pages/src/util/runtime.ts#L14

There is a check prior to that line `if (typeof process !== "undefined")` but there are cases where this variable can be defined in the browser context. 

In the situation I ran into, it's because dom elements with an `id` automatically create global variables of the same name (https://html.spec.whatwg.org/#dom-window-nameditem). So an element `<div id="process">` will cause the if check to pass but then error because `process.versions` is undefined. 

### Reproduction

https://tmp-knowledge-telescope-yext-com.preview.pagescdn.com/board/8587639397247715951 click the "Test Card"

### System Info

```shell
n/a
```

### Used Package Manager

yarn

### Used Browser

Firefox

### Validations

tatimblin:
fix #244 
Employ more comprehensive environment checking following the checks in [browser-or-node](https://github.com/flexdinesh/browser-or-node/blob/master/src/index.js)

J=SLAP-2435
TEST=manual

create and install pages binary in starter, compared old check to new check with a global variable `process` as described in Github issue.

### DIFF
--- a/packages/pages/src/util/runtime.ts
+++ b/packages/pages/src/util/runtime.ts
@@ -9,7 +9,7 @@ class Runtime {
   version: string;
 
   constructor() {
-    if (typeof process !== "undefined") {
+    if (this.isNodeProcess()) {
       this.name = "node";
       this.version = process.versions.node;
       this.isServerSide = true;
@@ -25,6 +25,13 @@ class Runtime {
       }
       this.isServerSide = true;
     }
+  }
+
+  isNodeProcess(): boolean {
+    return (
+      typeof process !== "undefined" &&
+      process.release?.name?.search(/node|io.js/) !== -1
+    );
   }
 
   getNodeMajorVersion(): number {
--- a/packages/pages/src/util/runtime.ts
+++ b/packages/pages/src/util/runtime.ts
@@ -1,5 +1,5 @@
 class Runtime {
-  name: "node" | "deno" | "browser";
+  name: "node" | "deno" | "browser" | "unknown";
   /**
    * Whether or not the current runtime is being executed server-side or client-side. If the runtime
    * is node or deno then isServerSide will be true. When the runtime is browser isServerSide is
@@ -9,29 +9,33 @@ class Runtime {
   version: string;
 
   constructor() {
-    if (this.isNodeProcess()) {
-      this.name = "node";
-      this.version = process.versions.node;
-      this.isServerSide = true;
-    } else if (typeof window !== "undefined" && !("Deno" in window)) {
+    if (this.isBrowser()) {
       this.name = "browser";
       this.version = navigator.userAgent;
       this.isServerSide = false;
-    } else {
+      return;
+    }
+
+    if (this.isNode()) {
+      this.name = "node";
+      this.version = process.versions.node;
+      this.isServerSide = true;
+      return;
+    }
+
+    if (this.isDeno()) {
       this.name = "deno";
       this.version = "";
       if (typeof window !== "undefined") {
         this.version = (window as any).Deno?.version.deno || "";
       }
       this.isServerSide = true;
+      return;
     }
-  }
 
-  isNodeProcess(): boolean {
-    return (
-      typeof process !== "undefined" &&
-      process.release?.name?.search(/node|io.js/) !== -1
-    );
+    this.name = "unknown";
+    this.isServerSide = true;
+    this.version = "";
   }
 
   getNodeMajorVersion(): number {
@@ -40,6 +44,33 @@ class Runtime {
     }
 
     return +this.version.split(".")[0];
+  }
+
+  isBrowser(): boolean {
+    return (
+      typeof window != "undefined" &&
+      typeof window.document !== "undefined" &&
+      !("Deno" in window)
+    );
+  }
+
+  isDeno(): boolean {
+    // @ts-expect-error Deno may be undefined
+    return (
+      typeof Deno !== "undefined" &&
+      // @ts-expect-error Deno may be undefined
+      typeof Deno.version !== "undefined" &&
+      // @ts-expect-error Deno may be undefined
+      typeof Deno.version.deno !== "undefined"
+    );
+  }
+
+  isNode(): boolean {
+    return (
+      typeof process !== "undefined" &&
+      process.versions != null &&
+      process.versions.node != null
+    );
   }
 }
 
--- a/packages/pages/src/util/runtime.ts
+++ b/packages/pages/src/util/runtime.ts
@@ -1,12 +1,12 @@
 class Runtime {
-  name: "node" | "deno" | "browser" | "unknown";
+  name?: "node" | "deno" | "browser";
   /**
    * Whether or not the current runtime is being executed server-side or client-side. If the runtime
    * is node or deno then isServerSide will be true. When the runtime is browser isServerSide is
    * false.
    */
-  isServerSide: boolean;
-  version: string;
+  isServerSide?: boolean;
+  version?: string;
 
   constructor() {
     if (this.isBrowser()) {
@@ -32,15 +32,15 @@ class Runtime {
       this.isServerSide = true;
       return;
     }
-
-    this.name = "unknown";
-    this.isServerSide = true;
-    this.version = "";
   }
 
   getNodeMajorVersion(): number {
     if (this.name != "node") {
       throw new Error("Not running in Node.");
+    }
+
+    if (!this.version) {
+      throw new Error("No version found.");
     }
 
     return +this.version.split(".")[0];
--- a/packages/pages/src/util/runtime.ts
+++ b/packages/pages/src/util/runtime.ts
@@ -55,8 +55,8 @@ class Runtime {
   }
 
   isDeno(): boolean {
-    // @ts-expect-error Deno may be undefined
     return (
+      // @ts-expect-error Deno may be undefined
       typeof Deno !== "undefined" &&
       // @ts-expect-error Deno may be undefined
       typeof Deno.version !== "undefined" &&
--- a/packages/pages/src/components/analytics/provider.tsx
+++ b/packages/pages/src/components/analytics/provider.tsx
@@ -42,7 +42,7 @@ export function AnalyticsProvider(
   }
 
   let enableDebuggingDefault = debuggingParamDetected();
-  if (typeof process !== "undefined") {
+  if (getRuntime().name === "node") {
     enableDebuggingDefault =
       enableDebuggingDefault || process.env?.NODE_ENV === "development";
   }


packages/pages/src/util/runtime.ts
```diff
@@ -1,29 +1,36 @@
 class Runtime {
-  name: "node" | "deno" | "browser";
+  name?: "node" | "deno" | "browser";
```
oshi97, line 2: should name ever be undefined?

packages/pages/src/util/runtime.ts
```diff
@@ -32,8 +39,39 @@ class Runtime {
...
+      // @ts-expect-error Deno may be undefined
+      typeof Deno !== "undefined" &&
+      // @ts-expect-error Deno may be undefined
+      typeof Deno.version !== "undefined" &&
+      // @ts-expect-error Deno may be undefined
+      typeof Deno.version.deno !== "undefined"
+    );
+  }
+
+  isNode(): boolean {
```
nmanu1, line 68: out of curiosity, why don't we use browser-or-node directly instead of writing it ourselves?

packages/pages/src/util/runtime.ts
```diff
@@ -1,29 +1,36 @@
...
-  name: "node" | "deno" | "browser";
+  name?: "node" | "deno" | "browser";
   /**
    * Whether or not the current runtime is being executed server-side or client-side. If the runtime
    * is node or deno then isServerSide will be true. When the runtime is browser isServerSide is
    * false.
    */
-  isServerSide: boolean;
-  version: string;
+  isServerSide?: boolean;
```
mkilpatrick, line 8: When are these optional?

packages/pages/src/util/runtime.ts
```diff
@@ -32,8 +39,39 @@ class Runtime {
...
+    }
+
     return +this.version.split(".")[0];
   }
+
+  isBrowser(): boolean {
+    return (
+      typeof window != "undefined" &&
+      typeof window.document !== "undefined" &&
+      !("Deno" in window)
```
mkilpatrick, line 53: Maybe the final check should be `&& !isDeno() && !isNode()`

packages/pages/src/util/runtime.ts
tatimblin: just to limit dependencies we have it's so small

oshi97: browser-or-node seems pretty tiny though? https://github.com/flexdinesh/browser-or-node/blob/master/src/index.js

mkilpatrick: I didn't know this existed when I made runtime. We could probably just use it. Though it's a breaking change to remove runtime completely, but we're in beta.

One thing ours has is that you can get the node version which I think is checked somewhere. I could be wrong and I had just added the version for fun in case we ever needed it. You'll have to check usages.

oshi97: We could definitely keep `getRuntime` and change the implementation to use browser-or-node. @tatimblin maybe we'd want to ask product what they would want here? I didn't realize this was a public export.

packages/pages/src/util/runtime.ts
tatimblin: Removed `isBrowser` check because it will be the default case since we want to confirm at least one environment is set

packages/pages/src/util/runtime.ts
tatimblin: I switched over to using the package, but since it exports variables instead of functions I was unable to mock different environments. Switched back to the calls I originally had.

packages/pages/src/util/runtime.test.ts
```diff
@@ -46,104 +57,53 @@ describe("runtime", () => {
...
-        ({
-          ...originalWindow,
-          Deno: {
-            version: {
-              deno: "1.24.0",
-            },
-          },
-        } as any)
-    );
+    global.Deno = {
```
oshi97, line 67: If we used the third party lib, I think you could mock the third party lib itself with jest.mock() instead of having to manually specify these globals? The benefit there is that we don't have to manually mock low level details like the contents of global.Deno, which could change or be slightly different than we expect. The downside is that we have to trust *somebody else* to do that, but the lib Nidhi found has pretty good usage so I think it's trustable?

oshi97: If you can imagine a scenario where a new version of Deno changes what stuff they provide in the global context, it would be annoying to have to come back and have to retest and update things.

tatimblin: I believe mocking the module will be a mess. Once mocked to return true for `isNode` or `isDeno`, we'd still have to mock the global variable because they are read inside the if statement to get the version i.e. `if (isDeno) this.version = Deno.version.deno;`. I would rather just mock the global.
While their globals changing would break the tests, it would be a major breaking change (node 20 or deno 2) and we wouldn't support those right away anyways

oshi97: I believe you can just call jest.mock() at the top and then jest.spyOn().mockImplementation() within each test case?
I don't think it should be messy since we regularly mock search-headless-react's useAnswersState in the component library.
https://github.com/yext/search-ui-react/blob/main/tests/components/DirectAnswer.test.tsx

packages/pages/src/util/runtime.test.ts
```diff
@@ -168,3 +96,45 @@ describe("runtime", () => {
...
+      },
+    };
+    (browserOrNode.isDeno as any) = true;
+    (browserOrNode.isNode as any) = false;
+  }
+
+  if (simulatedEnvironment === "node") {
+    global.process = {
+      ...originalProcess,
+      versions: {
```
oshi97, line 119: this might be a dumb question but why do we need to add these to global.process?

packages/pages/src/util/runtime.test.ts
```diff
@@ -168,3 +96,45 @@ describe("runtime", () => {
     global.process = originalProcess;
   });
 });
+
+const setupMockEnvironment = (
+  simulatedEnvironment?: "node" | "deno"
+): NodeJS.Process => {
+  const originalProcess = process;
```
oshi97, line 103: Would something like the below work?

```ts
import browserOrNode from 'browser-or-node'

const setupMockEnvironment = (env: 'node' | 'deno') => {
  jest.spyOn(browserOrNode).mockImplementation(() => {
    // mock whatever you want directly based on whether env is deno or node
  })
}

File: packages/pages/src/components/analytics/analytics.test.tsx [hunk 8]
```diff
@@ -10,13 +10,31 @@ import { AnalyticsProvider } from "./provider.js";
...
+});
+
 // The following section of mocks just exists to supress an error that occurs
 // because jest does not implement a window.location.navigate.  See:
 // https://www.benmvp.com/blog/mocking-window-location-methods-jest-jsdom/
 // for details.
 const oldWindowLocation = global.location;
 
 beforeAll(() => {
+  global.process = {
```
oshi97, line 32: Would it make sense to add an isDevelopment flag to runtime, since there are eccentricities with using process directly? Also little easier to mock lol

packages/pages/src/util/runtime.test.ts
tatimblin: Because the class will look for it on line 14

packages/pages/src/util/runtime.test.ts
tatimblin: I had attempted that as you suggested in a prior comment. The error from jest states you cannot mock a boolean literal only a function that returns a boolean (or whatever else what you're mocking returns).

packages/pages/src/components/analytics/analytics.test.tsx
tatimblin: isDevelopment as in when running tests on jest, or as in running the development server?

oshi97: Adding an isDevelopment to the Runtime class (the latter)

packages/pages/src/util/runtime.test.ts
oshi97: Oooh gotcha. I gave it a quick try and I think something doMock should work and would simplify this a bit?
```ts

const setupMockEnvironment = (
  simulatedEnvironment?: "node" | "deno"
) => {
  jest.doMock('browser-or-node', () => ({
    isDeno: simulatedEnvironment === "deno",
    isNode: simulatedEnvironment === "node"
  }))
};
```

tatimblin: I don't think this is technically any different than mocking at the beginning of the file and then setting the value of both variables in their if statements. I would like to continue with the current approach because it allows an environment to be entirely set in a single scope. This makes it easier to understand instead of splitting it into multiple environment checks.

packages/pages/src/components/analytics/analytics.test.tsx
tatimblin: Certain parts of pagesjs are expected to run server side in node, deno, or both. Passing in that flag would mean we'd have to figure out the runtime outside of the runtime class

packages/pages/src/util/runtime.ts
```diff
@@ -9,22 +11,26 @@ class Runtime {
...
-      this.isServerSide = true;
-    } else if (typeof window !== "undefined" && !("Deno" in window)) {
-      this.name = "browser";
-      this.version = navigator.userAgent;
-      this.isServerSide = false;
-    } else {
+    if (isDeno) {
       this.name = "deno";
       this.version = "";
       if (typeof window !== "undefined") {
```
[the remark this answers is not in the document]
tatimblin: I'll use browser-or-node's `isBrowser` check instead

packages/pages/src/util/runtime.test.ts
```diff
@@ -29,126 +36,49 @@ describe("runtime", () => {
...
-    windowSpy.mockImplementation(() => undefined as any);
-
-    const runtime = getRuntime();
-
-    expect(runtime.name).toEqual("deno");
-    expect(runtime.version).toEqual("");
-
-    global.process = originalProcess;
-    windowSpy.mockRestore();
+    global.Deno = undefined;
```
oshi97, line 57: Can we have this cleanup in a beforeEach or afterEach?

packages/pages/src/util/runtime.test.ts
oshi97: I feel like jest.doMock is more explicit that you're mocking something, as opposed to taking `browserOrNode`, asserting it to be `any` (which I think eslint will complain about if we add the @yext/eslint-config rules?)

packages/pages/src/util/runtime.test.ts
```diff
@@ -1,24 +1,31 @@
...
+jest.mock("browser-or-node", () => ({
+  __esModule: true,
+  isNode: false,
+  isDeno: true,
+  isBrowser: true,
+}));
+
+declare global {
+  // eslint-disable-next-line no-var
+  var Deno:
```
oshi97, line 16: Can this be const?

tatimblin: I can't find documentation that says it has to be, but it will cause any reference of `Deno` in the file to show an error https://www.typescriptlang.org/docs/handbook/namespaces.html#ambient-namespaces

packages/pages/src/util/runtime.test.ts
tatimblin: Is that snippet something you tried yourself? Trying it where you indicated I do not see the mock working @oshi97

packages/pages/src/util/runtime.test.ts
```diff
@@ -168,3 +97,45 @@ describe("runtime", () => {
...
+        modules: "foo",
+        openssl: "foo",
+      },
+    };
+    (browserOrNode.isDeno as any) = false;
+    (browserOrNode.isNode as any) = true;
+  }
+
+  if (!simulatedEnvironment) {
+    (browserOrNode.isDeno as any) = false;
```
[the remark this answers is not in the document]
tatimblin: I reduced the number of times one of those calls were made by moving to `afterEach` which hopefully resolves your concern over similar code being duplicated. Now each environment only needs one mock to be set. This lets me keep all the environment setting for a particular environment in the same code block as I [described in this comment](https://github.com/yext/pages/pull/272/files#r1043435596)

packages/pages/src/util/runtime.ts
```diff
@@ -9,22 +11,26 @@ class Runtime {
...
-      this.isServerSide = false;
-    } else {
+    if (isDeno) {
       this.name = "deno";
       this.version = "";
-      if (typeof window !== "undefined") {
+      if (isBrowser) {
         this.version = (window as any).Deno?.version.deno || "";
       }
       this.isServerSide = true;
```
oshi97, line 20: I might be misunderstanding but it looks like `runtime.isServerSide` can be true when `isBrowser` is true since we always set isServerSide to true here?

packages/pages/src/util/runtime.ts
```diff
@@ -9,22 +11,26 @@ class Runtime {
...
+      return;
+    }
+
+    if (isNode) {
```
oshi97, line 24: Nit but can we use an `else if` or switch statement? I miss the return statement sometimes when I read this.

packages/pages/src/util/runtime.test.ts
oshi97: TIL to use doMock() you have to reimport modules after calling it

packages/pages/src/util/runtime.ts
```diff
@@ -9,22 +11,26 @@ class Runtime {
...
-      this.version = process.versions.node;
-      this.isServerSide = true;
-    } else if (typeof window !== "undefined" && !("Deno" in window)) {
-      this.name = "browser";
-      this.version = navigator.userAgent;
-      this.isServerSide = false;
-    } else {
+    if (isDeno) {
       this.name = "deno";
       this.version = "";
```
oshi97, line 16: Just double checking that it's expected that `version` is a blank string when not in the browser?

tatimblin: That was the existing functionality, so it is expected in that sense. I can't speak much more to its reasoning

packages/pages/src/util/runtime.ts
tatimblin: Javascript can be run server side in the deno environment and then included on the rendered web page to be viewed in the browser

oshi97: When it's included on the rendered web page, wouldn't the user want isServerSide to be false though?

oshi97: If you're fine as is I'm ok with it

tatimblin: I would like to limit the scope of my changes to addressing the false positive possible in a node environment, and not change the existing functionality

oshi97 (dismissed):

packages/pages/src/util/runtime.ts
```diff
@@ -9,21 +11,21 @@ class Runtime {
...
+      if (isBrowser) {
+        this.version = (window as any).Deno?.version.deno || "";
+      }
+      this.isServerSide = true;
+    } else if (isNode) {
       this.name = "node";
       this.version = process.versions.node;
       this.isServerSide = true;
-    } else if (typeof window !== "undefined" && !("Deno" in window)) {
+    } else {
```
mkilpatrick, line 25: Should we check isBrowser here and add a new else condition that sets things to "unknown" or something like that?

tatimblin: I considered that and we could do it, but here was my reasoning. Javascript was created for the browser and spent its first 15 years only being run in the browser. Because of that origin it can be assumed that Javascript is being run in the browser, if no node or deno artifacts are detected.

oshi97 (approved):

PR #275, opened 2022-12-06
feat(dev): fix linux support + static site acceptance across OS

oshi97:
This PR adds back support for Linux that was removed in https://github.com/yext/pages/pull/276.
The PR linked above added support for Windows by removing a bash script (since Windows can't run
bash except through WSL) but removed Linux support (or at least Ubuntu support) since Ubuntu
does not parse the node shebang the same way as windows + mac, and would drop certain
important experimental NodeJS flags.

This PR adds those back by using a node script to spawnSync the PagesJS CLI. 

In hopes of preventing these types of issues, this PR adds a basic static site to the monorepo, and a basic acceptance test that ensures the dev server is able to start up. This test is run on the 3 major OS through GIthub Actions.
There were a lot of difficulties getting this to work with windows-latest on Github Actions.

J=SLAP-2496
TEST=auto,manual

installed this version of pages on my physical windows machine and saw the dev command serve pages correctly
tested this version on Ubuntu 22.04 through VirtualBox
saw that the test will fail on windows/ubuntu if the flags are not passed or a bash script is used

PR #276, opened 2022-12-07
fix(dev): fix dev command on Windows

oshi97:
Removes pages.sh because bash scripts don't work on Windows. A follow up PR will add an acceptance test so this doesn't break again.

J=SLAP-2496
TEST=manual

pages dev works on my physical windows machine after this change

### DIFF
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -50,7 +50,7 @@
     "prepare": "pnpm build"
   },
   "bin": {
-    "pages": "dist/bin/pages.sh"
+    "pages": "dist/bin/pages.js"
   },
   "files": [
     "dist/**/*",
--- a/packages/pages/src/bin/pages.sh
+++ b/packages/pages/src/bin/pages.sh
@@ -1,2 +1,0 @@
-#!/usr/bin/env bash
-/usr/bin/env node --experimental-specifier-resolution=node --experimental-vm-modules node_modules/@yext/pages/dist/bin/pages.js "$@"
--- a/packages/pages/src/bin/pages.ts
+++ b/packages/pages/src/bin/pages.ts
@@ -1,3 +1,5 @@
+#!/usr/bin/env node --experimental-specifier-resolution=node --experimental-vm-modules
+
 import { initCommandModule } from "../init/init.js";
 import { devCommandModule } from "../dev/dev.js";
 import { generateCommandModule } from "../generate/generate.js";
--- a/packages/pages/src/bundler.js
+++ b/packages/pages/src/bundler.js
@@ -21,7 +21,7 @@ const testFilter = (f) =>
   !f.endsWith(".test.ts") &&
   !f.endsWith(".test.tsx") &&
   !f.endsWith(".test.js");
-const filters = new Set(["./src/bundler.js", "./src/bin/pages.sh"]);
+const filters = new Set(["./src/bundler.js"]);
 
 const files = glob
   .sync("./src/**/*\\.*")
@@ -75,11 +75,3 @@ try {
 } catch (e) {
   console.error(e);
 }
-
-/**
- * Copy the sh binary.
- */
-copyFileSync(
-  path.resolve("./src/bin/pages.sh"),
-  path.resolve("./dist/bin/pages.sh")
-);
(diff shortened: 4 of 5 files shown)


oshi97:
> FYI I believe that this was added to fix an issue with this running on certain linux distros [123ed61](https://github.com/yext/pages/commit/123ed6198a30b31f6df8aeac65e616fe87f2542d)

We're aware of this but wanted to get the Windows fix out first. We couldn't replicate the issue on Ubuntu, and if the issue is only on Kubuntu it's probably better to support Windows in the meantime.

yen-tt (approved):

mkilpatrick:
> > FYI I believe that this was added to fix an issue with this running on certain linux distros [123ed61](https://github.com/yext/pages/commit/123ed6198a30b31f6df8aeac65e616fe87f2542d)
> 
> We're aware of this but wanted to get the Windows fix out first. We couldn't replicate the issue on Ubuntu, and if the issue is only on Kubuntu it's probably better to support Windows in the meantime.

I agree with Justin that this now [potentially] reverts a fix for Linux. It wasn't specific to Kubuntu. At least I reproduced the problem and fixed it for Ubuntu. What distro did you try and confirm that this change still works for it?

I believe it was related to multiple args not correctly being passed through - https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e/26056686#26056686

Anywho, I don't think breaking one OS for another is a great idea, but it appears that ship has already sailed in the past 5 min.

oshi97:
> 

The acceptance test I'm adding in https://github.com/yext/pages/pull/275 passes when run on ubuntu-latest.
It will also fail if pages.js is run without the flags, so I assumed pages dev works on ubuntu. Would you mind letting me know how you reproduced the issue on Ubuntu? I just started downloading and setting up VirtualBox with Ubuntu but not sure if that's the best way.

oshi97:
> > > FYI I believe that this was added to fix an issue with this running on certain linux distros [123ed61](https://github.com/yext/pages/commit/123ed6198a30b31f6df8aeac65e616fe87f2542d)
> > 
> > 
> > We're aware of this but wanted to get the Windows fix out first. We couldn't replicate the issue on Ubuntu, and if the issue is only on Kubuntu it's probably better to support Windows in the meantime.
> 
> I agree with Justin that this now [potentially] reverts a [fix](https://github.com/yext/pages/pull/199/files) for Linux. It wasn't specific to Kubuntu. At least I reproduced the problem and fixed it for Ubuntu. What distro did you try and confirm that this change still works for it?
> 
> I believe it was related to multiple args not correctly being passed through - https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e/26056686#26056686
> 
> Anywho, I don't think breaking one OS for another is a great idea, but it appears that ship has already sailed in the past 10 min.
@mkilpatrick 
The acceptance test I'm adding in https://github.com/yext/pages/pull/275 passes when run on ubuntu-latest.
It will also fail if pages.js is run without the flags, so I assumed pages dev works on ubuntu. Would you mind letting me know how you reproduced the issue on Ubuntu? I just started downloading and setting up VirtualBox with Ubuntu but not sure if that's the best way.

oshi97:
> > > FYI I believe that this was added to fix an issue with this running on certain linux distros [123ed61](https://github.com/yext/pages/commit/123ed6198a30b31f6df8aeac65e616fe87f2542d)
> > 
> > 
> > We're aware of this but wanted to get the Windows fix out first. We couldn't replicate the issue on Ubuntu, and if the issue is only on Kubuntu it's probably better to support Windows in the meantime.
> 
> I agree with Justin that this now [potentially] reverts a [fix](https://github.com/yext/pages/pull/199/files) for Linux. It wasn't specific to Kubuntu. At least I reproduced the problem and fixed it for Ubuntu. What distro did you try and confirm that this change still works for it?
> 
> I believe it was related to multiple args not correctly being passed through - https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e/26056686#26056686
> 
> Anywho, I don't think breaking one OS for another is a great idea, but it appears that ship has already sailed in the past 10 min.

The acceptance test I'm adding in https://github.com/yext/pages/pull/275 passes when run on ubuntu-latest.
It will also fail if pages.js is run without the flags, so I assumed pages dev works on ubuntu. Would you mind letting me know how you reproduced the issue on Ubuntu? I just started downloading and setting up VirtualBox with Ubuntu but not sure if that's the best way.

PR #282, opened 2022-12-22
chore(dev/plugin): add file extension to node_module/file import

tatimblin:
file extensions are required for relative imports but not for node_modules. react-dom/server is unique because it is a node_modules import, but it references a specific file. This is fine in Node 16-18, but errors in Node 19. This item fixes support for Node 19

J=SLAP-2501
TEST=manual

successfully served the site using Node 16, 17, 18, and 19

### DIFF
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -32,7 +32,7 @@
     }
   },
   "engines": {
-    "node": ">=17"
+    "node": "17"
   },
   "scripts": {
     "watch": "node src/bundler --watch",
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -32,7 +32,7 @@
     }
   },
   "engines": {
-    "node": "17"
+    "node": "^17 || ^18 || ^19"
   },
   "scripts": {
     "watch": "node src/bundler --watch",
--- a/.github/workflows/acceptance_test.yml
+++ b/.github/workflows/acceptance_test.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        node-version: [18.x]
+        node-version: [18.x, 19.x]
     runs-on: ${{ matrix.os }}
     defaults:
       run:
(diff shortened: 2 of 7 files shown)


yen-tt:
Did you look into the usage of `--experimental-specifier-resolution`? We use that in the page command, which exists in node [18](https://nodejs.org/docs/latest-v18.x/api/esm.html#customizing-esm-specifier-resolution-algorithm) and to lower versions to some extent, but it's dropped in node 19.

yen-tt:
Also, could you update the acceptance test workflow to run node 19 as well?

tatimblin:
> Did you look into the usage of `--experimental-specifier-resolution`? This experimental flag is use in the page command, which exists in node [18](https://nodejs.org/docs/latest-v18.x/api/esm.html#customizing-esm-specifier-resolution-algorithm) and to lower versions to some extent, but it's dropped in node 19.

I appears this [test hasn't succeeded](https://github.com/yext/pages/actions/workflows/acceptance_test.yml?page=2) in a while @yen-tt

tatimblin:
@yen-tt Looking more into the experimental node flags

It seems like the flag only does [one thing](https://nodejs.org/docs/latest-v18.x/api/cli.html#--experimental-specifier-resolutionmode), and now it is a requirement to specify the file extension in the import (including directory imports specifying index.js). For pagesjs Duval already made this change, so I am able to remove the flag and run the pages command in node 17. I've removed this flag since it is not needed anymore and `--experimental-vm-modules` is still available in node19, so I will leave it.

Nice call out!

packages/pages/package.json
```diff
@@ -32,7 +32,7 @@
     }
   },
   "engines": {
-    "node": ">=17"
+    "node": "17"
```
tatimblin, line 35: While we're working towards 18/19 support we should not state that we do. Updated based on [slack discussion](https://yext.slack.com/archives/C02UVSE7P6W/p1672183437986969?thread_ts=1672180916.657069&cid=C02UVSE7P6W)

yen-tt: the item for this PR is to have node 19 for PagesJS. Once this PR is merged, I don't think we need to restrict it to just 17. How about we narrow this to 17, 18, and 19.

yen-tt (approved):

PR #288, opened 2023-02-14
fix(dev): use Vite JS API for build command

nmanu1:
Use the Vite JS API instead of its CLI for the `pages build` command. This ensures that the version of Vite that's used to create the build is one that matches the PagesJS `vite` dependency, even if another version of Vite is installed in the repo.

J=SLAP-2621
TEST=manual

Use a packed version of PagesJS in a version of the locations starter running Vite 4 and see that Vite 3 is still used when `npx pages build` is run. Also tested that the scope flag still works for a multibrand repo.


mkilpatrick (approved):

packages/pages/src/build/build.ts
```diff
@@ -1,17 +1,15 @@
-import shell from "shelljs";
```
oshi97, line 1: I think we can remove the shelljs dep now!

packages/pages/src/build/build.ts
```diff
@@ -1,17 +1,15 @@
...
 const handler = async ({ scope }: BuildArgs) => {
   // Pass CLI arguments as env variables to use in vite-plugin
-  shell.exec("vite build", {
-    env: {
-      ...process.env,
-      YEXT_PAGES_SCOPE: scope,
-    },
-  });
+  if (scope) {
+    process.env.YEXT_PAGES_SCOPE = scope;
```
oshi97, line 10: Out of scope but it feels a little strange to me that we passed scope as an env variable here

packages/pages/src/build/build.ts
```diff
@@ -1,17 +1,15 @@
...
+  }
+  build();
```
tmeyer2115, line 12: When the `env` argument is passed to `shell.exec`, does it make the environment variables available to the child process under `process.env`? I'm only asking because I want to be sure it's okay that we're using `process.env` here and not a de-structured version.

nmanu1: Yes, `plugin.ts` accesses the env var as `process.env.YEXT_PAGES_SCOPE` even before this change, which is why `process.env` needed to be modified directly. Adding a variable to `process.env` like this only affects the current process, so after running the build script, `process.env.YEXT_PAGES_SCOPE` returns to being undefined if it wasn't defined before

packages/pages/src/build/build.ts
oshi97: I meant we can remove the dep from the package.json

oshi97: Don't worry I'm not colorblind

oshi97 (dismissed):

oshi97 (approved):
Not sure why snyk is upset we're only removing deps 🤷

packages/pages/src/build/build.ts
nmanu1: I poked around a little, but couldn't find another way to pass along CLI arguments if they weren't only meant to impact the vite config

PR #291, opened 2023-02-22
feat(dev): list serverless functions on index page

nmanu1:
Parse `pluginFileMap.json` and list the resulting serverless HTTP functions on the local dev index page.

J=SLAP-2568
TEST=manual

Point the locations starter to a local pack of PagesJS and see that all HTTP functions with paths are listed on the index page, and no event-driven functions.
Screen Shot 2023-02-23 at 1 41 13 PM

### DIFF
--- a/packages/pages/src/dev/server/middleware/indexPage.ts
+++ b/packages/pages/src/dev/server/middleware/indexPage.ts
@@ -16,6 +16,11 @@ import {
 import { ViteDevServer } from "vite";
 import { ProjectStructure } from "../../../common/src/project/structure.js";
 import { getTemplateFilepathsFromProjectStructure } from "../../../common/src/template/internal/getTemplateFilepaths.js";
+import {
+  EventType,
+  getPluginFileMap,
+  PluginSource,
+} from "../ssr/getPluginFileMap.js";
 
 type Props = {
   vite: ViteDevServer;
@@ -41,6 +46,7 @@ export const indexPage =
         vite,
         templateFilepaths
       );
+      const serverlessFuncs = Object.entries(getPluginFileMap());
 
       let indexPageHtml = index
         // Inject an informative message depending on if the user is in dynamic mode or not.
@@ -107,6 +113,33 @@ export const indexPage =
         );
       }
 
+      if (serverlessFuncs.length) {
+        const systemEventFuncs: Record<string, PluginSource> = {};
+        const apiFuncs: Record<string, PluginSource> = {};
+        serverlessFuncs.forEach(([funcName, source]) => {
+          switch (source.event) {
+            case EventType.ON_URL_CHANGE:
+            case EventType.ON_PAGE_GENERATE:
+              systemEventFuncs[funcName] = source;
+              break;
+            case EventType.API:
+              apiFuncs[funcName] = source;
+              break;
+            default:
+              console.error(
+                `Error in serverless function ${funcName}. It will not be listed on the index page.`
+              );
+              break;
+          }
+        });
+        indexPageHtml = indexPageHtml.replace(
+          `<!--serverless-functions-html-->`,
+          `<div class="section-title">Functions</div>
+          ${createSystemEventFuncList(systemEventFuncs)}
+          ${createApiFuncList(apiFuncs)}`
+        );
+      }
+
       if (displayGenerateTestDataWarning) {
         // If there was an issue regenerating the local test data on dev server start, then
         // display a warning message that informs the user. This will only be displayed when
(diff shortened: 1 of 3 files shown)


packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -107,6 +113,33 @@ export const indexPage =
         );
       }
 
+      if (serverlessFuncs.length) {
+        const systemEventFuncs: Record<string, PluginSource> = {};
+        const apiFuncs: Record<string, PluginSource> = {};
```
tmeyer2115, line 118: Nit: Can we call this `httpFuncs`? Product is moving away from the "Api" verbiage in favor of "Http".

packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -129,6 +162,54 @@ export const indexPage =
...
+  return Object.keys(systemEventFuncs).length
+    ? `<div class="list">
+      <div class="list-title">System Events</div>
+      <ul>${Object.values(systemEventFuncs).reduce(
+        (funcAccumulator, source) => {
+          return (
+            funcAccumulator +
+            `<li>
+            ${
+              source.event === EventType.ON_PAGE_GENERATE
```
tmeyer2115, line 177: Could we do something a bit safer here? In the future, if a new event type is added, it could be easy to forget to update this logic.

packages/pages/src/dev/server/ssr/getPluginFileMap.ts
```diff
@@ -0,0 +1,44 @@
...
+/** The type of event for the serverless function. */
+export enum EventType {
+  /** The function is in an error state. */
+  UNKNOWN_EVENT = 0,
+  /** A function that is invoked on a URL change event. */
+  ON_URL_CHANGE = 1,
+  /** A function that is invoked on a page generate event. */
+  ON_PAGE_GENERATE = 2,
+  /** An API (or HTTP) function. */
+  API = 3,
```
tmeyer2115, line 17: I think we should double check the name API with Spruce. As mentioned above, Product has stopped using this term in favor of HTTP.

packages/pages/src/dev/server/ssr/getPluginFileMap.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/** Source data associated with a serverless function. */
+export type PluginSource = {
```
tmeyer2115, line 21: Could we call this `FunctionSource` or `ServerlessFunctionSource`? I think it's bad enough that we use the terms Plugin and Serverless Function interchangeably in the platform. In our code, I'd like us to pick one term.

packages/pages/src/dev/server/middleware/indexPage.ts
nmanu1: This logic should be removed in the next few days once Spruce finishes the item for passing the event display name in `pluginFileMap.json`. We would directly render that here instead of conditioning on event type. I can add a `TODO` comment to call attention to it

packages/pages/src/dev/server/ssr/getPluginFileMap.ts
```diff
@@ -0,0 +1,44 @@
+import path from "path";
+import fs from "fs";
+
+const PLUGIN_FILE_MAP_PATH = path.resolve(
```
oshi97, line 4: Is this path.resolve necessary?

packages/pages/src/dev/server/ssr/getPluginFileMap.ts
```diff
@@ -0,0 +1,44 @@
...
+  /** {@inheritDoc EventType} */
+  event: EventType;
+  /** The endpoint path if it is an API function. */
+  serverlessFunctionPath?: string;
+};
+
+/** Returns the pluginFileMap as a mapping of function name to source data. */
+export const getPluginFileMap = (): Record<string, PluginSource> => {
+  try {
+    return JSON.parse(fs.readFileSync(PLUGIN_FILE_MAP_PATH).toString());
```
oshi97, line 35: I think we normally specify utf-8 encoding with `rs.readFileSync(filepath, 'utf-8')` instead of calling toString(), though I have no idea which is better and it probably doesn't matter

packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -107,6 +113,33 @@ export const indexPage =
         );
       }
 
+      if (serverlessFuncs.length) {
```
oshi97, line 116: Could we move this new logic into a separate function? I was thinking we could even move all the serverless funcs stuff into a separate file but I'm also fine keeping it here.
Imo each of the preexisting indexPageHtml.replace calls should have been separated out instead of this giant templating script (out of scope though!)
Eventually I think we'll want to refactor this index page to use react, it feels a little backward to have a js library that renders react pages but uses raw js templating to create the index page.

packages/pages/src/dev/server/ssr/getPluginFileMap.ts
nmanu1: I used `toString` to be consistent with all the other uses of `readFileSync` in the repo

packages/pages/src/dev/server/middleware/indexPage.ts
oshi97: I made an item to update the index page to use react https://yexttest.atlassian.net/browse/SLAP-2642

packages/pages/src/dev/server/ssr/getPluginFileMap.ts
nmanu1: Removed!

packages/pages/src/dev/server/middleware/indexPage.ts
nmanu1: Moved it into a separate function, but kept it in the same file for now so we can refactor it all from one place

oshi97 (approved):

lambdaFun94:
@nmanu1 In the screenshot above, I do not see the Http functions as blue hyperlinks. Will that be a future PR?

nmanu1:
> @nmanu1 In the screenshot above, I do not see the Http functions as blue hyperlinks. Will that be a future PR?

Yes, that will be part of another [item](https://yexttest.atlassian.net/browse/SLAP-2569)

PR #293, opened 2023-02-28
feat(plugin): generate functionMetadata.json

nmanu1:
This PR adds a step to the vite-plugin's `build` command (i.e. `pages build`) to generate `functionMetadata.json`. This file contains a mapping of the files in the `functions` directory to the names of the functions that are their default exports. If there is no top-level `functions` directory, `functionMetadata.json` is not generated.

J=SLAP-2625
TEST=manual

Point the locations starter to a local pack of PagesJS and see that the file is correctly generated when there is a `functions` directory, and is not generated if there is no `functions` directory.

### DIFF
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -25,7 +25,7 @@ const getFunctionMetadataMap = async (): Promise<
   Record<string, FunctionMetadata>
 > => {
   const filepaths = glob
-    .sync(path.join(FUNCTIONS_PATH, "**/*.{tsx,jsx,js,ts}"), { nodir: true })
+    .sync(path.join(FUNCTIONS_PATH, "**/*.{js,ts}"), { nodir: true })
     .map((f) => path.resolve(f));
   const functionMetadataArray: [string, FunctionMetadata][] = await Promise.all(
     filepaths.map(async (filepath) => {
@@ -53,7 +53,7 @@ export const generateFunctionMetadataFile = async () => {
   const functionMetadataMap = await getFunctionMetadataMap();
   fs.writeFileSync(
     FUNCTION_METADATA_PATH,
-    JSON.stringify(functionMetadataMap, null, "  ")
+    JSON.stringify(functionMetadataMap, null, 2)
   );
 };
 
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -1,0 +1,63 @@
+import fs from "fs";
+import path from "path";
+import esbuild from "esbuild";
+import { importFromString } from "module-from-string";
+import glob from "glob";
+
+const FUNCTIONS_PATH = path.resolve("functions");
+const FUNCTION_METADATA_PATH = path.join(
+  FUNCTIONS_PATH,
+  "functionMetadata.json"
+);
+const TEMP_DIR = ".temp";
+
+/** Metadata for a serverless function. */
+type FunctionMetadata = {
+  /** Name of the function. */
+  entrypoint: string;
+};
+
+/**
+ * Returns a mapping of file path (relative to the repo root) to the metadata
+ * for the function that is the default export of that file.
+ */
+const getFunctionMetadataMap = async (): Promise<
+  Record<string, FunctionMetadata>
+> => {
+  const filepaths = glob
+    .sync(path.join(FUNCTIONS_PATH, "**/*.{tsx,jsx,js,ts}"), { nodir: true })
+    .map((f) => path.resolve(f));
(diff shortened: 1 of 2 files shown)


packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -0,0 +1,63 @@
...
+    functionMetadataArray.filter(([, metadata]) => metadata.entrypoint)
+  );
+};
+
+/** Generates a functionMetadata.json file from the functions directory. */
+export const generateFunctionMetadataFile = async () => {
+  const functionMetadataMap = await getFunctionMetadataMap();
+  fs.writeFileSync(
+    FUNCTION_METADATA_PATH,
+    JSON.stringify(functionMetadataMap, null, "  ")
```
tmeyer2115, line 56: Small nit: Could we use `2` for the `space` argument in `JSON.stringify`? I think that's a bit more readable and it's what we do elsewhere in our codebase.

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Returns a mapping of file path (relative to the repo root) to the metadata
+ * for the function that is the default export of that file.
+ */
+const getFunctionMetadataMap = async (): Promise<
+  Record<string, FunctionMetadata>
+> => {
+  const filepaths = glob
+    .sync(path.join(FUNCTIONS_PATH, "**/*.{tsx,jsx,js,ts}"), { nodir: true })
```
tmeyer2115, line 28: For my edification, do we allow TSX or JSX server-less functions? What's the use-case there?

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -0,0 +1,63 @@
...
+    .map((f) => path.resolve(f));
+  const functionMetadataArray: [string, FunctionMetadata][] = await Promise.all(
+    filepaths.map(async (filepath) => {
+      const buildResult = await esbuild.build({
+        entryPoints: [filepath],
+        outdir: TEMP_DIR,
```
tmeyer2115, line 34: Do we delete `TEMP_DIR` once we're done?

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -0,0 +1,63 @@
...
+        write: false,
+        format: "esm",
+        bundle: true,
+      });
+      const importedFile = await importFromString(
+        buildResult.outputFiles[0].text
+      );
+      const relativePath = path.relative(process.cwd(), filepath);
```
tmeyer2115, line 42: Do you have a sense for how this method of getting the default function name compares to extracting it using an ES6 AST? I'm curious what's faster, running ESBuild & Importing or creating an AST. The AST method has the benefit of not needing a Temp Dir either. Another alternative would be a Regex. That seems like it would be fastest and fairly reliable.

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
nmanu1: The `TEMP_DIR` isn't actually created and nothing is written to it because we pass in `write: false`

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
nmanu1: I originally included TSX and JSX because I wasn't sure of all the envisioned use cases for serverless functions, but Andrew confirmed that only JS or TS would be used. Updated!

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
nmanu1: I didn't want to use a Regex because it would be more brittle. We'd either have to enforce a particular way of defining a function, or check for each syntax (function declaration, arrow function, etc.). Since we may also expand to named function exports in the future, we'd have to keep adding Regexs. Another drawback to using a Regex is that if the default export is a variable declaration, we'd want to check the variable to see if it's a function or not. Using the current method, the [`name`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) field wouldn't exist for a variable that wasn't a function (e.g. a number or string).

I went with the ESBuild & Import route because it's similar to what's already used for loading template modules. But, I haven't done a speed comparison between that and the TS AST. I can look into that more (probably with `ts-morph` since it's friendlier than directly using TS and we have experience using it in Studio).

tmeyer2115: As long as the writing of the file seems generally performant, I'm totally fine with the ESBuild & Import approach. My main concern was it had a side effect of generating a `TEMP_DIR`, but you addressed that above!

tmeyer2115 (approved):

PR #295, opened 2023-03-20
feat(plugin): update `functionMetadata` generation logic

tmeyer2115:
Updated the generation logic for the `functionMetadata.json` file. Now, if there's a file without a `default` export, we log an error in the console that includes the file name. 

J=SLAP-2658
TEST=manual

### DIFF
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -1,21 +1,21 @@
-import fs from "fs";
-import path from "path";
-import esbuild from "esbuild";
-import { importFromString } from "module-from-string";
-import glob from "glob";
+import fs from "fs"
+import path from "path"
+import esbuild from "esbuild"
+import { importFromString } from "module-from-string"
+import glob from "glob"
 
-const FUNCTIONS_PATH = path.resolve("functions");
+const FUNCTIONS_PATH = path.resolve("functions")
 const FUNCTION_METADATA_PATH = path.join(
   FUNCTIONS_PATH,
   "functionMetadata.json"
-);
-const TEMP_DIR = ".temp";
+)
+const TEMP_DIR = ".temp"
 
 /** Metadata for a serverless function. */
 type FunctionMetadata = {
   /** Name of the function. */
-  entrypoint: string;
-};
+  entrypoint: string
+}
 
 /**
  * Returns a mapping of file path (relative to the repo root) to the metadata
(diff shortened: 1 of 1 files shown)


packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -26,38 +26,55 @@ const getFunctionMetadataMap = async (): Promise<
...
-        buildResult.outputFiles[0].text
-      );
-      const relativePath = path.relative(process.cwd(), filepath);
-      return [relativePath, { entrypoint: importedFile.default?.name }];
-    })
-  );
+    .map(f => path.resolve(f))
+  const functionMetadataArray: [string, FunctionMetadata][] = []
+
+  await Promise.allSettled(filepaths.map(generateFunctionMetadata)).then(
```
oshi97, line 35: instead of `.then` I think using await and  then a `.filter` like `const results = await Promise.allSettled()` `const blah = results.filter(d => can console.error here)` is a little cleaner

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -26,38 +26,55 @@ const getFunctionMetadataMap = async (): Promise<
...
+    entryPoints: [filepath],
+    outdir: TEMP_DIR,
+    write: false,
+    format: "esm",
+    bundle: true,
+  })
+  const importedFile = await importFromString(buildResult.outputFiles[0].text)
+
+  if (!importedFile.default) {
+    return Promise.reject("Blah")
```
oshi97: I believe you can `throw` here and it'll be equivalent behavior (and also a little more async-awaitey)

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -1,21 +1,21 @@
...
 
-const FUNCTIONS_PATH = path.resolve("functions");
+const FUNCTIONS_PATH = path.resolve("functions")
 const FUNCTION_METADATA_PATH = path.join(
   FUNCTIONS_PATH,
   "functionMetadata.json"
-);
-const TEMP_DIR = ".temp";
+)
+const TEMP_DIR = ".temp"
```
oshi97, line 13: Little scope creep-ey but can we use os.tmpdir here https://nodejs.org/api/os.html#ostmpdir

oshi97 (approved):

PR #300, opened 2023-04-20
test(dev): playwright tests for static-site

oshi97:
Takes https://github.com/yext/pages/pull/224, simplifies the setup, and removes tests for entity pages.

I cannot run tests in safari locally, since my macos version is too old (and I'm unable to upgrade it due to IT restrictions).
At first, I wanted to only run chrome tests locally and only add safari and firefox in CI. However, in order to check-in
snapshots I have to run the tests locally.
Probably there should be a way to upload snapshots from CI, since we'll want to have snapshots  in additional environments
(windows + linux) that we don't have access to locally. Once that infrastructure is in place, we can remove the acceptance_test.yml we have for ensuring the pages dev server can start in windows and linux.

I don't think the current prepare script in PagesJS is ideal for fresh clone scenarios.
The first pnpm install will run the build for PagesJS, however because the build occurs AFTER the install,
on a fresh clone of the repo, the built assets will not be linked throughout the monorepo, so another pnpm install
is needed, which will also do an extra build of PagesJS. Running the build an extra time isn't ideal.

J=none
TEST=auto


packages/pages/src/dev/dev.ts
```diff
@@ -35,6 +44,12 @@ export const devCommandModule: CommandModule<unknown, DevArgs> = {
         type: "boolean",
         demandOption: false,
         default: true,
+      })
+      .option("open-browser", {
+        describe: "Open the browser",
```
nmanu1, line 49: I had just put up a placeholder description before. Can we change this to something like `Automatically opens the browser on server start-up`?

packages/pages/package.json
```diff
@@ -77,14 +77,14 @@
     "ink": "^3.2.0",
     "lodash": "^4.17.21",
     "mime-types": "^2.1.35",
-    "module-from-string": "^3.2.1",
+    "module-from-string": "^3.3.0",
```
oshi97, line 80: I just updated both of these deps to match the lockfile, for some reason when I ran pnpm install it ended up downgrading these (not exactly sure why, maybe I already had these versions in my node_modules)

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Install again to link deps
+        run: pnpm i
+
+      - name: Install playwright browsers
+        run: pnpm exec playwright install
+
+      - name: Run Playwright tests
+        run: pnpm --filter static-site run playwright
+
+      - name: Upload 'React' Playwright Report
```
nmanu1, line 38: nit: rename to static-site, rather than react

playground/static-site/package.json
```diff
@@ -1,20 +1,32 @@
...
-  "private": true,
-  "version": "1.0.0",
+  "version": "0.0.0",
+  "description": "An example repository to help users get started with React development on Yext Pages",
+  "author": "sumo@yext.com",
+  "license": "BSD-3-Clause",
   "type": "module",
   "scripts": {
-    "dev": "pages dev --local",
+    "dev": "pages dev --local --no-open-browser",
```
nmanu1, line 9: can we add `--no-prod-url` here so we don't get the logs about being unable to generate pages for documents without slugs?

.github/workflows/playwright.yml
oshi97: I just removed the specification

playground/static-site/package.json
oshi97: I can just remove the currently unused localData which would probably be cleaner

nmanu1: I'd rather keep at least one entity file in localData since I was imagining we'd just add an entity page test to static-site (and rename it). That way, we won't need another test repo for an entity page, and only a second one for multibrand

oshi97: I was thinking we can do that when the entity page test is added? since it would be unused in this PR

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,43 @@
+name: Playwright Tests
+
+on: push
```
nmanu1, line 3: do we want to run on pull request too?

playground/static-site/playwright.config.ts
```diff
@@ -0,0 +1,17 @@
...
+import base from "../playwright.config.js";
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+const config: PlaywrightTestConfig = {
+  ...base,
+  /* Run your local dev server before starting the tests */
+  webServer: {
+    command: "npm run dev",
```
nmanu1, line 11: could we add the `--no-open-browser` flag to this command instead of in the package.json script? that way, if you just want to look at the local dev pages, you don't have to manually open the browser

.github/workflows/playwright.yml
oshi97: I saw Duval only had it on push (and even set it to ignore the main branch), and if we run it on push AND pull_request that doubles the number of times the tests run from 3 actions to 6, and I was worried about test flakiness. BUT the tests should always be really simple so I think we can give it a try!

nmanu1 (approved):

PR #302, opened 2023-04-21
test(playground): upload Windows/Linux screenshots via CI

oshi97:
In order to run snapshot testing in Windows and Linux, snapshots for those environments must be committed to the repo.
Since we don't want to manually add snapshots from 3 different operating systems, this PR updates the playwright ci action to do that for us.

The action now only runs on push instead of on pull_request. When the playwright action runs on pull_request, we're not able to push to the original branch. This is because actions that run on pull_request are run on a new branch created automatically by github, this new branch is the merge commit between the head and base branch.
Running on push will run the action directly on the head branch.

J=none
TEST=manual

see that snapshots are uploaded from ubuntu and windows


.github/workflows/playwright.yml
```diff
@@ -1,43 +1,61 @@
 name: Playwright Tests
 
-on: [push, pull_request]
+on: workflow_call
 
 jobs:
-  test:
+  acceptance:
     strategy:
+      fail-fast: false
```
nmanu1, line 8: What does this do? Keep running other tests in the matrix when one fails?

.github/workflows/playwright.yml
```diff
@@ -1,43 +1,61 @@
...
-      - name: Install deps and build
-        run: pnpm i && pnpm run build
-
-      - name: Install again to link deps
-        run: pnpm i
+      # It's important to not run the build for @yext/pages on Windows.
+      # Currently the build does not work (on Windows) and will replace the
+      # downloaded build artifact without any warnings/errors.
+      - run: pnpm i --ignore-scripts
+      - run: ls ./node_modules/@yext/pages/dist/bin
```
nmanu1, line 32: Is this output useful to have?

.github/workflows/playwright.yml
```diff
@@ -1,43 +1,61 @@
...
+      fail-fast: false
       matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
         node-version: [17.x, 18.x, 19.x]
-    timeout-minutes: 60
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        working-directory: ./playground/locations-site
```
nmanu1, line 15: Do we need to set this working directory?

.github/workflows/playwright.yml
oshi97: Yeah, otherwise we have scenarios like the linux ones failing (due to missing snapshots) and preventing the windows actuons from uploading their snapshots

.github/workflows/playwright.yml
oshi97: I copy pasted from the older acceptance test workflow and missed this back then, will remove!

.github/workflows/playwright.yml
oshi97: Ah I can remove, it was only being used by the build-test-site step

.github/workflows/playwright.yml
```diff
@@ -1,45 +1,61 @@
...
         with:
           name: playwright-reports
           path: playground/*/playwright-report/
-          retention-days: 30
+          retention-days: 30   
+
+      - name: Ensure build doesn't error
+        run: |
+          pnpm --filter locations-site run build-test-site
+          pnpm --filter multibrand-site run build-test-site
```
nmanu1, line 61: We might want to add build-test-site scripts to the multibrand repo to build for the scoped brands as well (e.g. `pages build --scope sunglasses.rayban.com`)

.github/workflows/playwright.yml
```diff
@@ -1,45 +1,61 @@
...
+      - run: pnpm i --ignore-scripts
 
       - name: Install playwright browsers
         run: pnpm exec playwright install
-
+  
       - name: Run Playwright tests
         run: |
           pnpm --filter locations-site run playwright
           pnpm --filter multibrand-site run playwright
```
nmanu1, line 36: I forgot to check this in my PR, but if the locations-site tests fail, will it still run the multibrand tests?

.github/workflows/playwright.yml
```diff
@@ -1,45 +1,61 @@
...
+      # This step will commit any new snapshots that weren't taken locally.
+      # i.e. snapshots for Windows and Linux machines
+      - name: Commit New Snapshots
+        uses: EndBug/add-and-commit@v9
+        if: ${{ failure() && github.ref_name != 'main' }}
+        with:
+          message: Commit new snapshots for ${{ matrix.os }}
+          add: playground/*
+          push: true
+          pull: --rebase --autostash
```
nmanu1, line 47: What does this do?

.github/workflows/playwright.yml
nmanu1: Also, you can merge these commands using `pnpm --filter "./playground/*" run build-test-site`. Or even match on regex if we end up having `build-test-site:oakley` or something (`pnpm --filter "./playground/*" run "/^build-test.*/"`).

The reason I didn't merge the commands for running the Playwright tests is because it tries executing them in parallel. So the first site is served and then the second set of tests sees that the port is in use and tries to use that for its tests, resulting in errors.

.github/workflows/playwright.yml
oshi97: the rebase and autostash flags of git pull
theyre needed for multiple commits triggered on the same push

.github/workflows/playwright.yml
oshi97: Ah ok that's cool they run in parallel! I'll use the first I think the regex is overkill if we don't need it

.github/workflows/playwright.yml
oshi97: Ah apparently they don't, though the step will still fail. I added a site matrix which I think simplifies some stuff"

.github/workflows/playwright.yml
```diff
@@ -1,40 +1,50 @@
...
-          pnpm --filter multibrand-site run playwright
+        run: pnpm --filter ${{ matrix.site }} run playwright
+
+      # This step will commit any new snapshots that weren't taken locally.
+      # i.e. snapshots for Windows and Linux machines
+      - name: Commit New Snapshots
+        uses: EndBug/add-and-commit@v9
+        if: ${{ failure() && github.ref_name != 'main' }}
+        with:
+          message: Commit new snapshots for ${{ matrix.os }}
```
nmanu1, line 43: nit: can we update this name to also include which site the commit is for?

.github/workflows/playwright.yml
```diff
@@ -43,3 +53,6 @@ jobs:
           name: playwright-reports
           path: playground/*/playwright-report/
```
nmanu1, line 54: now that we have the site matrix, we should only upload the report for that site

.github/workflows/playwright.yml
oshi97: Ended up adding a matrix for the site so everything can be in parallel

.github/workflows/playwright.yml
nmanu1: also, the names should be different for each run so they don't write over each other

.github/workflows/playwright.yml
nmanu1: Sounds good. Can we still add the scoped build-test-site commands though?

oshi97: It's scoped to matrix.site right now, what do you have in mind?

nmanu1: Right now, `build-test-site` in the multibrand site only does a top-level, unscoped build. We could also test for builds scoped to the individual brands (e.g. `pages build --scope sunglasses.rayban.com`)

.github/workflows/playwright.yml
```diff
@@ -1,45 +1,58 @@
...
+          pull: --rebase --autostash
+          default_author: github_actions
 
       - name: Upload Playwright Reports
         uses: actions/upload-artifact@v3
         if: always()
         with:
-          name: playwright-reports
-          path: playground/*/playwright-report/
+          name: playwright-report-${{ matrix.site }}
```
nmanu1, line 53: Can we also reference the node version in the name? Now the GH workflow has separate artifacts for the locations-site and the multibrand-site, but the node versions are still overwriting each other

.github/workflows/playwright.yml
```diff
@@ -1,40 +1,53 @@
...
-      - name: Run Playwright tests
-        run: |
-          pnpm --filter locations-site run playwright
-          pnpm --filter multibrand-site run playwright
+  
+      - name: Locations Site Playwright
+        run: pnpm --filter locations-site run playwright
+
+      - name: Multibrand Playwright
+        if: failure()
```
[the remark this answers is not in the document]
oshi97: ah oops

.github/workflows/playwright.yml
```diff
@@ -1,40 +1,53 @@
...
+        run: pnpm --filter multibrand-site run playwright
+
+      # This step will commit any new snapshots that weren't taken locally.
+      # i.e. snapshots for Windows and Linux machines
+      - name: Commit New Snapshots
+        uses: EndBug/add-and-commit@v9
+        if: ${{ failure() && github.ref_name != 'main' }}
```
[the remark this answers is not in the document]
oshi97: https://docs.github.com/en/actions/learn-github-actions/expressions#failure

nmanu1 (approved):

PR #304, opened 2023-04-21
test(playground): add multibrand playwright tests

nmanu1:
Add Playwright tests for a multibrand repo.


playground/multibrand-site/__tests__/dev.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const waitForStyling = async () => {
+  await new Promise((r) => setTimeout(r, 500));
+};
+
+test("index page loads", async ({ page }) => {
+  await page.goto("/");
+  await waitForStyling();
+  await expect(page).toHaveTitle(/Pages Development Page/);
+  await expect(page).toHaveScreenshot();
```
oshi97, line 11: I only thought of this know but I think it would be good to check the contents of the index page (at least a little bit). I was thinking we could just add a single check that a link to /sunglasses/227213256(...) exists

oshi97 (approved):

PR #307, opened 2023-05-09
feat: add ability to map different custom fields for each feature on …

Jeffrey-Rhoads17:
…urlWriteback

J=PC-215312
TEST=manual


packages/plugins/urlWriteback/mod.ts
```diff
@@ -25,16 +25,27 @@ interface WritebackPayload {
...
  *
+ * The fieldToFeatureMap is an optional parameter that can be used to define
+ * a custom field to store other features that are powered by entity data
+ * i.e., custom review forms that are powered by entity data but have a different
+ * CF to store the prod URL of the page.
+ * 
  * @public
  */
-export default function urlWriteback(config: UrlWritebackConfig) {
+export default function urlWriteback(config: UrlWritebackConfig, fieldToFeatureMap: Record<string, string> = {}) {
```
bhainesva, line 35: think it would make sense for the map to be an optional property on the config, rather than a new second argument

packages/plugins/urlWriteback/mod.ts
```diff
@@ -25,16 +25,27 @@ interface WritebackPayload {
...
   function onUrlChange(event: WritebackPayload) {
     if (!(event.entityId && event.locale && event.url)) {
       return null;
     }
 
+    let field = config.field;
+
+    if (fieldToFeatureMap && fieldToFeatureMap[event.feature]) {
+      field = fieldToFeatureMap[event.feature];
```
bhainesva, line 44: This might just be me but when I read `fieldToFeatureMap` I assumed the keys would be fields and the values would be features.

Jeffrey-Rhoads17: yeah i realized that too, I'm updating the name to be `featureToFieldMap`

packages/plugins/urlWriteback/mod.ts
Jeffrey-Rhoads17: gotcha! I wasn't sure if I could add an option property on the config, but definitely makes more sense to have it there

mkilpatrick (approved):

PR #309, opened 2023-05-12
feat(dev/plugin)!: allow custom client/server rendering templates

mkilpatrick:
This adds the ability for a user to define their own _client.tsx/_server.tsx so that they have full control over the server and hydration rendering. If none are defined in the user's repo, it will default to built-in templates provided by PagesJS.

This is a **breaking change** for anyone not pinned to a specific previous version. This is because a new folder _render_ is output during the build process that YextCI must be aware of. The Generator plugin section in ci.json should now look like this:
```
{
  "pluginName": "Generator",
  "sourceFiles": [
    {
      "root": "dist/plugin",
      "pattern": "*{.ts,.json}"
    },
    {
      "root": "dist",
      "pattern": "assets/{server,static,renderer,render}/**/*{.js,.css}"
    }
  ],
  "event": "ON_PAGE_GENERATE",
  "functionName": "Generate"
}
```
_Note that a future PR will auto-generate the Generator ci.json section during build, making this a moot issue._

This change also has the added benefit that only a single hydration template (_client.tsx) is necessary. Previously, every template built required its own hydration template. This should cut down on the number of files necessary by the Yext Plugin system as it is now rendered at runtime.


packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
```diff
@@ -32,83 +35,31 @@ export const reactWrapper = <T extends TemplateRenderProps>(
...
 
-const getCssTags = (
-  filepath: string,
-  manifest: bundlerManifest,
-  seen: Set<string>
-): Set<string> => {
-  const entry = structuredClone(
-    Object.entries(manifest).find(([file]) => file === filepath)
+  const clientHtml = getHydrationTemplate(
+    path.join("../" + clientRenderPath),
```
mkilpatrick, line 49: Use comma and not +

packages/pages/src/vite-plugin/build/closeBundle/manifest.ts
```diff
@@ -27,12 +28,28 @@ export const generateManifestFile = (
     ]
   );
 
+  // Add the renderPaths to the manifest. This defines the _server entry.
```
mkilpatrick, line 31: _server and _client

packages/pages/src/common/src/template/hydration.ts
```diff
@@ -0,0 +1,188 @@
...
+  (info.imports || [])
+    .flatMap((f) => Array.from(getCssTags(f, manifest, seen)))
+    .forEach((f) => cssFiles.add(f));
+
+  return cssFiles;
+};
+
+const validateHeadTagExists = (serverHtml: string): void => {
+  if (serverHtml.indexOf("</head>") === -1) {
+    throw new Error("No head tag is defined _server.tsx");
```
mancusi, line 173: "_server.tsx: No head tag is defined" ?

packages/pages/src/bundler.js
```diff
@@ -24,7 +24,7 @@ const testFilter = (f) =>
 const filters = new Set(["./src/bundler.js"]);
 
 const files = glob
-  .sync("./src/**/*\\.*")
```
mancusi, line 27: not sure I understand this. what's the reason for this change?

packages/pages/src/common/src/template/hydration.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+};
+
+/**
+ * Finds the closing </head> tag and injects the input string into it.
+ * @param html
+ */
+const injectIntoHead = (html: string, stringToInject: string): string => {
+  const closingHeadIndex = html.indexOf("</head>");
```
mancusi, line 182: throw an error if this cannot be found

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -38,3 +49,89 @@ export const getTemplateFilepathsFromProjectStructure = (
       : [projectStructure.templatesRoot]
   );
 };
+
+const globalClientRenderFilename = "_client.tsx";
+const globalServerRenderFilename = "_server.tsx";
+
+/**
+ * Determines the client and server rendering templates to use. It first looks for a _client/server.tsx file in the scoped
```
mancusi, line 57: what is a "scoped path"?

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -38,3 +49,89 @@ export const getTemplateFilepathsFromProjectStructure = (
...
+  // files.
+  if (curDirectory === "JEST") {
+    curDirectory = path.join(
+      "file://",
+      process.cwd(),
+      "src/common/src/template/internal"
+    );
+  }
+
+  // console.log("import.meta.url", import.meta.url);
```
mancusi, line 130: remove

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -38,3 +49,89 @@ export const getTemplateFilepathsFromProjectStructure = (
...
+ * @param templatesRootPath the path where the templates live, typically src/templates
+ * @param scopedTemplatePath the subfolder path inside templatesRoot to scope to - used in multibrand setups
+ * @param globalFilename the file to find
+ * @returns the path to the appropriate file along with a boolean denoting if the path returned is a custom render template
+ */
+const findGlobalRenderFile = (
+  templatesRootPath: Path,
+  scopedTemplatePath: Path | undefined,
+  globalFilename: string
+): [string, boolean] => {
```
mancusi, line 98: it'd be more clear if you read the file right away and passed around a file object rather than the absolute path to the file

packages/pages/src/dev/server/middleware/sendAppHTML.ts
```diff
@@ -20,7 +23,8 @@ export default async function sendAppHTML(
   templateModuleInternal: TemplateModuleInternal<any, any>,
   props: TemplateRenderProps,
   vite: ViteDevServer,
-  pathname: string
+  pathname: string,
+  projectStructure: ProjectStructure
 ) {
   if (templateModuleInternal.render) {
```
mancusi, line 29: so client/server templates are only supported with a default export? I think that makes sense but just want to double check.

packages/pages/src/common/src/template/hydration.ts
```diff
@@ -0,0 +1,188 @@
...
+  headConfig?: HeadConfig
+): string => {
+  validateHeadTagExists(serverHtml);
+
+  // Add the language to the <html> tag if it exists
+  serverHtml.replace("<!--app-lang-->", appLanguage);
+
+  serverHtml = injectIntoHead(
+    serverHtml,
+    `<script type="module">${clientHtml}</script>`
```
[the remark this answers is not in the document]
mancusi: also why would the client and server html intermingle? They should be separate right?

packages/pages/src/common/src/template/types.ts
```diff
@@ -195,3 +197,18 @@ export interface TemplateRenderProps extends TemplateProps {
    */
   relativePrefixToRoot: string;
 }
+
+export interface ClientServerRenderTemplates {
```
mancusi, line 201: can you please document these interfaces?

packages/pages/src/dev/server/middleware/sendAppHTML.ts
```diff
@@ -30,28 +34,40 @@ export default async function sendAppHTML(
...
-    `<!--app-head-->`,
-    `<head>
-        <script type="text/javascript">
-          window._RSS_PROPS_ = ${JSON.stringify(props)};
-          window._RSS_TEMPLATE_PATH_ = '${templateModuleInternal.path}';
-          window._RSS_LANG_ = '${getLang(headConfig, props)}';
-        </script>
-        ${headConfig ? renderHeadConfigToString(headConfig) : ""}
-      </head>`
+  const clientHtml = getHydrationTemplate(
```
mancusi, line 55: I'm still unclear here why we need to get the client and server side templates here. shouldn't we just have to get the client html?

packages/pages/src/dev/server/middleware/serverRenderSlugRoute.ts
```diff
@@ -3,7 +3,10 @@ import { ViteDevServer } from "vite";
 import { propsLoader } from "../ssr/propsLoader.js";
```
mancusi, line 3: unfortunate we need to duplicate the work here and in serverRenderRoute

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -56,36 +62,66 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
...
       .map((file) => file.name)
+      .filter((f) => f !== "_client.tsx" && f !== "_server.tsx")
       .forEach((template) => {
         const parsedPath = parse(template);
         const outputPath = `server/${parsedPath.name}`;
         if (entryPoints[outputPath]) {
           return;
         }
         entryPoints[outputPath] = path.join(dir, template);
-        if (parsedPath.ext === ".tsx" || parsedPath.ext === ".jsx") {
```
mancusi, line 79: YES. So glad you were able to remove this.

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
```diff
@@ -1,26 +1,29 @@
...
+  getServerTemplatePlugin,
+} from "../../../../common/src/template/hydration.js";
+import path from "path";
 
-export const reactWrapper = <T extends TemplateRenderProps>(
+export const reactWrapper = async <T extends TemplateRenderProps>(
   props: T,
   templateModuleInternal: TemplateModuleInternal<any, any>,
-  template: string,
   hydrate: boolean,
```
mancusi, line 16: unused

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
```diff
@@ -1,26 +1,29 @@
...
-  getHeadConfig?: GetHeadConfig<any>
-): string => {
+  clientRenderPath: string,
```
mancusi, line 17: the server module should be imported ASAP and cached similar to what we do with the TemplateModules. The main reason for doing so is that dynamically importing without caching causes a memory leak in the plugin. A secondary reason is that it'd be clearer.

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
```diff
@@ -32,83 +35,31 @@ export const reactWrapper = <T extends TemplateRenderProps>(
...
+  });
 
-const getCssTags = (
-  filepath: string,
-  manifest: bundlerManifest,
-  seen: Set<string>
-): Set<string> => {
-  const entry = structuredClone(
-    Object.entries(manifest).find(([file]) => file === filepath)
+  const clientHtml = getHydrationTemplate(
```
mancusi, line 48: might be nice to do this up-front  and just pass the hydration template into this function? not really a big deal either way..

packages/pages/src/vite-plugin/build/closeBundle/moduleLoader.ts
```diff
@@ -25,7 +25,12 @@ export const loadTemplateModules = async (
     }
 
     const templateModuleInternal =
-      convertTemplateModuleToTemplateModuleInternal(p, templateModule, true);
+      convertTemplateModuleToTemplateModuleInternal(
+        p,
+        templateModule,
+        true,
+        false // doesn't matter here, it's already been validated at this point
```
mancusi, line 32: not a huge fan of this assumption here or elsewhere

mancusi (changes requested):

packages/pages/src/common/src/project/structure.ts
```diff
@@ -152,6 +156,11 @@ export class ProjectStructure {
         "/" +
         this.#config.filepathsConfig.serverBundleOutputRoot
     );
+    this.renderBundleOutputRoot = new Path(
+      this.#config.filepathsConfig.distRoot +
+        "/" +
```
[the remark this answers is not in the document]
mkilpatrick: It doesn't matter because usages of it handle joining, but I will do it to be safe.

packages/pages/src/common/src/template/hydration.ts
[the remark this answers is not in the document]
mancusi: Gotcha, that makes sense. Before I think it was just a separately-bundled file but now we're including it's dependencies in the server bundle. Any concerns about that increasing the server-size?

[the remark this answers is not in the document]
mkilpatrick: I'm not 100% clear what you mean. We're effectively doing the same thing as before. Prior to this change the wrapper.ts used in the plugin did this:

```
return `<!DOCTYPE html>
    <html lang=${lang}>
    <head>
        <script>window.__INITIAL__DATA__ = ${JSON.stringify(props)}</script>
        // css stuff
        // head config stuff
    </head>
    <body>
        <div id="reactele">${template}</div>${
    hydrate
      ? `<script type="module" src="${
          props.relativePrefixToRoot +
          findHydrationFilename(
            `${projectFilepaths.hydrationBundleOutputRoot}/${templateModuleInternal.templateName}.tsx`,
            props
          )
        }" defer></script>`
      : ""
  }
    </body>
    </html>`;
};
```

The only difference now is that we use the _server.tsx and _client.tsx from the dist/assets folder. The templates in dist/assets/server are the same where they are just the bundled template part. The things in hydration.ts just put it all together. The effect is the same:
```
// read _server.tsx with dist/assets/server stringified code in it (same as above <div id="reactele">${template}</div>)
// inject below into its <script type="module>HERE</script> for hydration (also same as above hydrate? part)

import {default as Component} from "pathToTemplateModule...aka the file in dist/assets/server";
import {render} from "_client.tsx";
render(
{
    Page: Component, // this is the dist/assets/server template again
    pageProps: ${JSON.stringify(props)},
}
);
```
The total size of what's returned here should be the same. The added benefit actually is that there are no longer compiled hydration templates required in the plugin AND we don't need to add a props json on the page to read from.

I guess maybe what you're saying is that in the script module we're dumping the string rather than having it be a referenced file that's separate?

Also did you want me to do this?
> I can change it from clientHtml to clientRenderString if you think that's better.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
mkilpatrick: Scoped paths are for multi-brand sites that use subfolders. Slapshot added this.

packages/pages/src/dev/server/middleware/sendAppHTML.ts
mkilpatrick: No, it's a named export of `render` in client/server, not default. The line you commented on is the older custom render func in the **template** if you want to go full custom.

packages/pages/src/dev/server/middleware/sendAppHTML.ts
mkilpatrick: We need the server template too because it defines how to render things:

```
const { Page, pageProps } = pageContext;
  
const viewHtml = ReactDOMServer.renderToString(
    <Page {...pageProps} /></CacheProvider>
);

return `<!DOCTYPE html>
    <html lang="${language}">
      <head>whatever here</head>
      <body>
        <div>Duval is the GOAT!!!</div>
        <div id="test">${viewHtml}</div>
      </body>
    </html>`;
```

We can't hardcode this because then:
1. a user isn't in full control
2. the target id of the div needs to match between the client and server
3. the original impetus beside full customizability was for Heartland to be able to use an Emotion cache, which requires wrapping things on both sides like

```
_server.tsx
const viewHtml = ReactDOMServer.renderToString(
    <CacheProvider value={cache}>
      <Page {...pageProps} />
    </CacheProvider>
  );

_client.tsx
ReactDOM.hydrate(
    <CacheProvider value={cache}>
      <Page {...pageProps} />
    </CacheProvider>,
    document.getElementById("test")
  );
```

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
mkilpatrick: Good call. I left it there for when it's configurable. I've updated to actually use the boolean now, but it's still hardcoded to true further up. Once we make it configurable though it'll just work.

packages/pages/src/vite-plugin/build/closeBundle/moduleLoader.ts
mkilpatrick: Same. I didn't want to have to do this everywhere though:
```
const clientServerRenderTemplates = getGlobalClientServerRenderTemplates(
    projectStructure.templatesRoot,
    projectStructure.scopedTemplatesPath
  );
```

The reason this flag is necessary is because loadTemplateModules and convertTemplateModuleToTemplateModuleInternal happen in a variety of places and validation happens during load. This is beneficial because it checks that you don't have `getHeadConfig` defined if you're using custom rendering templates (because those should set the `<head>`, `<title>`, etc tags instead.

I could move the validation out of here and do it somewhere else like during build, but it's nice for the user to get feedback immediately during local dev. Any ideas or thoughts on if we should move it? I could just do the actual determination of the custom rendering templates per above in all of these assumption places, but it will mean also updating a bunch of params that need to be passed through.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
mkilpatrick: I need the path vs the contents at times. For instance when we get the hydration template:
```
import {render} from "${convertToPosixPath(clientRenderTemplatePath)}";
```

and in the dev server:
```
const serverRenderTemplateModule = (await vite.ssrLoadModule(
    clientServerRenderTemplates.serverRenderTemplatePath
  )) as RenderTemplate;
```

Are you saying to load the file as a module directly here, but return some type that has other things, like the path and isCustom?
```
interface RenderTemplateInternal {
    module: RenderTemplate;
    path: string;
    isCustom: boolean;
}
```

packages/pages/src/dev/server/middleware/serverRenderSlugRoute.ts
mkilpatrick: Good point. The separate files were introduced by Slapshot. I'll make a point to refactor this in a followup PR, but I won't change it in this one.

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
mkilpatrick: You mean load it in renderer.ts instead? Good point. We don't need to keep loading these over and over. I will update. Before I do that I want to confirm some of the other questions first, like having a RenderTemplateInternal type.

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
mkilpatrick: I think this is effectively the same comment as the previous one to load the render templates in renderer.ts and pass them through. Yea?

packages/pages/src/vite-plugin/build/closeBundle/moduleLoader.ts
mancusi: is it possible to pass templateModuleInternal around rather than converting the template module? 

We're doing this on build start and then again when closeBundle is called. I wonder if we could persist that state. That gets around the problem a bit

packages/pages/src/vite-plugin/build/buildStart/rendering/wrapper.ts
mancusi: yeah.

packages/pages/src/dev/server/middleware/sendAppHTML.ts
mancusi: Okay that makes sense. thanks

packages/pages/src/common/src/template/hydration.ts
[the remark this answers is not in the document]
mancusi: > I guess maybe what you're saying is that in the script module we're dumping the string rather than having it be a referenced file that's separate?

Yeah I was thinking it might be better to not inline the JS there. It could cause issues on some sites which want to restrict it.

> Also did you want me to do this?

Yeah that makes it more clear to me.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
mancusi: ah right. thanks

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
mancusi: Yeah something along those lines.

packages/pages/src/vite-plugin/build/closeBundle/moduleLoader.ts
mkilpatrick: We're only loading templates in closeBundle. I updated one place where I had "doesn't matter" and then also removed moduleLoader altogether because it was actually unused (and added back the one validation it had directly into closeBundle).

mancusi (approved):

PR #311, opened 2023-05-18
feat(dev/plugin): add Generator plugin to ci.json during build

mkilpatrick:
There is a new pages generate ci command that will auto update the ci.json's Generator plugin. It also applies during the build step. This is because PagesJS understands how it builds its output to work on the Yext platform and shouldn't require a user to understand the details. Now they don't need to.


packages/pages/src/generate/ci/ci.ts
```diff
@@ -0,0 +1,94 @@
...
+      root: "dist/plugin",
+      pattern: "*{.ts,.json}",
+    },
+    {
+      root: "dist",
+      pattern: "assets/{server,static,renderer,render}/**/*{.js,.css}",
+    },
+  ],
+  event: "ON_PAGE_GENERATE",
+  functionName: "Generate",
```
mancusi, line 93: Maybe we change the name to be a bit more clear what this is in the platform?
pagesjs-generator?

packages/pages/src/generate/ci/ci.ts
```diff
@@ -0,0 +1,94 @@
...
+    defaultProjectStructureConfig.filenamesConfig.ciConfig;
+  const sitesConfigRoot =
+    defaultProjectStructureConfig.filepathsConfig.sitesConfigRoot;
+  const ciConfigAbsolutePath = new Path(
+    path.join(process.cwd(), sitesConfigRoot, ciConfigFilename)
+  );
+  updateCiConfig(ciConfigAbsolutePath.getAbsolutePath(), true);
+};
+
+export const ciCommandModule: CommandModule<unknown, unknown> = {
```
mancusi, line 20: add documentation to all exported functions

packages/pages/src/generate/ci/ci.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  const ciJson = mergeCiConfig(originalCiConfigJson);
+  if (ciJson) {
+    fs.writeFileSync(ciConfigPath, JSON.stringify(ciJson, null, "  "));
+  }
+};
+
+export const mergeCiConfig = (ciConfig: CiConfig): CiConfig | undefined => {
```
mancusi, line 54: rather than mutate the existing config -- thoughts on just copying it and returning a new one?

packages/pages/src/generate/ci/ci.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (!ciConfig.artifactStructure.plugins) {
+    ciConfig.artifactStructure.plugins = [];
+  }
+
+  let hasGeneratorPlugin = false;
+
+  // replace the "Generator" plugin if it was already defined
+  ciConfig.artifactStructure.plugins = ciConfig.artifactStructure.plugins.map(
+    (plugin) => {
+      if (plugin.pluginName === generatorPluginName) {
```
nbramblett, line 64: what if they name their plugin differently? Would it be better to check for a plugin with the ON_PAGE_GENERATE event?

mancusi (commented):
a few comments

mancusi (approved):

PR #318, opened 2023-06-07
feat(plugin): add reverse proxy support

mkilpatrick:
Use import paths relative to the domain rather than the filesystem. This allows reverse proxies to work, including those using a subfolder.

### DIFF
--- a/packages/pages/src/vite-plugin/build/build.ts
+++ b/packages/pages/src/vite-plugin/build/build.ts
@@ -46,6 +46,19 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
           },
         },
         define: processEnvVariables(projectStructure.envVarPrefix),
+        experimental: {
+          renderBuiltUrl(filename, { hostType }) {
+            // Assets are returned with a leading slash for some reason. This adjusts the
+            // paths to be relative for reverse proxy support.
+            if (hostType === "js") {
+              if (filename.at(0) === "/") {
+                return filename.substring(1);
+              }
+              return filename;
+            }
+            return { relative: true };
+          },
+        },
       };
     },
     buildStart: buildStart(projectStructure),
--- a/packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
+++ b/packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
@@ -59,7 +59,6 @@ export interface PluginRenderTemplates {
 export const getPluginRenderTemplates = async (
   manifest: Manifest
 ): Promise<PluginRenderTemplates> => {
-  const clientRenderPath = pathLib.join("..", manifest.renderPaths._client);
   const serverRenderPath = manifest.renderPaths._server.replace("assets", "..");
 
   const serverRenderTemplateModule = (await import(
@@ -68,7 +67,7 @@ export const getPluginRenderTemplates = async (
 
   return {
     server: serverRenderTemplateModule,
-    client: clientRenderPath,
+    client: manifest.renderPaths._client,
   };
 };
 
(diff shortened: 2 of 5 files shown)


colton-demetriou (approved):

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -46,6 +46,19 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
           },
         },
         define: processEnvVariables(projectStructure.envVarPrefix),
+        experimental: {
+          renderBuiltUrl(filename, { hostType }) {
+            // Assets are returned with a leading slash for some reason. This adjusts the
```
mancusi, line 51: https://vitejs.dev/config/shared-options.html#base

They're returned with a leading slash because the default base is `/`

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -46,6 +46,19 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
           },
         },
         define: processEnvVariables(projectStructure.envVarPrefix),
+        experimental: {
+          renderBuiltUrl(filename, { hostType }) {
```
mancusi, line 50: https://vitejs.dev/config/shared-options.html#base

They're returned with a leading slash because the default base is /

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -46,6 +46,19 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
...
+            // Assets are returned with a leading slash for some reason. This adjusts the
+            // paths to be relative for reverse proxy support.
+            if (hostType === "js") {
```
mancusi, line 53: why does this only need to happen for JS assets? do you think you could add a comment about the reasoning?

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -46,6 +46,19 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
           },
         },
         define: processEnvVariables(projectStructure.envVarPrefix),
+        experimental: {
```
mancusi, line 49: Vite recommends pinning to a minor version when using experimental builds https://vitejs.dev/guide/build.html#advanced-base-options

packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
```diff
@@ -59,7 +59,6 @@ export interface PluginRenderTemplates {
 export const getPluginRenderTemplates = async (
   manifest: Manifest
 ): Promise<PluginRenderTemplates> => {
-  const clientRenderPath = pathLib.join("..", manifest.renderPaths._client);
   const serverRenderPath = manifest.renderPaths._server.replace("assets", "..");
```
mancusi, line 62: hm, what is the reasoning something needs to be done in the client path?

mancusi: Is that because in the server case it is relative to the filesystem?

packages/pages/src/vite-plugin/build/build.ts
mkilpatrick: Talked offline, but this won't change until the next major - https://github.com/vitejs/vite/pull/8762#issuecomment-1527147910

packages/pages/src/vite-plugin/build/build.ts
mkilpatrick: This is for js files that have assets in them, not js files only.

packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
mkilpatrick: yes

PR #319, opened 2023-06-12
feat(dev/plugin)!: update head injection logic

mkilpatrick:
Updates the `getHeadConfig` logic to add it to the start of `<head>` instead of the end. Also removes validation that previously did not let you have both a `getHeadConfig` and a custom render template.


mancusi:
Just noting that this is a breaking change and you probably want to note it as such so it can be part of the 1.0 release..

PR #327, opened 2023-06-20
feat(dev/plugin)!: added hydration option to TemplateConfig

shrishayap:
Right now, hydration is set to true by default. This PR adds a hydration field to `TemplateConfig` field (to both internal and externals) and makes it treated as false by default. It is an optional argument, so undefined is treated as false (default). 

This was tested by setting hydrate to `true` and `false` (both dev and with `build:serve` prod) and a react component that counted how many clicks was placed. When set to true, the counter worked and when set to false, the counter did not (as expected).


packages/pages/src/common/src/template/hydration.ts
```diff
@@ -9,6 +9,9 @@ import { TemplateRenderProps } from "./types.js";
  * Dev has a separate function than {@link getHydrationTemplate} due to how Vite messes
  * with the import.meta.url.
  *
+ * Result is split into two strings becuase if user when hydration is disabled,
```
mkilpatrick, line 12: This method isn't necessary to call at all if we're not hydrating.

packages/pages/src/common/src/template/internal/types.ts
```diff
@@ -52,6 +52,8 @@ export interface TemplateModuleInternal<
 export interface TemplateConfigInternal {
   /** The name of the template feature. If not defined uses the template filename (without extension) */
   name: string;
+  /** Determines if hydration is allowed or not for webpages. Defaults to false if not set*/
+  hydration?: boolean;
```
mkilpatrick, line 56: hydrate

mkilpatrick: Also this shouldn't be optional in the internal type. The point of internal is to get the inferred values from the external type so that they are available when we use it in code.

packages/pages/src/common/src/template/internal/types.ts
```diff
@@ -52,6 +52,8 @@ export interface TemplateModuleInternal<
 export interface TemplateConfigInternal {
   /** The name of the template feature. If not defined uses the template filename (without extension) */
   name: string;
+  /** Determines if hydration is allowed or not for webpages. Defaults to false if not set*/
```
mkilpatrick, line 55: Remove the Defaults... part

packages/pages/src/dev/server/middleware/sendAppHTML.ts
```diff
@@ -52,10 +52,16 @@ export default async function sendAppHTML(
...
+    templateModuleInternal.config.hydration === undefined
+      ? false
+      : templateModuleInternal.config.hydration;
+
   const clientHydrationString = getHydrationTemplateDev(
     clientServerRenderTemplates.clientRenderTemplatePath,
     templateModuleInternal.path,
-    props
+    props,
+    hydrate
```
mkilpatrick, line 64: Once you update the internal type you can just do templateModuleInternal.config.hydration

packages/pages/src/dev/server/middleware/sendAppHTML.ts
```diff
@@ -52,10 +52,16 @@ export default async function sendAppHTML(
     ? templateModuleInternal.getHeadConfig(props)
     : undefined;
 
+  const hydrate =
+    templateModuleInternal.config.hydration === undefined
```
mkilpatrick, line 56: This logic should live in the conversion of the templateModule to templateModuleInternal

packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
```diff
@@ -161,11 +161,16 @@ const renderHtml = async (
     return render(props);
   }
 
+  // since hydrate is optional, if not set by user, default to false
+  const hydrate =
```
mkilpatrick, line 165: Same comment

packages/pages/src/common/src/template/hydration.ts
shrishayap: When I tested in dev, without the 

```
import {default as Component} from "${convertToPosixPath(
    templateModulePath
  )}";
```

It does not apply the CSS. I think the importing might be needed by vite to apply the CSS when in dev. I can guard it in vite-pack (prod) but not in dev. Also something to note is that before I made any changes, the vite-pack already had the `if (hydrate)` guard (not functional, but there with a `TODO` comment) but there weren't any for dev.

packages/pages/src/common/src/template/internal/types.test.ts
```diff
@@ -114,4 +120,93 @@ describe("internal/types - convertTemplateModuleToTemplateModuleInternal", () =>
...
+      filename: "myTemplateName.tsx",
+      templateName: "myTemplateName",
+    };
+
+    expect(JSON.stringify(templateConfigInternal)).toEqual(
+      JSON.stringify(expected)
+    );
+  });
+
+  it("convert's hydrate to true whenever hydrate is set to true", async () => {
```
mkilpatrick, line 168: "converts" shouldn't be possessive

packages/pages/src/common/src/template/hydration.ts
mkilpatrick: Ah yea, this is how Vite needs to load the entrypoint. Ok you can ignore my previous comment, but the spelling and incomplete sentence need work.

mkilpatrick (approved):

PR #328, opened 2023-06-21
feat(dev)!: serverless functions on local dev server

benlife5:
Moves serverless functions/complex plugins to `src/functions`. 
Adds support for running serverless functions in the local dev server.

**HTTP / API**
HTTP/API functions should be located in `src/functions/http`. Nested subdirectories are allowed. Each file must have a default export that accepts type `HttpFunctionArgument` and returns type `HttpFunctionResponse`.

These functions are called when a specific URL route is hit. This route is defined by the directory structure. `src/functions/http/api/users/get` will be hosted at `localhost:8000/api/users/get`. 

Query and path params are supported. To use path params, encode the parameter in square bracket. For examples `src/functions/http/api/user/[userId]` would match `localhost:8000/api/user/9373` and call the function with `userId: 9373`.

**onUrlChange**
onUrlChange functions should be located in `src/functions/onUrlChange`. Nested subdirectories are _not_ allowed. Each file must have a default export that accepts type `onUrlChangeFunctionArgument` and returns void.

These functions will be called when the path of a production page changes.

- If there's nothing in `src/functions/onUrlChange` there are no onUrlChange plugins
- If there's one file in `src/functions/onUrlChange`, every feature automatically registers that plugin (this is not pagesJS behavior, this is Yext platform behavior)
- If there's >1 file in `src/functions/onUrlChange`, no feature automatically has an onUrlChangePlugin, an onUrlChangeFunctionName must be specified in a feature's TemplateConfig to register a plugin to that feature. 
    -Ex. `onUrlChangeFunctionName: "onChange2"`  if `src/functions/onUrlChange/onChange2.ts` is the desired plugin

**Dev Server**
The dev server started by `pages dev` can now run HTTP/API type serverless functions. Additionally, all plugins are listed on the development index page.


packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * The exported `config` function's definition.
+ * @public
+ */
+export interface ServerlessFunctionConfig {
+  /** The name of the serverless function feature. */
+  name?: string;
+  /** The http event. */
+  event?: PluginEvent;
```
mkilpatrick, line 35: We don't need this either after all. It will always default to API for serverless. Keep in mind we still want this in the internal type though.

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,131 @@
...
+ * Defines the functions and fields that are available to the serverless function.
+ * @public
+ */
+export interface ServerlessFunctionModule<
+  U extends ServerlessFunctionArgument
+> {
+  /** The exported config function */
+  config?: ServerlessFunctionConfig;
+  /** The exported getPath function */
+  getPath?: GetPath<void>;
```
mkilpatrick, line 14: This doesn't need to be typed.

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,131 @@
...
+  /** HTTP response headers (refer to MDN Web Docs).  */
+  headers: object;
+}
+
+/**
+ * The argument passed to a serverless function
+ * @public
+ */
+export interface ServerlessFunctionArgument {
+  /** Object containing each query parameter as  */
```
mkilpatrick, line 60: was there more to add after the "as?"

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,131 @@
...
+  queryParams: { [key: string]: string };
+
+  /** Internal ID of the site branch. */
```
mkilpatrick, line 63: Is this comment correct?

packages/pages/src/common/src/function/internal/validateServerlessFunctionModuleInternal.ts
```diff
@@ -0,0 +1,81 @@
...
+  ServerlessFunctionModule,
+  ServerlessFunctionConfig,
+} from "../types.js";
+
+export const validateServerlessFunctionDefaultExport = (
+  filepath: string,
+  serverlessFunction: ServerlessFunctionModule<any>
+) => {
+  if (!serverlessFunction.default) {
+    throw new Error(`${filepath} is missing an default export function.`);
```
mkilpatrick, line 15: ${filepath} is missing a default export.

packages/pages/src/common/src/function/internal/validateServerlessFunctionModuleInternal.ts
```diff
@@ -0,0 +1,81 @@
...
+  serverlessFunction: ServerlessFunctionModuleInternal<any>
+) => {
+  validateConfigInternal(
+    serverlessFunction.filename,
+    serverlessFunction.config
+  );
+
+  if (!serverlessFunction.getPath) {
+    throw new Error(
+      `Function ${serverlessFunction.filename} is missing an exported getPath function.`
```
mkilpatrick, line 29: I would use the filepath here too to be consistent with the error above

packages/pages/src/common/src/function/internal/validateServerlessFunctionModuleInternal.ts
```diff
@@ -0,0 +1,81 @@
...
+  filename: string,
+  serverlessFunctionConfigInternal: ServerlessFunctionConfigInternal
+) => {
+  if (!serverlessFunctionConfigInternal.name) {
+    throw new Error(
+      `Function ${filename} is missing a "name" in the config function.`
+    );
+  }
+
+  if (!serverlessFunctionConfigInternal.functionName) {
```
mkilpatrick, line 44: You can remove this. See my other comment.

packages/pages/src/common/src/function/internal/validateServerlessFunctionModuleInternal.ts
```diff
@@ -0,0 +1,81 @@
...
+export const validateServerlessFunctionDefaultExport = (
+  filepath: string,
+  serverlessFunction: ServerlessFunctionModule<any>
+) => {
+  if (!serverlessFunction.default) {
+    throw new Error(`${filepath} is missing an default export function.`);
+  }
+};
+
+export const validateServerlessFunctionModuleInternal = (
```
mkilpatrick, line 19: We don't need to validate the internal domain object, only the external one. If the external passes validation then the internal one we construct should be fine.

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,128 @@
...
+      );
+      const defaultPath = serverlessFunctionFilepath
+        .split("/functions/http/")[1]
+        .split(".")
+        .slice(-2)[0];
+      serverlessFunctionInternal = {
+        ...serverlessFunctionModule,
+        config: {
+          name: defaultPath,
+          functionName: serverlessFunctionPath.name,
```
mkilpatrick, line 82: This should be the name of the default export

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * The exported `config` function's definition.
+ */
+export interface ServerlessFunctionConfigInternal {
+  /** The name of the serverless function feature. */
+  name: string;
+  /** The http event. */
+  event?: PluginEvent;
```
mkilpatrick, line 48: Not optional

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,128 @@
...
+        serverlessFunctionModule
+      );
+      const defaultPath = serverlessFunctionFilepath
+        .split("/functions/http/")[1]
+        .split(".")
+        .slice(-2)[0];
+      serverlessFunctionInternal = {
+        ...serverlessFunctionModule,
+        config: {
+          name: defaultPath,
```
mkilpatrick, line 81: event should default to api here (or w/e the correct string is)

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,128 @@
...
+    return serverlessFunctionInternal;
+  };
+
+const convertServerlessFunctionConfigToServerlessFunctionConfigInternal = (
+  serverlessFunctionName: string,
+  serverlessFunctionConfig: ServerlessFunctionConfig | undefined
+): ServerlessFunctionConfigInternal => {
+  return {
+    name: serverlessFunctionConfig?.name ?? serverlessFunctionName,
+    functionName:
```
mkilpatrick, line 124: Add event

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,128 @@
...
+      serverlessFunctionConfig?.functionName ?? serverlessFunctionName,
```
mkilpatrick, line 125: Infer from default function name

packages/pages/src/common/src/function/types.ts
benlife5: fixed in the code but needs to be fixed on Hitchhikers too

packages/pages/src/common/src/function/types.ts
benlife5: fixed in the code but needs to be fixed on Hitchhikers too

packages/pages/src/common/src/template/types.ts
```diff
@@ -52,7 +52,7 @@ export type TransformProps<T extends TemplateProps> = (props: T) => Promise<T>;
  *
  * @public
  */
-export type GetPath<T extends TemplateProps> = (props: T) => string;
+export type GetPath<T extends TemplateProps | void> = (props: T) => string;
```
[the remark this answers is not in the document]
benlife5: was used when I had used the other GetPath type, now that it is just a simple () => string, I've removed

packages/pages/src/common/src/function/internal/types.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    const functionModuleInternal =
+      convertFunctionModuleToFunctionModuleInternal(
+        "src/functions/http/api/example.ts",
+        functionModule,
+        false
+      );
+    const expected = {
+      default: "exampleFunction",
+      config: {
+        name: "api/example",
```
mkilpatrick, line 32: This is the path, not the name.

packages/pages/src/common/src/function/internal/types.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      default: exampleFunction,
+    };
+    const functionModuleInternal =
+      convertFunctionModuleToFunctionModuleInternal(
+        "src/functions/http/api/example.ts",
+        functionModule,
+        false
+      );
+    const expected = {
+      default: "exampleFunction",
```
mkilpatrick, line 30: This shouldn't be a string.

packages/pages/src/common/src/function/internal/types.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      config: {
+        name: "api/example",
+        functionName: "exampleFunction",
+        event: "API",
+      },
+      path: "src/functions/http/api/example.ts",
+      filename: "example.ts",
+      slug: "api/example",
+      getPath: "getPath",
```
mkilpatrick, line 39: This also shouldn't be a string. getPath is a function

packages/pages/src/common/src/function/internal/types.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    };
+    expect(
+      JSON.stringify(functionModuleInternal, stringifyIncludeFunctionNames)
+    ).toEqual(JSON.stringify(expected));
+  });
+
+  it("throws an error when a function outside /functions/http is missing a config", async () => {
```
mkilpatrick, line 46: Remove / from before functions. Having it indicates root. Same in other places.

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -1,128 +1,88 @@
...
-          ),
-        path: serverlessFunctionFilepath,
-        filename: serverlessFunctionPath.base,
-        functionName: serverlessFunctionPath.name,
-        slug: serverlessFunctionModule.getPath
-          ? serverlessFunctionModule.getPath()
-          : "",
-      };
-    }
+    const defaultPath = functionFilepath
```
mkilpatrick, line 54: This is a weird name. Isn't this the filename?

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -1,128 +1,88 @@
...
+      .split("/functions/http/")[1]
+      .split(".")
+      .slice(-2)[0];
 
-    validateServerlessFunctionModuleInternal(serverlessFunctionInternal);
+    functionInternal = {
+      ...functionModule,
+      config: {
+        name: defaultPath,
```
mkilpatrick, line 62: The name should come from the FunctionConfig or the filename if http.

packages/pages/src/common/src/template/types.ts
```diff
@@ -52,7 +52,7 @@ export type TransformProps<T extends TemplateProps> = (props: T) => Promise<T>;
  *
  * @public
  */
-export type GetPath<T extends TemplateProps | void> = (props: T) => string;
+export type GetPath<T extends TemplateProps> = (props: T) => string;
```
mkilpatrick: This shouldn't take in any props. You should also use it as the type for getPath which you changed to ()=>string

benlife5: This file is now reverted to main, which does take props but it's for the feature templates

packages/pages/src/common/src/function/internal/types.ts
benlife5: I've changed the variable name to defaultSlug

I have to do it this way because the filename for `functions/http/get/getAll.ts` is `getAll` which ignores the nested directory structure

packages/pages/src/common/src/function/internal/types.ts
benlife5: similar to above -- we need to differentiate name in the case of `functions/http/get/index.ts` and `functions/http/post/index.ts` (unless we want to force all filenames to be unique)

packages/pages/src/common/src/function/internal/types.test.ts
benlife5: The tests I copied from features were comparing with JSON.stringify, which doesn't output functions.
I used a modifier to replace the function with it's name for testing but I can look on Monday if there's a better way to test it

packages/pages/src/common/src/function/internal/loader.ts
```diff
@@ -6,104 +6,95 @@ import {
...
       );
     }
     return prev.set(module.slug, module);
   }, new Map());
 };
 
 /**
- * A ServerlessFunctionModuleCollection is a collection of serverless function modules
+ * A FunctionModuleCollection is a collection of function modules
  * indexed by feature name.
```
mkilpatrick, line 86: "by name" since feature doesn't make sense here

packages/pages/src/common/src/function/internal/types.test.ts
```diff
@@ -92,23 +109,33 @@ describe("internal/types - convertFunctionModuleToFunctionModuleInternal", () =>
         false
       );
     const expected = {
-      default: "exampleFunction",
       config: {
```
mkilpatrick, line 112: I keep getting confused by the property names we have here
```
{
      config: {
        name: "Test Function",
        functionName: "exampleFunction",
        event: "API",
      },
      path: "src/functions/example.ts",
      filename: "example.ts",
      slug: "myFunction",
    };
```
because `path` really means path to the literal file while slug is the output of getPath. Can you update `path` to `filePath`?

packages/pages/src/dev/server/server.ts
```diff
@@ -61,22 +62,41 @@ export const createServer = async (
     ));
   }
 
-  const loadedFunctions = await loadServerlessFunctions();
-  const loadedFunctionPaths = [...loadedFunctions.keys()].map(
-    (slug) => "/" + slug
-  );
+  // Load functions from their source files
+  const functionModules: Map<string, FunctionModuleInternal> = new Map();
```
mkilpatrick, line 66: Can't you use FunctionModuleCollection here?

mkilpatrick: I know the comments for FMC was that it's indexed by name, but do we actually use it that way? If it's always by slug then update the comment.

packages/pages/src/dev/server/server.ts
```diff
@@ -61,22 +62,41 @@ export const createServer = async (
...
+          );
+        }
+        serveServerlessFunction(req, res, next, loadedFunction);
+      });
     });
   }
 
+  // Reload functions if their source files change
+  // Only supports updates in the default export, changes to the slug require server restart
+  vite.watcher.on("change", async (filepath) => {
```
mkilpatrick, line 94: Does HMR not work for these out of the box? If you refresh the function page does the updated code show up?

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,88 @@
...
+    const defaultSlug = functionFilepath
+      .split("/functions/http/")[1]
+      .split(".")
+      .slice(-2)[0];
+
+    functionInternal = {
+      ...functionModule,
+      config: {
+        name: defaultSlug,
+        functionName: functionModule.default?.name ?? functionPath.name,
```
benlife5, line 63: just "default"

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,140 @@
...
+  /** Site object containing all deploy-related information. */
+  site: Site;
+}
+
+/**
+ * The return value for an onPageGenerate plugin
+ * @public
+ */
+export interface OnPageGenerateReturnValue {
+  path: string;
```
benlife5, line 66: These need jsdoc descriptions but I'm not sure what they are, I just know they're required

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,140 @@
...
+  content: string;
+
+  redirects: string[];
+}
+
+/**
+ * The return value for an onUrlChange plugin
+ * @public
+ */
+export type OnUrlChange = HttpFunctionReturnValue;
```
benlife5, line 77: yext pages render is fine with onUrlChange plugins returning the same things as http plugins
Not sure if there's supposed to be anything different

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,116 @@
...
+  filePath: FunctionFilePath;
+  /** The exported config function */
+  config: FunctionConfigInternal;
+  /** The exported getPath function */
+  getPath?: () => string;
+  /** The exported function */
+  default?: ServerlessFunction;
+  /** The slug to host the function at on the Express dev server */
+  slug: {
+    production: string;
```
mkilpatrick, line 23: Can you give an example of these and explain why it's necessary?

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,116 @@
...
+ * A domain representation of a serverless function module. Contains all fields from an imported
+ * module as well as metadata about the module used in downstream processing.
+ */
+export interface FunctionModuleInternal {
+  /** The filepath to the serverless function file. */
+  filePath: FunctionFilePath;
+  /** The exported config function */
+  config: FunctionConfigInternal;
+  /** The exported getPath function */
+  getPath?: () => string;
```
mkilpatrick, line 18: We've gone back and forth a little on this. Does this still get incorporated if it's set? As in, does it override the path determined by the folder structure? If it does I think it's fine to leave but want to make sure it actually does something.

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,116 @@
...
+      },
+      filePath: functionFilepath,
+      slug: {
+        original: defaultSlug,
+        dev: defaultSlug.replaceAll("[", ":").replaceAll("]", ""),
+        production: defaultSlug.replaceAll("[", "{{").replaceAll("]", "}}"),
+      },
+      getPath: () => defaultSlug,
+    };
+  } else {
```
mkilpatrick, line 94: We might just want to error out if files exist outside of our 3 defined folders. We'll need to verify the functionality we want with Product. You can ask in #discuss-pagesjs and cc Andrew Sanchez.

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  if (
+    (functionType === "http" ||
+      functionType === "onUrlChange" ||
+      functionType === "onPageGenerate") &&
+    Object.keys(functionModule).length === 1
+  ) {
+    validateFunctionDefaultExport(functionFilepath.absolute, functionModule);
+
+    const defaultSlug = functionFilepath.relative.replace(
```
mkilpatrick, line 74: To my previous question, it looks like getPath does not override here. We should also ask Product if we want this at all. I could see value in it, but it might be better to start with less configurability to keep it simpler for users and documentation.

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -0,0 +1,116 @@
+import { ServerlessFunction, FunctionModule } from "../types.js";
+import {
+  validateFunctionDefaultExport,
+  validateFunctionModule,
+} from "./validateFunctionModule.js";
+import { convertToPluginEvent, PluginEvent } from "../../ci/ci.js";
```
mkilpatrick, line 6: If you move these types to here then ci.js doesn't need them at all. The internal type should already have the converted values by the time it needs to dump to the ci.json file.

packages/pages/src/common/src/function/internal/validateFunctionModule.ts
```diff
@@ -0,0 +1,36 @@
...
+export const validateFunctionDefaultExport = (
+  filepath: string,
+  serverlessFunction: FunctionModule
+) => {
+  if (!serverlessFunction.default) {
+    throw new Error(`${filepath} is missing a default export.`);
+  }
+};
+
+export const validateConfig = (
```
mkilpatrick, line 12: We don't need this now right? Or did you mean to validate the Internal type (vs external)? We can validate this for the internal but for external the user isn't defining this anymore.

packages/pages/src/common/src/function/internal/validateFunctionModule.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (!functionConfig.name) {
+    throw new Error(`${filepath} is missing "name" in the config function.`);
+  }
+};
+
+export const validateFunctionModule = (
+  filepath: string,
+  serverlessFunction: FunctionModule
+) => {
+  if (!serverlessFunction.config) {
```
mkilpatrick, line 25: Same comment for these. I don't think we need these anymore.

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,140 @@
+/**
+ * Defines the functions and fields that are available to the serverless function.
+ * @public
+ */
+export interface FunctionModule {
+  /** The exported config function */
+  config?: FunctionConfig;
```
mkilpatrick, line 7: Another similar comment. We need to check if these are relevant anymore for a user to define. I think the answer is going to be no.

packages/pages/src/common/src/function/types.ts
mkilpatrick: I'm not entirely sure either. Can you ask in #team-web-publishing-internal?

packages/pages/src/common/src/function/types.ts
mkilpatrick, line 77: This should be OnUrlChangeResponse

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  /** Site object containing all deploy-related information. */
+  site: Site;
+}
+
+/**
+ * The return value for an onPageGenerate plugin
+ * @public
+ */
+export interface OnPageGenerateReturnValue {
```
mkilpatrick, line 65: Change "ReturnValue" to "Response"

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
```diff
@@ -35,6 +42,16 @@ export default (projectStructure: ProjectStructure) => {
       throw new Error(e);
     }
 
+    finisher = logger.timedLog({ startLog: "Validating function modules" });
```
mkilpatrick, line 45: Validating functions, similar for the succeed msg

packages/pages/src/common/src/function/internal/types.ts
benlife5: I moved the convert function to internal/types

ci/ci.ts defines the types used in ci.json so I think PluginEvent still makes sense there
generate/ci.ts is where ci.json is actually written

packages/pages/src/dev/server/server.ts
benlife5: Yeah it doesn't work automatically for the functions. Vite registers a reload but then doesn't load the new code or something -- I'm not familiar with HMR so I'm not really sure why. I can dive into it more if you want

packages/pages/src/common/src/project/structure.ts
```diff
@@ -82,6 +84,7 @@ export const defaultProjectStructureConfig: ProjectStructureConfig = {
     distRoot: "dist",
     serverBundleOutputRoot: "assets/server",
     renderBundleOutputRoot: "assets/render",
+    functionBundleOutputRoot: "assets/functions",
```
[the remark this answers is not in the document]
benlife5: This gets defined in build.ts:116 
`entryPoints[`functions/${functionPath.relative}`] = functionPath.absolute;`
which seems to put everything in assets. Not sure if there's some config elsewhere to change that behavior

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -61,23 +59,33 @@ export const convertFunctionModuleToFunctionModuleInternal = (
...
-  const functionType = functionFilepath.relative.split("/")[1];
+  const functionType = functionFilepath.relative.split("/")[0];
+
+  if (
+    (functionType === "onUrlChange" || functionType === "onPageGenerate") &&
+    functionFilepath.relative.split("/").length > 2
+  ) {
+    throw new Error(
+      "Nested directories are not supported for onUrlChange and onPageGenerate" +
+        " plugins. All functions must be located in src/functions/onUrlChange or" +
```
mkilpatrick, line 70: must be located at the root of ...

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
```diff
@@ -44,8 +46,29 @@ export default (projectStructure: ProjectStructure) => {
 
     finisher = logger.timedLog({ startLog: "Validating functions" });
     try {
-      // loading the modules checks for all required aspects
-      await loadFunctions(projectStructure.functionBundleOutputRoot.path);
+      const functionFilepaths = getFunctionFilepaths("dist/functions");
+      await Promise.all(
+        functionFilepaths.map(async (filepath) => {
+          const buildResult = await esbuild.build({
```
mkilpatrick, line 52: You shouldn't need to run esbuild. The files have already been built at this point. We should just import them directly like we do in loader.ts to validate the template modules. You'll see that we we call it will transpile=false in `templateModules = await loadTemplateModules(serverBundles, false, true);`

packages/pages/src/common/src/function/internal/getFunctionFilepaths.ts
```diff
@@ -4,27 +4,43 @@ import { FunctionFilePath } from "./types.js";
...
+      // the matcher should be able to find a relative path
+      // with [0] being the absolute path and [1] being the relative path
       throw new Error(
         `Cannot resolve relative path from ${root} for ${filepath}`
       );
     }
+
     return {
+      // absolute: /Users/example/Desktop/mySite/src/functions/http/getInfo.ts
       absolute: filepath,
```
mkilpatrick, line 30: Thinking about this, it would be easier to just use  path.parse(filepath) to get some of the other things. Relative might be the only one you need to calculate yourself still. It would be much clearer though.

mkilpatrick: Additionally, maybe we should just return the path.ParsedPath too instead of each one of these things.

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
```diff
@@ -44,35 +44,41 @@ export default (projectStructure: ProjectStructure) => {
...
     }
 
+    /*
+     * Functions are bundled as mod.ts. This code runs as closeBuild.js. JS files cannot
+     * import TS, so we cannot simply import the function file. We also cannot do
+     * loadFunctionModules because that makes assumptions about the directory structure of src,
+     * not dist.
+     *
+     * This code makes a copy of mod.ts named mod.js so we can import it. After importing,
+     * it deletes the copy file and then checks the import for a default export.
```
mkilpatrick, line 54: It checks the default export before deleting the copy technically

packages/pages/src/dev/server/server.ts
mkilpatrick: If you refresh the page after saving your code change does it work without this? And with this what happens? The page will refresh after you save?

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -33,56 +33,133 @@ export interface FunctionConfigInternal {
...
   event: PluginEvent;
-  /** The default export's name */
+  /** The default export's name. */
   functionName: string;
 }
 
+/**
+ * Stores the filepath to a function.
+ */
+export interface FunctionFilePath {
```
mkilpatrick, line 43: Is this still used?

packages/pages/src/common/src/function/types.ts
```diff
@@ -3,34 +3,46 @@
  * @public
  */
 export interface FunctionModule {
-  /** The exported config function */
-  config?: FunctionConfig;
-  /** The exported getPath function */
-  getPath?: () => string;
   /** The exported function */
-  default?: ServerlessFunction;
+  default?: FunctionTypes;
```
mkilpatrick, line 7: Make this singular

packages/pages/tests/fixtures/src/functions/http/[param].js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
```
Comment (mkilpatrick) [inline packages/pages/tests/fixtures/src/functions/http/[param].js:2]: Why is this needed?

Comment (benlife5) [reply packages/pages/tests/fixtures/src/functions/http/[param].js:2]: It's there because `loadFunctionModules()` in `loader.ts` has the `transpile` argument so this file is to test `transpile = false`. `tests/fixtures/src/template.jsx` does the same for the template loader.

But, we actually don't load transpile functions anymore -- since the dist directory structure is different it would fail to load. The stuff with copying to a `.js` file yesterday is what replaces this transpiled load.

So, this file needs to exist if we want to test `transpile = false` but I suppose we could also remove that as an argument if it is never used?

benlife5:
The important parts of this latest push:

- `packages/pages/src/common/src/template/types.ts:98` Add onUrlChangeFunctionName to TemplateConfig
- `packages/pages/src/common/src/feature/features.ts:91` Add some logic to write the hashed onUrlChange function name to features.json
    - If there's no name in TemplateConfig or it cannot resolve the name to a file in src/functions/onUrlChange , it prints a warning but continues with no plugin name written to features.json for that feature
- `packages/pages/src/common/src/function/types.ts:97`  Added a type for an onUrlChange function argument. Probably need better jsdoc messages for those fields

New onUrlChange behavior:
- If there's nothing in `src/functions/onUrlChange` there are no onUrlChange plugins
- If there's one file in `src/functions/onUrlChange`, every feature automatically registers that plugin (this is not something we do, it's something the platform does)
- If there's >1 file in `src/functions/onUrlChange`, no feature automatically has an onUrlChangePlugin, an onUrlChangeFunctionName must be specified in a feature's TemplateConfig to register a plugin to that feature. 
    -Ex. onUrlChangeFunctionName = onChange2  if src/functions/onUrlChange/onChange2.ts is the desired plugin

packages/pages/src/common/src/function/types.ts
```diff
@@ -0,0 +1,181 @@
...
+  content: string;
+  /** A list of optional paths at which to install redirects to the page. */
+  redirects?: string[];
+}
+
+/**
+ * The argument passed to an onUrlChange type plugin.
+ * @public
+ */
+export interface OnUrlChangeArgument {
```
benlife5, line 97: Will likely need to update after this item is done: [Slack Discussion](https://yext.slack.com/archives/C02LLE9BW2K/p1689017518031249)

packages/pages/src/common/src/feature/features.ts
```diff
@@ -75,21 +87,48 @@ export const convertTemplateConfigToFeatureConfig = (
     alternateLanguageFields: config.alternateLanguageFields,
   };
 
+  // If an onUrlChange function name is specified in the feature config, lookup that plugin and
```
mkilpatrick, line 90: look up

packages/pages/src/common/src/feature/features.ts
```diff
@@ -75,21 +87,48 @@ export const convertTemplateConfigToFeatureConfig = (
...
+  // calculate it's hashed name for use in features.json
```
mkilpatrick, line 91: its

packages/pages/src/common/src/template/internal/types.ts
```diff
@@ -60,6 +60,8 @@ export interface TemplateConfigInternal {
   stream?: Stream;
   /** The specific fields to add additional language options to based on the stream's localization */
   alternateLanguageFields?: string[];
+  /** The name of the onUrlChange function to use. */
+  onUrlChangeFunctionName?: string;
```
mkilpatrick, line 64: This is just called onUrlChange on the external type. Let's keep it consistent.

mkilpatrick (approved):

mkilpatrick (approved):

PR #333, opened 2023-07-03
fix(dev): index page QA - update warning/info text and add favicon

benlife5:
- adds Yext favicon to the dev index page
- updates the info text in the sidebar
- removes the generate test data failed error/bug
- adjust warning div styles
- allows getPath to have non-standard URL characters, particularly a space. Limitation: a getPath with `%` in it will crash. 

### Encoded URI Strings
Vite seems to have inconsistent uses of URI encoded and decoded strings. `vite.transformIndexHtml` requires a valid URI string because it runs decode to get the html file path. However, it caches the transformations as the literal string provided and then looks them up using a decoded string.

**For example:**
Previously if `getPath` returned `'a b c'` it would be passed to `vite.transformIndexHtml` as `'a%20b%20c%20'`. The html file would be looked up as `path/to/abc.html` but the transformation would be stored as `'a%20b%20c%20'`. Then, the html transformation would look up `'a b c'`, creating the the bug. Now, the path is decoded before being sent to `vite.transformIndexHtml`, so it receives as `'abc'`, which passes through the decoder fine for the html and then also works for the transformation lookup.

The new limitation is that if `getPath` returns `'abc%'` the decoder throws an error because that is not a valid URI string. Similarly, if `getPath` returns an encoded string like `'abc%20'`, the `%` it gets encoded by the server creating `'abc%2520'` which then causes the original bug to occur.


packages/pages/src/dev/server/server.ts
```diff
@@ -43,20 +48,24 @@ export const createServer = async (
   // register vite's middleware
   app.use(vite.middlewares);
 
-  // Ignore favicon requests if it doesn't exist
+  // Load favicon for index page
+  app.use("/favicon.png", (req, res) => {
+    const faviconRelativePath = Favicon.default.slice(3); // the import has one too many "../"
```
mkilpatrick, line 53: This seems overly complicated to serve back a file. Is there a different way that works without need to change the paths? I know the favicon.ico is already being requested, hence the ignore part below. If our file is a .ico instead of .png would it just work?

packages/pages/src/bundler.js
```diff
@@ -36,7 +36,7 @@ const commonBuildOpts = {
   loader: {
     ".ts": "ts",
     ".html": "text",
-    ".png": "file",
+    ".png": "copy",
```
mkilpatrick, line 39: copy is a type?

packages/pages/src/dev/server/server.ts
```diff
@@ -50,11 +49,10 @@ export const createServer = async (
 
   // Load favicon for index page
   app.use("/favicon.png", (req, res) => {
-    const faviconRelativePath = Favicon.default.slice(3); // the import has one too many "../"
     res
       .status(200)
       .header({ "content-type": "image/png" })
-      .sendFile(path.resolve(__dirname, faviconRelativePath));
+      .sendFile(path.resolve(__dirname, "./public/favicon.png"));
```
mkilpatrick, line 55: Can we use an ico instead of png?

packages/pages/src/dev/server/middleware/sendAppHTML.ts
```diff
@@ -67,7 +67,9 @@ export default async function sendAppHTML(
   );
 
   const html = await vite.transformIndexHtml(
-    pathname,
+    // vite decodes request urls when caching proxy requests we have to
```
mkilpatrick, line 70: so we have to*

packages/pages/src/bundler.js
benlife5: [https://esbuild.github.io/content-types/#copy](https://esbuild.github.io/content-types/#copy)

From what I can tell, the two esbuild loaders for something like an .ico are `file` or `copy`. `file` creates a JS file with a default export being the relative path to the file. But, since `server.ts` isn't at the same directory level as the `server/public` folder, the relative path was incorrect and required string modification. 

The documentation reads like `copy` is supposed to move the image to the bundle root and then update imports. But I could only get it to keep the file where it is in the source code (and if I try to import it node throws a `Unknown file extension` error. However, since it's in the same spot as it is in the source code, I can just create a path in `server.ts` that points to the file.

mkilpatrick: Ah ok.

mkilpatrick (approved):

PR #334, opened 2023-07-03
fix(dev): duplicated generate-test-data

benlife5:
The updated `pages dev` called `pages generate features` then `yext pages generate-test-data` then `createServer()`.
However, `createServer()` calls `yext pages generate-test-data` in [server.ts:57](https://github.com/yext/pages/blob/8b3454f33a9ed35e2ebbf4455ad98eb9fa1cb2a1/packages/pages/src/dev/server/server.ts#L57)

The pages dev call is managed by `--noGenTestData` while the `createServer` call is managed by `--local` (ie a user would need both `--noGenTestData` and `--local` to disable generation).

This PR removes the `pages dev` call and leaves the original one in `createServer`.

### DIFF
--- a/packages/pages/src/dev/dev.ts
+++ b/packages/pages/src/dev/dev.ts
@@ -52,6 +52,7 @@ export const devCommandModule: CommandModule<unknown, DevArgs> = {
         demandOption: false,
       })
       .option("local", {
+        alias: "noGenTestData",
         describe: "Disables dynamically generated test data",
         type: "boolean",
         demandOption: false,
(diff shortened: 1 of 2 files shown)


packages/pages/src/dev/dev.ts
```diff
@@ -23,7 +23,6 @@ const handler = async ({
   noInit,
   scope,
   noGenFeatures,
-  noGenTestData,
```
mkilpatrick, line 26: Can we still use this flag to disable the call in the server?

benlife5: added an alias so you can use --local or --noGenTestData and it won't run the generate command

mkilpatrick (approved):

PR #338, opened 2023-07-06
perf(plugin)!: remove manifest from TemplateProps

jwartofsky-yext:
Removes the manifest from the props as it was causing changes to every file for new builds even if the template didn’t change. This caused excessive file uploads on the serving side because the file contents were different.


packages/yext-function/mod.ts
```diff
@@ -6,8 +6,10 @@ import { default as render } from "./assets/renderer/templateRenderer.js";
  * the result.
  */
 export const PagesGenerator = async (data): Promise<Record<any, any>> => {
-  return await render({
-    document: data.streamOutput,
-    __meta: { manifest },
-  });
+  return await render(
+    {
+      document: data.streamOutput,
```
mkilpatrick, line 11: Can you add the __meta.mode? It was missing so we may as well add it now. Should be "production" here.

mkilpatrick (approved):

PR #339, opened 2023-07-07
fix(dev): static page locale support

benlife5:
Support static page locales based on `features.json`

**`features.json` does not include key `locales` or `features.json` does not exist**
The dev server behaves as before, with default locale `en`.

**`features.json` includes `locales` with array length 1**
The dev server index page behaves as before, default locale is set to the locale in the `locales` array.

**`features.json` includes `locales` with array length >1**
The dev server index page includes a static page listing for each locale. The default local is set to the first locale in the `locales` array.

Screen Shot 2023-07-07 at 10 27 08 AM

### DIFF
--- a/packages/pages/src/dev/server/middleware/indexPage.ts
+++ b/packages/pages/src/dev/server/middleware/indexPage.ts
@@ -64,9 +64,9 @@ export const indexPage =
 
       // If there is any localData, display hyperlinks to each page that will be generated
       // from each data document.
-      if (localDataManifest.static.length + localDataManifest.entity.size) {
+      if (localDataManifest.static.size + localDataManifest.entity.size) {
         // If there are any data documents for static pages, render that section.
-        if (localDataManifest.static.length) {
+        if (localDataManifest.static.size) {
           indexPageHtml = indexPageHtml.replace(
             `<!--static-pages-html-->`,
             `<h3>Static Pages</h3>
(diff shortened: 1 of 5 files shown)


packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -64,9 +64,9 @@ export const indexPage =
 
       // If there is any localData, display hyperlinks to each page that will be generated
       // from each data document.
-      if (localDataManifest.static.length + localDataManifest.entity.size) {
+      if (localDataManifest.static.size + localDataManifest.entity.size) {
```
mkilpatrick, line 67: I think we should be using length over size: https://stackoverflow.com/questions/14202601/array-size-vs-array-length

packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -174,24 +174,40 @@ const addHttpFuncs = (indexPageHtml: string) => {
...
+              locales.length > 1
+                ? `<td>
+                <a href="http://localhost:${devServerPort}/${encodeURIComponent(
+                    staticURL
+                  )}?locale=${locale}">
+                  ${staticURL}?locale=${locale}
+                </a>
+              </td>`
+                : `<td>
+                <a href="http://localhost:${devServerPort}/${encodeURIComponent(
```
mkilpatrick, line 201: So if I have a single locale set to something other than "en" and we render without the ?locale it still works?

packages/pages/src/dev/server/server.ts
```diff
@@ -31,6 +32,30 @@ export const createServer = async (
...
+        ),
+        "utf-8"
+      )
+    );
+    if (featuresJson.locales && featuresJson.locales.length > 0) {
+      defaultLocale = featuresJson.locales[0];
+    }
+  } catch (e) {
+    if (e === "Error: ENOENT: no such file or directory") {
+      console.warn("Could not find features.json");
```
mkilpatrick, line 53: This is going to be the case most of the time so let's not log anything in this case.

mkilpatrick (commented):
What happens if you change the features.json list of locales and refresh the index page? Do the static templates update or does that require a server restart? Either is fine, I'm just wondering.

packages/pages/src/dev/server/middleware/serverRenderSlugRoute.ts
```diff
@@ -17,22 +17,28 @@ type Props = {
...
+      const locale = req.query.locale?.toString() ?? defaultLocale;
       const slug = decodeURI(url.pathname.substring(1));
 
       const templateFilepaths =
         getTemplateFilepathsFromProjectStructure(projectStructure);
       const matchingStaticTemplate: TemplateModuleInternal<any, any> | null =
         await findMatchingStaticTemplate(vite, slug, templateFilepaths);
       if (matchingStaticTemplate) {
-        sendStaticPage(
+        await sendStaticPage(
```
benlife5, line 41: `sendStaticPage` returns a promise. Without this await, if it doesn't match a locale, the server crashes. This await keeps the server up and sends the standard error 500 page.

I presume this is what we want, but I'm not sure if there's a reason `sendStaticPage` wasn't awaited previously

benlife5 (commented):
> What happens if you change the features.json list of locales and refresh the index page? Do the static templates update or does that require a server restart? Either is fine, I'm just wondering.

Seems to require a server restart. I think features.json is only read then?

packages/pages/src/dev/server/middleware/serverRenderSlugRoute.ts
mkilpatrick: Most likely a bug

mkilpatrick (approved):

PR #345, opened 2023-07-11
feat(dev)!: do not require localization.primary:false

benlife5:
Change the public type for `TemplateConfig.Stream.localization` 
to allow `{locales: [...]}` or `{primary: true}` 
but never `{primary: false}` or `{locales: [...], primary: true/false}`.

Under the hood, convert the public type to an internal type that always has `primary` defined with this behavior:

```localization: {} -> localization: {primary: true}
localization: {primary: true} -> localization: {primary: true}
localization: {locales: []} -> localization: {primary: true}
localization: {locales: ['fr']} -> localization: {locales: ['fr'], primary: false}
localization: {locales: ['fr'], primary: true} -> localization: {locales: ['fr'], primary: false}
```
Some of these inputs are not valid TS but we don't call `tsc` when we do `pages generate features` so they could exist.

Breaking because most repos will have `{locales: [...], primary: false}` which is now not allowed.

### DIFF
--- a/packages/pages/src/common/src/template/internal/types.ts
+++ b/packages/pages/src/common/src/template/internal/types.ts
@@ -10,7 +10,6 @@ import {
   TemplateModule,
   TemplateProps,
   TemplateRenderProps,
-  LocalizationOptions,
 } from "../types.js";
 import { validateTemplateModuleInternal } from "./validateTemplateModuleInternal.js";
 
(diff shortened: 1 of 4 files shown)


jwartofsky-yext (dismissed):
some checks failed

otherwise lgtm

packages/pages/src/common/src/template/internal/types.ts
```diff
@@ -57,13 +58,20 @@ export interface TemplateConfigInternal {
   /** The stream that this template uses. If a stream is defined the streamId is not required. */
   streamId?: string;
   /** The stream configuration used by the template */
-  stream?: Stream;
+  stream?: Stream<LocalizationInternal>;
```
mkilpatrick, line 61: You probably want a Stream and StreamInternal. The external types shouldn't rely on the Internal ones.

mkilpatrick (approved):

PR #348, opened 2023-07-12
fix(plugin): fail platform build when PagesJS build fails (fix #347)

### LINKED ISSUE #347: Platform build doesn't fail when PagesJS command fails
### Describe the bug

When a PagesJS command runs (ex 'pages build'), it does not return an exit code. This causes subsequent commands to run when PagesJS fails and creates confusing error logs.

Ex: in starting repo, running `npm run build` will call `yext pages build && yext pages generate-test-data && pages dev`. If there is an error in PagesJS build, it will continue to generate-test-data and error there.

### Reproduction

_No response_

### System Info

```shell
System:
    OS: macOS 12.6.3
    Memory: 1.66 GB / 32.00 GB
    Shell: 5.2.15 - /usr/local/bin/bash
  Binaries:
    Node: 18.16.1 - ~/.nvm/versions/node/v18.16.1/bin/node
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.1/bin/npm
    pnpm: 8.4.0 - /usr/local/bin/pnpm
  Browsers:
    Chrome: 114.0.5735.198
    Safari: 16.5
  npmPackages:
    @yext/pages: 1.0.0-beta.21 => 1.0.0-beta.21
```

### Used Package Manager

npm

### Used Browser

Chrome

### Validations

jwartofsky-yext:
PagesJS commands now throw error codes properly. Subsequent commands no longer get executed on failure.

Example with an error added in closeBundle
errrorInCloseBundle

### DIFF
--- a/packages/pages/src/bin/spawn.ts
+++ b/packages/pages/src/bin/spawn.ts
@@ -22,5 +22,5 @@ const results = spawnSync(
 );
 
 if (results?.status != undefined && results.status !== 0) {
-  process.exit(results.status);
+  process.exitCode = results.status;
 }
--- a/packages/pages/src/bin/spawn.ts
+++ b/packages/pages/src/bin/spawn.ts
@@ -13,10 +13,14 @@ const pathToPagesScript = path.resolve(
   "pages.js"
 );
 
-spawnSync(
+const results = spawnSync(
   "node",
   ["--experimental-vm-modules", pathToPagesScript, ...process.argv.slice(2)],
   {
     stdio: "inherit",
   }
 );
+
+if (results?.status != undefined && results.status !== 0) {
+  process.exit(results.status);
+}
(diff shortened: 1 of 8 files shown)


packages/pages/src/bin/spawn.ts
```diff
@@ -13,10 +13,14 @@ const pathToPagesScript = path.resolve(
...
-spawnSync(
+const results = spawnSync(
   "node",
   ["--experimental-vm-modules", pathToPagesScript, ...process.argv.slice(2)],
   {
     stdio: "inherit",
   }
 );
+
+if (results?.status != undefined && results.status !== 0) {
```
mkilpatrick, line 24: Pretty sure you can just do results?.status !== 0

jwartofsky-yext: TypeScript doesn't seem to like that
image

mkilpatrick: Hmm ok

mkilpatrick (dismissed):

mkilpatrick (approved):

PR #353, opened 2023-07-14
feat(dev/plugin): support React 18

mkilpatrick:
Tested against Node 17, 18, 19 for both React 17 and 18.


asanehisa (approved):

shrishayap:
Is the reason why some of the acceptance tests failed because of the new react version?

mkilpatrick:
> Is the reason why some of the acceptance tests failed because of the new react version? Otherwise LGTM
The Windows tests are failing but those were already failing. We'll have to investigate that but it shouldn't be a result of this change.

benlife5 (approved):

PR #356, opened 2023-07-19
refactor: restore command handlers

jwartofsky-yext:
Use command handlers that were used previously with Yargs.

As opposed to action written directly in command.

This also fixes 'useProdUrls' in dev.ts

Tested all commands and ran auto tests.

### DIFF
--- a/packages/pages/src/dev/dev.ts
+++ b/packages/pages/src/dev/dev.ts
@@ -8,8 +8,8 @@ import { ProjectFilepaths } from "../common/src/project/structure.js";
 
 interface DevArgs extends Pick<ProjectFilepaths, "scope"> {
   local?: boolean;
-  "prod-url"?: boolean;
-  "open-browser": boolean;
+  prodUrl?: boolean;
+  openBrowser: boolean;
   noInit?: boolean;
   scope?: string;
   noGenFeatures?: boolean;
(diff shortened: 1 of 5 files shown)


packages/pages/src/dev/dev.ts
```diff
@@ -4,6 +4,39 @@ import runSubProcess from "../util/runSubprocess.js";
...
+  }
+  if (!noGenFeatures)
+    await runSubProcess(
+      "pages generate features",
+      scope ? ["--scope" + " " + scope] : []
+    );
+
+  await createServer(!local, !!useProdURLs, scope);
+
+  if (openBrowser) await open(`http://localhost:${devServerPort}/`);
```
benlife5, line 38: The openBrowser argument still seems to not be passed through correctly

[JIRA ticket](https://yexttest.atlassian.net/browse/SUMO-5190?atlOrigin=eyJpIjoiYjQ5MTI3YjM1ZDAyNGYzODk4NDFmMWJmOTBkNTdiYjYiLCJwIjoiaiJ9)

benlife5 (commented):
Should also update prod.ts

packages/pages/src/dev/dev.ts
jwartofsky-yext: The openBrowser argument works. When running with npm run dev, the argument has to be passed to the inner command 'pages dev'. 

You can do so like this: `npm run dev -- --no-open-browser`

mkilpatrick (commented):
Update type and remove the scope to `refactor: restore...`

plugin refers to our Vite plugin, aka the prod build. You are just refactoring the binary commands which we don't have a scope for.

benlife5 (approved):

PR #359, opened 2023-07-19
feat(dev): default hydrate to true

asanehisa:
Due to changing direction, undid some of what shrisha did [earlier](https://github.com/yext/pages/commit/f6a71fc5667f5a081475a1c6377b7d5736e24942)


asanehisa:
@shrishayap let me know if it looks like I missed anything important. I did a similar test as you with a button and react counter to check when hydrate was true and false locally.

PR #365, opened 2023-07-28
fix(dev): hydration does not reflect template data mutations

### LINKED ISSUE #298: mutations to template data are persisted between server and client renders
### Describe the bug

If template code mutates the template data passed into it during the initial generation of the page, those changes to the data are present during the initial client side render. 

For example, I'm using [this modification](https://github.com/yextconsulting/pages-starter-react-consulting/compare/mutation-report?expand=1) to our starter template, connected to account 3861963.

I would expect this code change to reverse the order of the contents of the `dm_directoryChildren` field. But what happens is that the field is reversed once during generation, then reversed again during the hydration pass. 

The result is that the server rendered content doesn't match the client rendered content which triggers a React error. 
image
In Firefox, the specific way the difference is reconciled leads to the text of the directory links not matching the hrefs, which is a particularly bad outcome. 

In general we should avoid this kind of mutation to props in our frontend code, but I also think this particular situation shouldn't be possible. 

### Reproduction

See description

### System Info

```shell
System:
    OS: macOS 12.3
    Memory: 21.74 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.4.0 - ~/.nvm/versions/node/v17.4.0/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v17.4.0/bin/npm
  Browsers:
    Chrome: 112.0.5615.49
    Firefox: 111.0.1
    Safari: 15.4
  npmPackages:
    @yext/pages: 1.0.0-beta.16 => 1.0.0-beta.16
```

### Used Package Manager

npm

### Used Browser

Firefox

### Validations

benlife5:
fixes #298

### DIFF
--- a/packages/pages/src/dev/server/middleware/sendAppHTML.ts
+++ b/packages/pages/src/dev/server/middleware/sendAppHTML.ts
@@ -44,6 +44,8 @@ export default async function sendAppHTML(
   )) as RenderTemplate;
 
   const getServerHtml = async () => {
+    // using this wrapper function prevents SRR client-server mistmatches if
+    // the template modifies props
     return await serverRenderTemplateModule.render({
       Page: templateModuleInternal.default!,
       pageProps: props,
--- a/packages/pages/src/dev/server/middleware/sendAppHTML.ts
+++ b/packages/pages/src/dev/server/middleware/sendAppHTML.ts
@@ -43,10 +43,12 @@ export default async function sendAppHTML(
     clientServerRenderTemplates.serverRenderTemplatePath
   )) as RenderTemplate;
 
-  const serverHtml = await serverRenderTemplateModule.render({
-    Page: templateModuleInternal.default!,
-    pageProps: props,
-  });
+  const getServerHtml = async () => {
+    return await serverRenderTemplateModule.render({
+      Page: templateModuleInternal.default!,
+      pageProps: props,
+    });
+  };
 
   const headConfig = templateModuleInternal.getHeadConfig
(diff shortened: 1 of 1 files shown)


packages/pages/src/dev/server/middleware/sendAppHTML.ts
```diff
@@ -61,7 +63,7 @@ export default async function sendAppHTML(
 
   const clientInjectedServerHtml = getServerTemplateDev(
     clientHydrationString,
-    serverHtml,
+    await getServerHtml(),
```
mkilpatrick, line 66: This fixes the issue? Can you add a comment to the code about why we're doing this for future reference? Also, is there a way we could add a test for this?

mkilpatrick (commented):
Can you remove the issue number from the PR title? It's linked in the description. We don't need it in both places.

packages/pages/src/dev/server/middleware/sendAppHTML.ts
benlife5: Added comment. Honestly, I'm not entirely sure why this works, but it does. 

I started to look into testing it, I think it's doable but it takes a good bit of setup, possibly creating a test vite+express server. I think I have some higher priority items right now so I stopped but I'm happy to loop back to it if you want.

mkilpatrick (approved):

PR #375, opened 2023-08-14
feat(dev): export createServer and add --port arg

oshi97:
This PR exports createServer as createDevServer, and adds a port CLI arg and also to createServer itself.
The main purpose of this is to address inconsistencies with spawning pages as a child process from within `@yext/studio`.
But I think these changes are also useful outside of that purpose.

J=SLAP-2885
TEST=manual

test that I can start up the playground/locations-site with a --port arg
test that I can import createDevServer in a separate package and start the pages dev server with it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ lerna-debug.log*
 
 dist
 *.local
+**/.yextrc
 
 # Editor directories and files
 .vscode/*
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ lerna-debug.log*
 
 dist
 *.local
-**/.yextrc
 
 # Editor directories and files
 .vscode/*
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -17,6 +17,7 @@ lerna-debug.log*
 
 dist
 *.local
+**/.yextrc
```
mkilpatrick, line 20: This shouldn't be ignored. The original purpose is for ProServe who often swap between repos often and need to re-init. This makes it easy so they don't need to figure out which business it's for and it auto-inits when they start dev'ing.

oshi97: I'm a little confused, is ProServe helping to develop yext/pages? This gitignore only affects development within this repo right?

oshi97: This gitignore doesn't somehow affect https://github.com/yext/pages-starter-react-locations does it?

mkilpatrick: Yea that's true. Though that file shouldn't even exist in this repo.

mkilpatrick (approved):

.gitignore
oshi97: Oh it's from when I was testing in the playground folder.

PR #376, opened 2023-08-17
fix(dev/plugin)!: redo reverse proxy setup

mkilpatrick:
Prior to this change, assets were imported relatively, also taking into account relativePathToPrefix. While this generally worked, there were other issues due to Vite's build optimizations: specifically module preloading and inline assets. Due to this we decided to take a new approach where assets are now imported absolutely and it's now up to the RP config to handle the assets. A working example of a Netlify setup for a RP at the `subdirectory` subfolder looks like this:

```
[[redirects]]
  from = "/subdirectory/assets/*"
  to = "https://rp-test.pagesprod.yextengtest.com/subdirectory/assets/:splat"
  status = 200
  force = true
  headers = {Host = "https://rp-test.pagesprod.yextengtest.com/"}
  
[[redirects]]
  from = "/subdirectory/*"
  to = "https://rp-test.pagesprod.yextengtest.com/:splat"
  status = 200
  force = true
  headers = {Host = "https://rp-test.pagesprod.yextengtest.com/"}
  
[[redirects]]
  from = "/subdirectory"
  to = "https://rp-test.pagesprod.yextengtest.com/index.html"
  status = 200
  force = true
  headers = {Host = "https://rp-test.pagesprod.yextengtest.com/"}
```

In order to properly configure the Vite side, a user must set the assetsDir in the vite.config.js (`subdirectory/assets` for the above example) or in the config.yaml (which will be globally supported in some upcoming PRs).

Other changes in the PR include a massive refactor of ProjectStructure in order to simplify its usage and is now being used everywhere, meaning that all filenames and folders are fully configurable.


packages/pages/src/common/src/assets/getAssetsFilepath.ts
```diff
@@ -0,0 +1,39 @@
...
+export const determineAssetsFilepath = async (
+  defaultAssetsDir: string,
+  configYamlPath: string,
+  viteConfigPath: string
+) => {
+  if (configYamlPath === "" || viteConfigPath === "") {
+    return defaultAssetsDir;
+  }
+
+  if (fs.existsSync(configYamlPath)) {
```
mancusi, line 25: I feel like the logic to read the config.yaml file should happen elsewhere and shouldn't be specific to asset filepath determination.

Other places need to read it anyways right?

packages/pages/src/common/src/assets/getAssetsFilepath.ts
```diff
@@ -0,0 +1,39 @@
...
+import { UserConfig } from "vite";
+import { import_ } from "./import.js";
+import yaml from "js-yaml";
+import { ConfigYaml } from "../config/config.js";
+
+/**
+ * Determines the assets directory to use by checking the following, in order:
+ * 1. config.yaml
+ * 2. vite.config.json assetDir
+ * 3. default to "assets"
```
mancusi, line 12: this default is also the vite default. Is there a way we can just use the vite default?

packages/pages/src/common/src/function/internal/types.ts
```diff
@@ -50,19 +45,24 @@ export interface FunctionConfigInternal {
...
 export const convertFunctionModuleToFunctionModuleInternal = (
   functionFilepath: path.ParsedPath,
-  functionModule: FunctionModule
+  functionModule: FunctionModule,
+  projectStructure: ProjectStructure
 ): FunctionModuleInternal => {
-  let functionInternal;
+  const serverlessSourcePath =
+    projectStructure.config.rootFolders.source +
+    "/" + // purposefully using string concatentation and not path.join
```
mancusi, line 53: can you explain why? It's because this is for importing modules right?

mancusi: What if the project structure has `\` instead? does that get properly handled? Maybe something on the project structure end to canonicalize things?

packages/pages/src/common/src/project/structure.ts
```diff
@@ -71,25 +87,52 @@ export interface EnvVar {
...
+    features: "features.json",
+    siteStream: "site-stream.json",
+    serving: "serving.json",
   },
-  filenamesConfig: {
-    ciConfig: "ci.json",
-    featuresConfig: "features.json",
-    siteStreamConfig: "site-stream.json",
+  rootFiles: {
+    config: "config.yaml",
```
mancusi, line 135: not sure how the config.yaml stuff works yet but does this handle scoping?

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -98,22 +91,11 @@ export const getGlobalClientServerRenderTemplates = (
...
-      globalFilename
-    );
-    if (fs.existsSync(pathToGlobalFile)) {
-      return [pathToGlobalFile, true];
-    }
-  }
-
   const pathToGlobalFile = path.join(
-    templatesRootPath.getAbsolutePath(),
+    templatePaths[0].getAbsolutePath(),
```
mancusi, line 98: hm it's not clear to me why we just choose the first path here.

packages/pages/src/dev/server/middleware/indexPage.ts
```diff
@@ -159,7 +159,7 @@ export const indexPage =
       }
 
       const functionsList = [
-        ...(await loadFunctions("src/functions")).values(),
+        ...(await loadFunctions("src/functions", projectStructure)).values(),
```
mancusi, line 162: should this default be accounted for automatically by projectStructure?

packages/pages/src/generate/ci/ci.ts
```diff
@@ -142,18 +154,20 @@ export const getUpdatedCiConfig = async (
...
+  return {
+    pluginName: "PagesGenerator",
+    sourceFiles: [
+      {
+        root: `${projectStructure.config.rootFolders.dist}/${projectStructure.config.subfolders.plugin}`,
+        pattern: "*{.ts,.json}",
+      },
+      {
+        root: `${projectStructure.config.rootFolders.dist}`,
+        pattern: `${projectStructure.config.subfolders.assets}/{${projectStructure.config.subfolders.serverBundle},${projectStructure.config.subfolders.static},${projectStructure.config.subfolders.renderer},${projectStructure.config.subfolders.renderBundle}}/**/*{.js,.css}`,
```
mancusi, line 167: const {assets, renderer, serverBundle, static, renderBundle} = projectStructure.config.subfolders;

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -60,22 +57,11 @@ export const build = (projectStructure: ProjectStructure): Plugin => {
             },
           },
           reportCompressedSize: false,
+          assetsDir: projectStructure.config.subfolders.assets,
         },
-        define: processEnvVariables(projectStructure.envVarPrefix),
-        base: "", // makes static assets relative for reverse proxies and doesn't affect non-RP sites
```
mancusi, line 65: I think it's probably worth detailing how someone would migrate their assets to the new way if they've already updated to this case.

A migration to 1.0 guide will be very useful.

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -150,16 +137,15 @@ const discoverRenderTemplates = (
...
   // Move the [compiled] _server.ts and _client.ts render template to /assets/render
   const clientServerRenderTemplates = getGlobalClientServerRenderTemplates(
-    projectStructure.templatesRoot,
-    projectStructure.scopedTemplatesPath
+    projectStructure.getTemplatePaths()
   );
 
   // server
-  entryPoints["render/_server"] =
+  entryPoints[`${projectStructure.config.subfolders.renderBundle}/_server`] =
```
mancusi, line 144: you could destructure here as well to make things easier to read.

packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
```diff
@@ -19,9 +20,13 @@ const pathToModule = new Map();
...
 export const readTemplateModules = async (
   feature: string,
-  manifest: Manifest
+  manifest: Manifest,
+  projectStructure: ProjectStructure
 ): Promise<TemplateModuleInternal<any, any>> => {
-  const path = manifest.bundlePaths[feature].replace("assets", "..");
+  const path = manifest.bundlePaths[feature].replace(
+    projectStructure.config.subfolders.assets,
+    ".."
```
mancusi, line 28: now that the assets/template directory is always configurable it won't always be one directory away from it right?

packages/pages/src/vite-plugin/build/closeBundle/bundleValidator.ts
```diff
@@ -18,9 +19,16 @@ export const validateBundles = () => {
   validateTotalSourceSize(sizeOfAllBundles);
 };
 
-const getBundlePaths = (): string[] => {
+const getBundlePaths = (projectStructure: ProjectStructure): string[] => {
   return glob.sync(
-    `${path.resolve("dist/assets")}/{render,renderer,server,static}/**/*.*`
+    `${path.resolve(
+      projectStructure.config.rootFolders.dist,
```
mancusi, line 25: destructure for clarify.

packages/pages/src/vite-plugin/build/closeBundle/closeBundle.ts
```diff
@@ -22,22 +22,30 @@ export default (projectStructure: ProjectStructure) => {
   return async () => {
     let finisher = logger.timedLog({ startLog: "Validating template modules" });
     let templateModules: TemplateModuleCollection;
+
     try {
       const serverBundles = glob.sync(
         path.join(
-          projectStructure.serverBundleOutputRoot.getAbsolutePath(),
+          path.resolve(
```
mancusi, line 29: destructure

packages/pages/src/common/src/project/structure.ts
```diff
@@ -112,69 +155,59 @@ export type Optional<T> = {
...
-      this.#config.filepathsConfig.sitesConfigRoot
-    );
-    this.templatesRoot = new Path(this.#config.filepathsConfig.templatesRoot);
-    this.serverlessFunctionsRoot = new Path(
-      this.#config.filepathsConfig.functionsRoot
+    const mergedConfig = merge(defaultProjectStructureConfig, config);
+    this.config = mergedConfig;
+  }
+
+  static init = async (
```
mancusi, line 165: Won't we need some validation here to ensure that these paths don't conflict with each other? Are relative paths allowed in the config options? if so you could probably get into some weird combinations.

mancusi (changes requested):

packages/pages/src/common/src/assets/getAssetsFilepath.ts
mkilpatrick: That's a good callout. I had the config.yaml stuff as a TODO but decided to add it in here now since I already wrote the code and tests for reading a file. I will refactor this in an upcoming PR when we introduce the true consolidated config work. Going to leave here for now though.

packages/pages/src/common/src/assets/getAssetsFilepath.ts
mkilpatrick: Since I'm just reading the file directly and not going through Vite where it actually merges the user-defined config with its internal defaults I don't have a way to get that. It doesn't actually matter what this is anyways because it ultimately gets set on the Vite config assetDir during the plugin build.

packages/pages/src/common/src/function/internal/types.ts
mkilpatrick: This deals with the functions foldering to determine the function path/url. It's possible this is what broke Windows. If that's the case I'll fix in another PR. I didn't want to change the logic though of what was here before.

Why would the project structure have a `\`? You're saying if someone tried to do something custom? That seems weird unless they're doing something weird for Windows. I'm leaning towards not changing this as part of this change.

packages/pages/src/common/src/project/structure.ts
mkilpatrick: I think scope will be an attribute on the config, though not entirely sure. It shouldn't matter for this change though.

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
mkilpatrick: TAL at `getTemplatePaths` in structure.ts. If there's a scope defined it returns [the scoped templates, non-scoped templates]. If no scope is defined it's just [non-scoped templates]. The 0 index always gives us the appropriate set of templates, whether scoped or not.

packages/pages/src/vite-plugin/build/build.ts
mkilpatrick: What do you mean by destructure here?

packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
mkilpatrick: This has to do with the plugin files in Deno. We set the sourceFiles in ci.json with the follwing:

```
{
    "root": "dist",
    "pattern": "subdirectory/assets/{server,static,renderer,render}/**/*{.js,.css}"
}
```

I believe the way that it works is that everything copied at the pattern path actually gets copied to the root, and since we are replacing the full path here it is actually relative only one level deep to that root.

packages/pages/src/common/src/project/structure.ts
mkilpatrick: Hmm, that's an interesting point. I've been tempted for awhile to just remove this config as a plugin option anyways. When would people ever use it? Maybe in a future world of custom stuff or trying to support some other framework? Validation will be necessary if we keep it (or even just for the assetDir you can configure), but it will be pretty annoying. Check that literally no field overlaps or matches and there are no `\`. I will add validation in another PR.

packages/pages/src/common/src/function/internal/types.ts
mancusi: > Why would the project structure have a \? You're saying if someone tried to do something custom?

We're treating the file structure as both the literal filepath and also as a valid npm import path. On windows, filepaths must use `\` but that'd be invalid for npm. I was just calling it out that if someone configured these paths on windows I think it'd break things.

mancusi (commented):
LGTM for the stuff i haven't yet commented on but still a few unresolved conversations

packages/pages/src/common/src/function/internal/types.ts
mkilpatrick: You're correct, but they should be setting them as npm paths, not local OS paths. We use path.resolve everywhere to handle fixing the slashes when necessary. I'll add validation for backslash in a future PR. I wonder if Vite validates this type of thing...nope. They just let it go through. So it's on the user to not do something silly.

mancusi (approved):

PR #378, opened 2023-08-18
fix: async serverless functions and body parsing

mkilpatrick:
This fixes an issue where the request body was not properly parsed and sent to the function. It also fixes async functions not working.


asanehisa (dismissed):
lgtm

packages/pages/src/dev/server/middleware/serveHttpFunction.ts
```diff
@@ -15,11 +15,14 @@ export const serveHttpFunction = async (
   const argument: HttpFunctionArgument = {
     queryParams: req.query as { [p: string]: string },
     pathParams: req.params,
+    method: req.method,
```
jwartofsky-yext, line 18: Doesn't really matter, but it's method header body here where elsewhere its headers method body

jwartofsky-yext (dismissed):
lgtm

PR #386, opened 2023-09-06
fix(dev): fix dev and tests on Windows

jwartofsky-yext:
Fixes issues with paths in various places.

Convert to posix paths in globSync and Windows paths for filepaths when
appropriate.

Unit tests pass and npm run dev works as expected.
Pages are fully functional and appear the same as on MacOS.

### DIFF
--- a/packages/pages/src/bin/spawn.ts
+++ b/packages/pages/src/bin/spawn.ts
@@ -6,9 +6,9 @@ import path from "path";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+const filePrefix = path.sep !== path.posix.sep ? "file:\\\\" : "";
 const pathToPagesScript = path.resolve(__dirname, "./pages.js");
-
-const pathToLoader = path.resolve(__dirname, "./loader.js");
+const pathToLoader = filePrefix + path.resolve(__dirname, "./loader.js");
 
 const nodeVersion = Number(
   spawnSync("node", ["-v"], { encoding: "utf-8" })
--- a/packages/pages/src/bin/spawn.ts
+++ b/packages/pages/src/bin/spawn.ts
@@ -6,7 +6,7 @@ import path from "path";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
-const filePrefix = path.sep !== path.posix.sep ? "file:\\\\" : "";
+const filePrefix = path.sep === path.win32.sep ? "file:\\\\" : "";
 const pathToPagesScript = path.resolve(__dirname, "./pages.js");
 const pathToLoader = filePrefix + path.resolve(__dirname, "./loader.js");
 
--- a/packages/pages/src/common/src/function/internal/getFunctionFilepaths.ts
+++ b/packages/pages/src/common/src/function/internal/getFunctionFilepaths.ts
@@ -1,6 +1,7 @@
 import { globSync } from "glob";
 import path from "path";
 import { Path } from "../../project/path.js";
+import { convertToPosixPath } from "../../template/paths.js";
 
 /**
  * Get all the functions files in a directory
@@ -11,7 +12,9 @@ export const getFunctionFilepaths = (root: string): path.ParsedPath[] => {
   const functionsRoot = new Path(root); // resolve the functions root
   // Get all js/ts files in the directory or subdirectories
   const filepaths = globSync(
-    `${functionsRoot.getAbsolutePath()}/**/*.{tsx,jsx,js,ts}`
+    convertToPosixPath(
+      `${functionsRoot.getAbsolutePath()}/**/*.{tsx,jsx,js,ts}`
+    )
   );
 
   return filepaths.map((filepath) => {
--- a/packages/pages/src/common/src/function/internal/loader.test.ts
+++ b/packages/pages/src/common/src/function/internal/loader.test.ts
@@ -1,7 +1,8 @@
 import path from "path";
 import { loadFunctionModules, FunctionModuleCollection } from "./loader.js";
 import { ProjectStructure } from "../../project/structure.js";
-import { convertToOSPath } from "../../template/paths.js";
+
+const httpPath = path.join("tests", "fixtures", "src", "functions", "http");
 
 // our jest configuration doesn't support file urls so update pathToFileURL to do nothing during
 // this test.
@@ -27,9 +28,7 @@ describe("loadTemplateModules", () => {
 
   it("loads and transpiles raw templates", async () => {
     const functionFile: path.ParsedPath[] = [
-      path.parse(
-        convertToOSPath("tests/fixtures/src/functions/http/[param].ts")
-      ),
+      path.parse(path.join(httpPath, "[param].ts")),
     ];
 
     const functionModules = await loadFunctionModules(
@@ -42,7 +41,7 @@ describe("loadTemplateModules", () => {
 
   it("loads transpiled templates", async () => {
     const functionFile: path.ParsedPath[] = [
-      path.parse(path.resolve("tests/fixtures/src/functions/http/[param].js")),
+      path.parse(path.resolve(path.join(httpPath, "[param].js"))),
     ];
     const functionModules = await loadFunctionModules(
       functionFile,
--- a/packages/pages/src/common/src/function/internal/loader.test.ts
+++ b/packages/pages/src/common/src/function/internal/loader.test.ts
@@ -2,8 +2,6 @@ import path from "path";
 import { loadFunctionModules, FunctionModuleCollection } from "./loader.js";
 import { ProjectStructure } from "../../project/structure.js";
 
-const httpPath = path.join("tests", "fixtures", "src", "functions", "http");
-
 // our jest configuration doesn't support file urls so update pathToFileURL to do nothing during
 // this test.
 jest.mock("url", () => {
@@ -28,7 +26,7 @@ describe("loadTemplateModules", () => {
 
   it("loads and transpiles raw templates", async () => {
     const functionFile: path.ParsedPath[] = [
-      path.parse(path.join(httpPath, "[param].ts")),
+      path.parse(path.resolve("tests/fixtures/src/functions/http/[param].ts")),
     ];
 
     const functionModules = await loadFunctionModules(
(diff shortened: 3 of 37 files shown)


asanehisa:
After snapshots get added via github, you need to do an empty commit to trigger tests

packages/pages/src/common/src/template/paths.ts
```diff
@@ -21,5 +23,12 @@ export const getRelativePrefixToRootFromPath = (path: string): string => {
...
  */
 export const convertToPosixPath = (p: string) => {
-  return p.split(path.sep).join(path.posix.sep);
+  return p.split(winSep).join(path.posix.sep);
+};
+
+/**
+ * Convert any path to match the format of the OS. If it already matches no change occurs.
+ */
+export const convertToOSPath = (p: string) => {
```
mkilpatrick, line 32: Unused

packages/pages/src/common/src/function/internal/getFunctionFilepaths.ts
```diff
@@ -11,7 +12,9 @@ export const getFunctionFilepaths = (root: string): path.ParsedPath[] => {
   const functionsRoot = new Path(root); // resolve the functions root
   // Get all js/ts files in the directory or subdirectories
   const filepaths = globSync(
-    `${functionsRoot.getAbsolutePath()}/**/*.{tsx,jsx,js,ts}`
+    convertToPosixPath(
```
mkilpatrick, line 15: Noting that this is correct. globSync requires posix separators.

packages/pages/src/common/src/template/paths.ts
```diff
@@ -21,5 +23,12 @@ export const getRelativePrefixToRootFromPath = (path: string): string => {
  * paths are supported with import.
  */
 export const convertToPosixPath = (p: string) => {
-  return p.split(path.sep).join(path.posix.sep);
+  return p.split(winSep).join(path.posix.sep);
```
mkilpatrick, line 26: You can leave the split as path.sep since it will be whatever OS it's run on.

packages/pages/src/common/src/template/paths.ts
```diff
@@ -23,12 +21,5 @@ export const getRelativePrefixToRootFromPath = (path: string): string => {
...
 export const convertToPosixPath = (p: string) => {
-  return p.split(winSep).join(path.posix.sep);
-};
-
-/**
- * Convert any path to match the format of the OS. If it already matches no change occurs.
- */
-export const convertToOSPath = (p: string) => {
-  return p.split(path.posix.sep).join(path.sep).split(winSep).join(path.sep);
+  return p.split("\\").join(path.posix.sep);
```
mkilpatrick, line 24: Use path.sep in the split

jwartofsky-yext: The reason for this is that convertToPosixPath would fail on Posix systems. If we are are on on a Posix system and are given a Windows path, convertToPosixPath will not do anything when using path.sep. Using "\\" works on both Windows and Posix.

This case only occurs in unit tests, but I still think we would want this method to work on both.

packages/pages/src/bin/spawn.ts
```diff
@@ -6,9 +6,9 @@ import path from "path";
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+const filePrefix = path.sep !== path.posix.sep ? "file:\\\\" : "";
```
mkilpatrick, line 9: Annoying you had to do this. Can you instead make this the positive and do 

`path.sep === path.win32.sep ? "file..." : "";`

packages/pages/src/bin/spawn.ts
```diff
@@ -6,9 +6,9 @@ import path from "path";
...
 const pathToPagesScript = path.resolve(__dirname, "./pages.js");
-
-const pathToLoader = path.resolve(__dirname, "./loader.js");
+const pathToLoader = filePrefix + path.resolve(__dirname, "./loader.js");
```
mkilpatrick, line 11: Doesn't dirname include the Windows drive prefix, like `C:`? (I think it was actually D the Github actions. Is that the correct thing? `file:\\\\D:\\path\to\loader.js`?

packages/pages/src/common/src/function/internal/loader.test.ts
```diff
@@ -26,27 +28,35 @@ describe("loadTemplateModules", () => {
...
 
     const functionModules = await loadFunctionModules(
       functionFile,
       true,
       projectStructure
     );
-    commonTests(functionModules, "param-47543");
+    if (path.sep === path.posix.sep) {
+      commonTests(functionModules, "param-47543");
+    } else {
```
mkilpatrick, line 41: Why do we need an else case? The hash we add is the path up to and including the filename. I assume that's the issue. Instead, we should convertToPosixPath in the function code that generates the hash.

See the `unsecureHashPluginName(relativePath)` line in functionMetadataParser.ts

packages/pages/src/common/src/function/internal/loader.test.ts
```diff
@@ -26,27 +28,35 @@ describe("loadTemplateModules", () => {
...
+      commonTests(functionModules, "param-19926");
+    }
   });
 
   it("loads transpiled templates", async () => {
     const functionFile: path.ParsedPath[] = [
-      path.parse(path.resolve("tests/fixtures/src/functions/http/[param].js")),
+      path.parse(path.resolve(path.join(httpPath, "[param].js"))),
```
mkilpatrick, line 48: I don't think you need path.join if you're using path.resolve

> All methods on path should work cross-platform, and use the platform-specific separator.

https://github.com/nodejs/node/issues/12156#issuecomment-290912558

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -21,7 +22,9 @@ export const getTemplateFilepaths = (paths: Path[]): string[] => {
   const templateFilepaths: string[] = [];
   const addedFilenames: Set<string> = new Set();
   paths.forEach((p) => {
-    const filepaths = globSync(`${p.getAbsolutePath()}/*.{tsx,jsx,js,ts}`);
+    const filepaths = globSync(
+      convertToPosixPath(`${p.getAbsolutePath()}/*.{tsx,jsx,js,ts}`)
```
mkilpatrick, line 26: There's a slash here. Is it working ok?

packages/pages/src/util/runSubprocess.test.ts
```diff
@@ -8,6 +9,10 @@ describe("runSubprocess", () => {
 
   it("returns non-zero when subprocess fails", async () => {
     const exitCode = await runSubprocess("echo", ["$((0/0))"]);
-    expect(exitCode).not.toEqual(0);
+    if (path.sep === path.posix.sep) {
+      expect(exitCode).not.toEqual(0);
+    }
+    // commands do not return exit codes on Windows, so we skip this test in that case
```
mkilpatrick, line 15: Is this true? Wouldn't that mean any process we spawn that results in an exit code because we throw means it just hangs on Windows forever and never shuts down?

packages/pages/src/bin/spawn.ts
jwartofsky-yext: Yes that is the correct thing. It expects `file:\\\\c:\\path\\etc`

packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
jwartofsky-yext: Yes, globSync() requires a Posix path (even on Windows).
Just flipping the slashes is ok though, so it can accept a path like C:/root/myLocation/tacos.exe
https://stackoverflow.com/questions/55512907/node-glob-sync-returns-empty-array

packages/pages/src/util/runSubprocess.test.ts
jwartofsky-yext: It exits with code 0, so there is a code, but it doesn't get set.
It might be getting set in "errorlevel"

The command ${echo (0/0)} will produce an exit code of 1 in MacOs but 0 in Windows.

mkilpatrick: This is because you can't execute a command during echo in Windows. It just returns it as a string. Try doing this:

`await runSubprocess("Set", ["/a", "foo=0/0"]);`

This will fail in unix because `Set` isn't a command. In Windows it will also fail with a divide by 0 error.

packages/pages/src/dev/server/server.ts
```diff
@@ -51,7 +51,7 @@ export const createServer = async (
       projectStructure.config.envVarConfig.envVarPrefix
     ),
     optimizeDeps: {
-      include: ["react-dom", "react-dom/client"],
+      include: ["react-dom"],
```
mkilpatrick, line 54: The client was required for react 18 even though I know it throws a warning. Did this actually change anything for you other than the console warning?

packages/pages/src/util/runSubprocess.test.ts
jwartofsky-yext: It looks like it's actually that the echo command doesn't return the exit code on Windows when it tries to print $((0/0)), even though there is an error inside. On MacOS the error gets bubbled up.

I changed the test to run a powershell command that will throw an error with exit code when on Windows.

packages/pages/src/vite-plugin/build/closeBundle/manifest.ts
```diff
@@ -29,11 +29,13 @@ export const generateManifestFile = (
 
   // Add the renderPaths to the manifest. This defines the _client and _server entries.
   const renderPaths = glob.sync(
```
mkilpatrick, line 31: Make sure that all of the paths that are output to dist/plugin/manifest.json are posix-style.

packages/pages/src/dev/server/server.ts
jwartofsky-yext: It just cleared the console warning for me. I'll change it back.

packages/pages/src/vite-plugin/build/closeBundle/manifest.ts
jwartofsky-yext: Just checked. All of the paths there are posix-style.

mkilpatrick (approved):

PR #387, opened 2023-09-08
fix(dev): use ts-morph to parse template configs

mkilpatrick:
Previous to this change, we used esbuild to compile the templates into modules in order to pull off the config const. This led to various issues for people. It works fine with Vite since Vite does a lot of CJS/polyfill magic, but we weren't handling that with esbuild. This was overkill anyways as we only need to grab the config off of the templates, so compilation is irrelevant at this step.

### DIFF
--- a/packages/pages/src/generate/features/features.ts
+++ b/packages/pages/src/generate/features/features.ts
@@ -1,21 +1,15 @@
 import { ProjectStructure } from "../../common/src/project/structure.js";
 import { getTemplateFilepaths } from "../../common/src/template/internal/getTemplateFilepaths.js";
-import { loadTemplateModules } from "../../common/src/template/internal/loader.js";
-import { createFeaturesJson } from "./createFeaturesJson.js";
 import { Command } from "commander";
+import { createTemplatesJson } from "../templates/createTemplatesJson.js";
 
 const handler = async ({ scope }: { scope: string }): Promise<void> => {
   const projectStructure = await ProjectStructure.init({ scope });
   const templateFilepaths = getTemplateFilepaths(
     projectStructure.getTemplatePaths()
   );
-  const templateModules = await loadTemplateModules(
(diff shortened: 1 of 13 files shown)


asanehisa (approved):
also not entirely sure how ts-morph works, but if it works!

packages/pages/src/generate/features/features.ts
```diff
@@ -1,21 +1,15 @@
...
     projectStructure.getTemplatePaths()
   );
-  const templateModules = await loadTemplateModules(
-    templateFilepaths,
-    true,
-    false
-  );
 
-  createFeaturesJson(templateModules, projectStructure);
+  createTemplatesJson(templateFilepaths, projectStructure, "FEATURES");
```
jwartofsky-yext, line 12: Could "FEATURES" and "TEMPLATES" be an enum?

packages/pages/src/generate/templates/templates.ts
```diff
@@ -9,13 +8,8 @@ const handler = async ({ scope }: { scope: string }): Promise<void> => {
...
     projectStructure.getTemplatePaths()
   );
-  const templateModules = await loadTemplateModules(
-    templateFilepaths,
-    true,
-    false
-  );
 
-  createTemplatesJson(templateModules, projectStructure);
+  createTemplatesJson(templateFilepaths, projectStructure, "TEMPLATES");
```
jwartofsky-yext, line 12: same enum comment

colton-demetriou (approved):

jwartofsky-yext (approved):

packages/pages/src/generate/features/features.ts
mkilpatrick: enums are advised against in Typescript for various reasons. Union types as I have here are preferred.

PR #393, opened 2023-09-20
feat(plugin): move urlWriteback into pagesJS

asanehisa:
Originally urlWriteback was only exported to deno, but now it's available in pagesJS as well.


asanehisa:
Should we delete the deno urlWriteback?

mkilpatrick:
> Should we delete the deno urlWriteback?

No. We need to leave it for backwards compatibility.

mkilpatrick (approved):

PR #395, opened 2023-09-22
feat(dev): add upgrade command

jwartofsky-yext:
This adds a new upgrade command which can be called in a repo with 'npx pages upgrade'.

It first runs everything from the upgrader script here: https://github.com/lambdaFun94/pages-updater

It then migrates site-stream, serving, and sitemap to the new config.yaml. It then clears everything in sites-config except for redirects.csv (though this may change).

Example config.yaml when run with this [sites-config](https://drive.google.com/drive/folders/1x7-oIyvX_ZbaCI-PRkrDflV4sCCMFufo?usp=sharing):
```
buildConfiguration:
  buildCommand: npm run build:local
  installDependenciesStep:
    command: npm install
    requiredFiles:
      - package.json
      - package-lock.json
      - .npmrc
livePreviewConfiguration:
  setupCommand: ":"
siteStream:
  id: site-stream
  filter:
    entityIds:
      - site-entity
  source: knowledgeGraph
  fields:
    - name
    - logo
  localization:
    locales:
      - en
    primary: false
reverseProxy:
  displayUrlPrefix: www.[brand].com/[subdirectory]
sitemap:
  disableSitemapGeneration: false
  fileName: custom-sitemap-name.xml
  exclude_list:
    - feedback/
    - robots.txt
authorization:
  policyName: example-yext-policy


```


packages/pages/package.json
```diff
@@ -81,7 +81,8 @@
     "ts-morph": "^19.0.0",
     "typescript": "^5.1.6",
     "vite": "^4.4.9",
-    "vite-plugin-node-polyfills": "^0.11.0"
+    "vite-plugin-node-polyfills": "^0.11.0",
+    "yaml": "^2.3.1"
```
mkilpatrick, line 85: js-yaml is already imported. Do we need this one too?

packages/pages/src/upgrade/pages_updater.ts
```diff
@@ -0,0 +1,83 @@
...
+import path from "path";
+
+const pagesImportRegex = /"@yext\/pages\/components"/g;
+const replacementText = '"@yext/sites-components"';
+
+// Function to install npm packages
+const installPackages = (targetDirectory: string) => {
+  try {
+    execSync(
+      `npm install @yext/sites-components@latest @vitejs/plugin-react@4.0.4 vite@latest --prefix ${targetDirectory}`,
```
mkilpatrick, line 12: We can't assume they are using npm. We should really add it to the package.json manually.

mkilpatrick, line 12: We need to pin the specific Vite version because Vite 5 is coming out soon. See the updater script. There was a change for this recently.

packages/pages/src/upgrade/pages_updater.ts
```diff
@@ -0,0 +1,83 @@
...
+// Function to update package.json scripts
+const updatePackageScripts = (targetDirectory: string) => {
+  const packageJsonPath = path.resolve(targetDirectory, "package.json");
+
+  try {
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
+    packageJson.scripts = {
+      dev: "pages dev",
+      prod: "pages prod",
+      "build:local": "pages build",
```
mkilpatrick, line 69: Why is this one quoted but dev and prod aren't?

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -0,0 +1,74 @@
+import path from "path";
+import fs from "fs";
+import yaml from "yaml";
+
+const migrateServing = async (
```
mkilpatrick, line 5: Can you order these functions in the file in the same order they're executed?

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -0,0 +1,74 @@
...
+  configYamlPath: string,
+  sitesConfigPath: string
+) => {
+  const servingPath = path.resolve(sitesConfigPath, "serving.json");
+  if (fs.existsSync(servingPath)) {
+    const servingJson = JSON.parse(fs.readFileSync(servingPath, "utf-8"));
+    const displayUrlPrefix = servingJson.displayUrlPrefix;
+    if (displayUrlPrefix !== null) {
+      console.info(
+        "migrating reverse proxy info from serving.json to config.yaml"
```
mkilpatrick, line 15: Some of your logs begin with uppercase and some don't. Be consistent.

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -0,0 +1,74 @@
...
+    return;
+  }
+  const configYamlPath = path.resolve(target, "config.yaml");
+  if (!fs.existsSync(configYamlPath)) {
+    console.info("config.yaml does not exist, cannot migrate");
+    return;
+  }
+  await migrateGlobalData(configYamlPath, sitesConfigPath);
+  await migrateSiteMap(configYamlPath, sitesConfigPath);
+  await migrateServing(configYamlPath, sitesConfigPath);
```
mkilpatrick, line 73: We also need to run the features -> templates and ci -> artifacts code.

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -0,0 +1,18 @@
+import { Command } from "commander";
+import path from "path";
+import { updatePages } from "./pages_updater.js";
+import { migrateConfigs } from "./migrate_config.js";
+
+const handler = async () => {
+  // Pass CLI arguments as env variables to use in vite-plugin
```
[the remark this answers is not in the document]
jwartofsky-yext: There's no cli arguments. I'll delete that

packages/pages/src/upgrade/pages_updater.ts
jwartofsky-yext: It may have been because of the colon. I'll try removing the quotes and see if it works

packages/pages/src/upgrade/pages_updater.ts
```diff
@@ -0,0 +1,83 @@
...
+    const filePath = path.join(currentPath, file);
+    const isDirectory = fs.statSync(filePath).isDirectory();
+    const isGitDirectory = path.basename(filePath) === ".git";
+    const isNodeModulesDirectory = path.basename(filePath) === "node_modules";
+    const fileExtension = path.extname(filePath);
+
+    if (isDirectory && !isGitDirectory && !isNodeModulesDirectory) {
+      processDirectoryRecursively(filePath);
+    } else if (
+      [".js", ".ts", ".tsx"].includes(fileExtension) &&
```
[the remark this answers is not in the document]
jwartofsky-yext: It may be. I copied this from the upgrader script. I'll change to use src/ only

packages/pages/package.json
jwartofsky-yext: I found this one easier to work with. It's also being updated more frequently. Last update 1 month ago vs 2 years for js-yaml

packages/pages/src/upgrade/pages_updater.ts
jwartofsky-yext: It is the colon. I gave them all quotes so that it is consistent.

packages/pages/package.json
mkilpatrick: Ah ok. Can you remove this one then and update my usage of it? Should be easy to swap out.

packages/pages/src/dev/dev.ts
```diff
@@ -25,6 +26,12 @@ const handler = async ({
   noGenFeatures,
   port,
 }: DevArgs) => {
+  if (!fs.existsSync("config.yaml")) {
+    console.warn(
+      "config.yaml not found. Please run 'npx pages upgrade' to create it."
```
mkilpatrick, line 31: Can you check what the PMs want the warning message to be? I could see them wanting something more like "It looks like you're using an older setup. Please run 'npx pages upgrade' to upgrade to the latest format."

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -2,58 +2,116 @@ import path from "path";
 import fs from "fs";
 import yaml from "yaml";
 
-const migrateServing = async (
+function readJsonSync(jsonPath: string): any {
```
mkilpatrick, line 5: Use `const readJsonSync = () => {` way

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -65,10 +123,14 @@ export const migrateConfigs = async (target: string) => {
...
   const configYamlPath = path.resolve(target, "config.yaml");
   if (!fs.existsSync(configYamlPath)) {
-    console.info("config.yaml does not exist, cannot migrate");
-    return;
+    console.info("config.yaml does not exist, creating it");
+    fs.writeFileSync(configYamlPath, "");
   }
-  await migrateGlobalData(configYamlPath, sitesConfigPath);
-  await migrateSiteMap(configYamlPath, sitesConfigPath);
+  await migrateCiJson(configYamlPath, sitesConfigPath);
```
mkilpatrick, line 129: sitesConfigPath defined from line 119 needs to come from the projectStructure. Same for any other files.

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -2,58 +2,116 @@ import path from "path";
...
-      fs.readFileSync(configYamlPath, "utf-8")
+    writeYamlSync(configYamlPath, "globalData", sitesJson);
+  }
+};
+
+const migrateRedirects = async (
+  configYamlPath: string,
+  sitesConfigPath: string
+) => {
+  const redirectsPath = path.resolve(sitesConfigPath, "redirects.csv");
```
mkilpatrick, line 65: We'll need to check with the PMs and Joe about keeping redirects.csv. This file can be large so it doesn't really make sense to jam it all into the yaml. The second part of the question is if this file still needs to live in sites-config or does it now live at the root?

Or even more, is the file path referenced in the config.yaml?

mkilpatrick, line 65: You didn't add a console.info for this one.

packages/pages/src/upgrade/pages_updater.ts
```diff
@@ -7,15 +6,19 @@ const replacementText = '"@yext/sites-components"';
...
+  }
   try {
-    execSync(
-      `npm install @yext/sites-components@latest @vitejs/plugin-react@4.0.4 vite@latest --prefix ${targetDirectory}`,
-      {
-        stdio: "inherit",
-      }
-    );
+    const packageJson = JSON.parse(fs.readFileSync(packagePath, "utf-8"));
+    packageJson.dependencies["@yext/sites-components"] = "^1.0.0-rc.4";
```
mkilpatrick, line 16: Is there a way to get the latest for this one instead of hardcoding?

packages/pages/src/upgrade/pages_updater.ts
```diff
@@ -75,9 +78,13 @@ const updatePackageScripts = (targetDirectory: string) => {
...
+/**
+ * @param target the root directory to run on
+ * Install packages, recursively process imports (excluding .git and node_modules directories),
+ * and update package.json scripts in the specified directory
+ */
 export const updatePages = async (target: string) => {
-  // Install packages, recursively process imports (excluding .git and node_modules directories), and update package.json scripts in the specified directory
   installPackages(target);
-  processDirectoryRecursively(target);
+  processDirectoryRecursively(path.resolve(target, "src"));
```
mkilpatrick, line 88: Use ProjectStructure

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -2,12 +2,14 @@ import { Command } from "commander";
...
 import { updatePages } from "./pages_updater.js";
 import { migrateConfigs } from "./migrate_config.js";
+import { execSync } from "child_process";
 
 const handler = async () => {
-  // Pass CLI arguments as env variables to use in vite-plugin
   const rootPath = path.resolve("");
   await updatePages(rootPath);
   await migrateConfigs(rootPath);
+  execSync("pages generate templates");
```
mkilpatrick, line 11: Hmm, it's probably more correct to call the functions rather than execute the shell script.

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -0,0 +1,20 @@
...
+import { migrateConfigs } from "./migrate_config.js";
+import { execSync } from "child_process";
+
+const handler = async () => {
+  const rootPath = path.resolve("");
+  await updatePages(rootPath);
+  await migrateConfigs(rootPath);
+  execSync("pages generate templates");
+  execSync("pages generate artifacts");
+};
```
mkilpatrick, line 13: Please test what happens after you run your script and then try to `npm run prod`. I'm pretty sure there might be errors due to have config.yaml and the older sites-config files (features.json). If there is an error, tell Spruce. Whether there's an error or not, we should clean up (delete) the old files/folders.

packages/pages/src/upgrade/pages_updater.ts
jwartofsky-yext: I think caret is the best we can do without calling npm or pnpm
https://github.com/npm/node-semver#versions

packages/pages/src/upgrade/upgrade.ts
jwartofsky-yext: npm run dev results in this error when opening a page:
`Error: Multiple documents found for locale: "en"`
image

npm run prod results in this:
```
pages-starter-react-locations/node_modules/@yext/pages/dist/vite-plugin/build/closeBundle/closeBundle.js:53
Error: TypeError [ERR_INVALID_ARG_TYPE]: The "paths[1]" argument must be of type string. Received undefined
code: 'PLUGIN_ERROR',
  plugin: 'vite-plugin:build',
  hook: 'closeBundle'
```

Note that if I remove config.yaml, both npm run dev and npm run prod work as expected

jwartofsky-yext: Perhaps we should give users the option to run the upgrade and migration separately?

Unless we're planning to drop support for sites-config soon

packages/pages/src/upgrade/pages_updater.ts
jwartofsky-yext: This package works: https://www.npmjs.com/package/latest-version

packages/pages/src/upgrade/upgrade.ts
jwartofsky-yext: I fixed the prod error. It was caused by checking for null or empty instead of checking truthy.

However the multiple documents for locale error remains

packages/pages/src/dev/dev.ts
```diff
@@ -28,7 +28,9 @@ const handler = async ({
 }: DevArgs) => {
   if (!fs.existsSync("config.yaml")) {
     console.warn(
-      "config.yaml not found. Please run 'npx pages upgrade' to create it."
+      "It looks like you’re using an older setup of a Pages repo. Please run ‘npx pages upgrade’" +
+        " to upgrade to the latest format. This will setup a new configuration file config.yaml" +
+        " and install some new required dependencies."
```
mkilpatrick, line 33: It doesn't actually run install though. Did this come from the PMs?

packages/pages/src/upgrade/clean_configs.ts
```diff
@@ -0,0 +1,29 @@
+import { ProjectStructure } from "../common/src/project/structure.js";
+import fs from "fs";
+import path from "path";
+
+/**
+ * cleans up all config files under sites-config except for redirects.csv
```
mkilpatrick, line 6: We should copy redirects.csv to the root (or scoped folder) and then we can just delete the whole [scoped] folder recursively.

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -12,33 +13,25 @@ function readJsonSync(jsonPath: string): any {
...
-const migrateCiJson = async (
-  configYamlPath: string,
-  sitesConfigPath: string
-) => {
-  const ciJson = readJsonSync(path.resolve(sitesConfigPath, "ci.json"));
+const migrateCiJson = async (configYamlPath: string, ciPath: string) => {
+  const ciJson = readJsonSync(ciPath);
   if (ciJson !== null) {
-    ciJson.artifactStructure = null;
+    ciJson.artifactStructure = undefined;
```
mkilpatrick, line 27: This isn't even a key. Just delete this line.

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -1,20 +1,28 @@
...
-  await updatePages(rootPath);
-  await migrateConfigs(rootPath);
-  execSync("pages generate templates");
-  execSync("pages generate artifacts");
+const handler = async ({ scope }: { scope: string }) => {
+  const scoped = { scope: scope };
+  const projectStructure = await ProjectStructure.init(scoped);
+  await updatePages(projectStructure);
+  await migrateConfigs(projectStructure);
+  await templatesHandler(scoped);
```
mkilpatrick, line 14: Can you not pass scope directly through?

packages/pages/src/upgrade/clean_configs.ts
```diff
@@ -0,0 +1,29 @@
+import { ProjectStructure } from "../common/src/project/structure.js";
```
mkilpatrick, line 1: Use snakeCase for these filenames. No other files in this project use underscores.

packages/pages/src/upgrade/migrate_config.ts
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Migrates configuration files from sites-config to config.yaml
+ * @param projectStructure the structure of the project
+ */
+export const migrateConfigs = async (projectStructure: ProjectStructure) => {
+  const sitesConfigPath = projectStructure
+    .getSitesConfigPath()
+    .getAbsolutePath();
+  const configYamlPath = path.resolve(projectStructure.config.rootFiles.config);
```
mkilpatrick, line 107: We'll need to incorporate the scope when creating the config.yaml. I had a meeting today and we decided that we will use folder scopes. sites-config/ is already set up like this so we just need to move the things in the same way.

```
brandone.com/
    config.yaml
brandtwo.com/
   config.yaml
config.yaml -- this is non-scoped 
```

packages/pages/src/upgrade/upgrade.ts
jwartofsky-yext: We could, but they would all be init-ing the project structure on the other end anyways

packages/pages/src/dev/dev.ts
jwartofsky-yext: Yes

packages/pages/src/upgrade/pages_updater.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// Function to replace imports in a file
+const replaceImports = (filePath: string) => {
+  try {
+    const fileContent = fs.readFileSync(filePath, "utf8");
+    const modifiedContent = fileContent.replace(
+      pagesImportRegex,
+      replacementText
+    );
+    fs.writeFileSync(filePath, modifiedContent, "utf8"); // Update the file content
```
jwartofsky-yext, line 60: Add legacy Markdown warning

packages/pages/src/upgrade/upgrade.ts
mkilpatrick: I meant can you pass scope instead of scoped

jwartofsky-yext: oh it doesn't accept that

It expects {scope: scope}, so I could write that multiple times or pass the object through

packages/pages/src/upgrade/migrate_config.ts
jwartofsky-yext: This is actually necessary. Without it it will copy artifactStructure from ci.json

packages/pages/src/upgrade/migrate_config.ts
jwartofsky-yext: If there a multiple scopes like so:
sites-config:
    ..configStuff
    tacos: ..configStuff

And a user runs npx pages upgrade without a scope, should tacos get deleted? The user may still need to migrate it.

Or should upgrade not accept a scope and simply migrate everything?

I think the latter

mkilpatrick: Good point. We should just do everything during the one-time upgrade command and move everything, including all scopes.

packages/pages/src/upgrade/migrate_config.ts
jwartofsky-yext: Current implementation is to move redirects.csv next to config.yaml

packages/pages/src/upgrade/upgrade.ts
jwartofsky-yext: Cleaned this up a bit. We pass {scope: scope} through

packages/pages/src/upgrade/migrate_config.ts
jwartofsky-yext: Done both

If you give it a scope, it migrates everything from that scope and down and doesn't delete sites-config for the unmigrated scopes.

If you don't provide a scope (or just root), it simply migrates everything.

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -1,19 +1,14 @@
...
-  const scoped = { scope: scope };
-  const projectStructure = await ProjectStructure.init(scoped);
-  await updatePages(projectStructure);
-  await migrateConfigs(projectStructure);
-  await templatesHandler(scoped);
-  await artifactsHandler(scoped);
-  await cleanConfigs(projectStructure);
+  await updatePages(scope);
+  await migrateConfigs(scope);
+  await templatesHandler({ scope: scope });
```
mkilpatrick, line 10: It's weird that the functions take in different things, an object that contains a scope or just scope directly. Also, it's preferable to just init ProjectStructure once like you had before instead of in each handler.

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -82,16 +91,51 @@ const updatePackageScripts = (targetDirectory: string) => {
   }
 };
 
+//Function to install dependencies
```
mkilpatrick, line 94: space before `Function`

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -82,16 +91,51 @@ const updatePackageScripts = (targetDirectory: string) => {
...
+const installDependencies = async (
```
mkilpatrick, line 95: Neat!

packages/pages/src/upgrade/upgrade.ts
jwartofsky-yext: I think the weird scope arguments are an artifact of how Commander handles option arguments.

I had changed it because we have to init projectStructure multiple times during the recursive migrate, but I changed  upgrade.ts it to use less inits().

mkilpatrick (approved):

PR #396, opened 2023-09-27
revert(dev): use esbuild instead of ts-morph to parse template configs

mkilpatrick:
This is a manual revert to use esbuild again instead of ts-morph because the latter doesn't compile anything. This led to issues when a template used a function inside of its exported config.

The change for esbuild now is that there is a fallback loader to load any unknown filetype using the `file` esbuild Loader.


packages/pages/src/vite-plugin/build/closeBundle/manifest.ts
```diff
@@ -1,11 +1,11 @@
 import fs from "fs-extra";
 import path from "path";
 import { ProjectStructure } from "../../../common/src/project/structure.js";
-import { TemplateModuleCollection } from "../../../common/src/template/internal/loader.js";
 import { convertToPosixPath } from "../../../common/src/template/paths.js";
 import { Manifest } from "../../../common/src/template/types.js";
 import { glob } from "glob";
 import { Path } from "../../../common/src/project/path.js";
+import { TemplateModuleCollection } from "../../../common/src/template/loader/loader.js";
```
jwartofsky-yext, line 8: Why all these import shuffling changes?

mkilpatrick: Not sure. Probably just me removing and adding things back. There's no format for actual import ordering.

PR #397, opened 2023-09-29
feat(dev): add scope to yextrc

asanehisa:
.yextrc file can store non-scoped accountId / universe info as well as scoped info.


packages/pages/src/dev/server/autoInit.ts
```diff
@@ -45,10 +48,21 @@ Let's connect to your Yext Account
...
         process.exit(1);
       });
-    fs.writeFileSync(
-      ".yextrc",
-      `accountId: ${accountId}\nuniverse: ${universe}`
-    );
+    if (scope) {
+      fs.appendFileSync(
+        ".yextrc",
+        `"${scope}":\n` +
```
jwartofsky-yext, line 54: alternatively you could also use a JSON object then parse to string instead of writing in the colons and \n characters

jwartofsky-yext (dismissed):
lgtm

packages/pages/src/dev/server/autoInit.ts
asanehisa: I could! I could just leave it as json actually (i think json is a valid format in yaml), only thing I can't add is the comment I think

packages/pages/src/dev/server/autoInit.ts
```diff
@@ -5,17 +5,19 @@ import chalk from "chalk";
 import { parseYextrcContents } from "../../util/yextrcContents.js";
 import { logErrorAndExit } from "../../util/logError.js";
 
-export const autoYextInit = async () => {
+export const autoYextInit = async (scope: string | undefined) => {
   if (!fs.existsSync(".yextrc")) {
-    await autoYextInitWithoutYextrc();
+    await autoYextInitUpdateYextrc(scope);
```
mkilpatrick, line 10: I think I prefer the previous name. Though really we should change the flow. If the file doesn't exist then we create it. After that we try to retrieve the data. If it's not there, then we do the prompt and add it.

packages/pages/src/util/yextrcContents.ts
```diff
@@ -11,21 +11,34 @@ const validUniverses = ["sandbox", "production", "sbx", "prod", "qa", "dev"];
...
       accountId = parsedContents.accountId;
       universe = parsedContents.universe;
+    } else {
+      shouldUpdate = true;
     }
   } catch (e: any) {
     // Return undefined for both fields if .yextrc cannot be read
   }
-  return { accountId, universe };
+  return { accountId, universe, shouldUpdate };
```
mkilpatrick, line 43: I don't think this should return an update status. It should only return the accountId and universe given a scope. If those come back as undefined then we can add them.

asanehisa: Removed.

packages/pages/src/dev/server/autoInit.ts
asanehisa: Adjusted flow.

packages/pages/src/dev/server/autoInit.ts
```diff
@@ -29,7 +26,12 @@ const autoYextInitWithYextrc = async () => {
   }
 };
 
-const autoYextInitWithoutYextrc = async () => {
+const createYextrc = async (scope: string | undefined) => {
```
mkilpatrick, line 29: This doesn't need to be async if you're not calling any async functions.

packages/pages/src/dev/server/autoInit.ts
```diff
@@ -26,7 +26,7 @@ const autoYextInitWithYextrc = async (scope: string | undefined) => {
   }
 };
 
-const createYextrc = async (scope: string | undefined) => {
+const createYextrc = () => {
   fs.writeFileSync(".yextrc", "");
```
mkilpatrick, line 30: This function just calls another function and only in one place. Call fs.writeFileSync directly on line 10.

mkilpatrick (approved):

jwartofsky-yext (approved):

PR #399, opened 2023-10-03
fix(pages-plugins): slugManager searchIds not working on webhook update

mtusman:
Also refactor slugFormat type.

J=CR-3203

### DIFF
--- a/packages/plugins/slugManager/manager.ts
+++ b/packages/plugins/slugManager/manager.ts
@@ -6,20 +6,22 @@ interface ProfileUpdate {
   [key: string]: string | boolean | Meta;
 }
 
-export type SlugFormatType =
-  | string
-  | {
-      generatorFn: (lang: string, profile: BaseEntity) => string;
-      fields?: string[];
-    };
+interface SlugFormatString {
+  slugFormat: string;
+  fields?: never;
+}
 
-export interface InternalSlugManagerConfig {
-  slugFormat: SlugFormatType;
+interface SlugFormatFunc {
+  slugFormat: (lang: string, profile: BaseEntity) => string;
+  fields: string[];
+}
+
+export type InternalSlugManagerConfig = {
   searchIds?: string[];
   entityTypes?: string[];
   slugField?: string;
   api: Pick<IAPI, "updateField" | "listLanguageProfiles" | "listEntities">;
-}
+} & (SlugFormatString | SlugFormatFunc);
 
 export function createManager(config: InternalSlugManagerConfig) {
   const {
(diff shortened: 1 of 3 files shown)


packages/plugins/slugManager/manager.ts
```diff
@@ -6,41 +6,41 @@ interface ProfileUpdate {
   [key: string]: string | boolean | Meta;
 }
 
+export type SlugFormatType =
+  | string
+  | {
+      generatorFn: (lang: string, profile: BaseEntity) => string;
+      fields?: string[];
```
bhainesva, line 13: I think this should be required if someone's using a generatorFn, even if that means they have to explicitly provide an empty array if they're not using any fields.

packages/plugins/slugManager/manager.ts
```diff
@@ -6,41 +6,41 @@ interface ProfileUpdate {
   [key: string]: string | boolean | Meta;
 }
 
+export type SlugFormatType =
```
bhainesva, line 9: I have a slight preference for typing the config like: 

```js
{
  slugFormat: string | (lang, profile) => string;
  fields?: string[];
}
```

I like that it removes the nested object and reduces the total number of options (don't have to remember generatorFn vs. slugFormat anymore). But was there a reason we might prefer this way instead?

packages/plugins/slugManager/manager.ts
```diff
@@ -54,7 +54,7 @@ export function createManager(config: InternalSlugManagerConfig) {
     const entitiesResponse = await api.listEntities(params);
 
     const profileParams = new URLSearchParams({
-      fields: ["meta", slugField, ...(slugGeneratorFnFields || [])].join(","),
+      fields: ["meta", slugField, ...slugGeneratorFnFields].join(","),
```
bhainesva, line 57: It was already there, but wondering if there's a reason we always request `slugField` here?

packages/plugins/slugManager/manager.ts
```diff
@@ -85,25 +83,33 @@ export function createManager(config: InternalSlugManagerConfig) {
...
+    const profile =
+      lang == event.primaryProfile.meta.language
+        ? event.primaryProfile
+        : event.languageProfiles.find(
+            (profile) => profile.meta.language == lang
+          );
+    const slug =
+      typeof slugFormat == "string"
+        ? slugFormat
+        : slugFormat.generatorFn(lang, profile!);
```
bhainesva, line 112: instead of `profile!` I think we could throw an error here, that way hopefully whatever shows up in the platform logs is more explicit about what went wrong

packages/plugins/slugManager/manager.ts
mtusman: I think it was added because `slug` is included as an optional value for the `BaseEntity` interface, but I think we should remove since its possible that slugField isn't necessarily always equal to "slug".

bhainesva (approved):

mkilpatrick (approved):

PR #400, opened 2023-10-04
feat(dev): add noMigration flag for upgrade

jwartofsky-yext:
Adds a --noMigration flag for the upgrade script. With noMigration, upgrade will skip the migration to config.yaml and deletion of sites-config.

Tested manually on a local repo. Migration is skipped with the flag and still occurs when the flag is missing.


asanehisa (approved):
lgtm

PR #401, opened 2023-10-04
feat(dev): set siteInternalHostName locally

asanehisa:
siteInternalHostName is manually set to scope

### DIFF
--- a/packages/pages/src/dev/server/ssr/getLocalData.ts
+++ b/packages/pages/src/dev/server/ssr/getLocalData.ts
@@ -126,8 +126,6 @@ const getLocalData = async (
           )
           .toString()
       );
-      data.siteInternalHostName = process.env.YEXT_PAGES_SCOPE;
-
       if (criterion(data)) {
         return data;
       }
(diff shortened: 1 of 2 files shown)


packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -126,6 +126,8 @@ const getLocalData = async (
           )
           .toString()
       );
+      data.siteInternalHostName = process.env.YEXT_PAGES_SCOPE;
```
mkilpatrick, line 129: I don't think this is where we want to add this. It doesn't take into account getting data dynamically. I think the better place is in propsLoader.ts. Also, is YEXT_PAGES_SCOPE always available on process.env? We might need to pass scope through to it, which is set on the projectStructure.

asanehisa: Moved to propsLoader.ts. YEXT_PAGES_SCOPE is set in packages/pages/src/build/build.ts so it should be available on process.env. Tested and it works with pages dev --scope "someScope".

mkilpatrick (approved):

PR #402, opened 2023-10-06
fix(plugin): remove import.meta.resolve dependency

mancusi:
Removes the dependency on import.meta.resolve which
    is only supported in the newest versions of Safari.
    
    It ultimately was unnecessary as pagesJS was providing
    absolute paths to the call which returns a URL to the module.
    These get resolved during an import in the same
    manner as an absolute path so it can just removed.
    
    Before:
    ```
    // url = "https://domain.com/abs/path"
    const url = import.meta.resolve("/abs/path")
    // import("https://domain.com/abs/path")
    import(url)
    ```
    After
    ```
    import(new URL("/abs/path", import.meta.url).toString())
    ```
    
    TEST=auto,manual
    
    ran unit tests. Manually tested on a RP site and a non-RP site.
    Since the asset paths are absolute, behavior is the same
    assuming the RP is properly forwarding assets.


mkilpatrick (approved):
I wonder why it only failed for Windows.

jwartofsky-yext:
I think the issue may be that import.meta.resolve was converting to \ which was working on Windows, but this change undoes that.

If you use path.resolve it should convert to Windows format

mancusi:
> I think the issue may be that import.meta.resolve was converting to \ which was working on Windows, but this change undoes that.
> 
> If you use path.resolve it should convert to Windows format

The only thing ECMAScript specifies for import specifiers are relative and absolute URLs. Since this is neither it may have been a case of undefined behavior (although it worked in the set of browsers I tested on). Instead of relying on undefined behavior I updated the PR in my latest commit to use import.meta.url (supported in older Safari version) to generate an absolute URL. Hopefully this allows tests to pass

jwartofsky-yext:
> > I think the issue may be that import.meta.resolve was converting to \ which was working on Windows, but this change undoes that.
> > If you use path.resolve it should convert to Windows format
> 
> The only thing ECMAScript specifies for import specifiers are relative and absolute URLs. Since this is neither it may have been a case of undefined behavior (although it worked in the set of browsers I tested on). Instead of relying on undefined behavior I updated the PR in my latest commit to use import.meta.url (supported in older Safari version) to generate an absolute URL. Hopefully this allows tests to pass

Tests are still failing :(

I can play around with this PR using the Windows laptop tomorrow and find out what is going wrong

mancusi:
> > > I think the issue may be that import.meta.resolve was converting to \ which was working on Windows, but this change undoes that.
> > > If you use path.resolve it should convert to Windows format
> > 
> > 
> > The only thing ECMAScript specifies for import specifiers are relative and absolute URLs. Since this is neither it may have been a case of undefined behavior (although it worked in the set of browsers I tested on). Instead of relying on undefined behavior I updated the PR in my latest commit to use import.meta.url (supported in older Safari version) to generate an absolute URL. Hopefully this allows tests to pass
> 
> Tests are still failing :(
> 
> I can play around with this PR using the Windows laptop tomorrow and find out what is going wrong

Just reverted all of these changes and it's still failing so I don't think it has anything to do with my changes.

jwartofsky-yext:
This is based on rc.2, which is missing the windows fixes. You should ignore the windows tests failures if we're making a rc.2.1

jwartofsky-yext (approved):
lgtm

PR #403, opened 2023-10-09
fix(plugin): remove import meta resolve

mancusi:
Removes the dependency on import.meta.resolve which
is only supported in the newest versions of Safari.

It ultimately was unnecessary as pagesJS was providing
absolute paths to the call which returns a URL to the module.
These get resolved during an import in the same
manner as an absolute path so it can just removed.

Before:
```
// url = "https://domain.com/abs/path"
const url = import.meta.resolve("/abs/path")
// import("https://domain.com/abs/path")
import(url)
```
After
```
import(new URL("/abs/path", import.meta.url).toString())
```

TEST=auto,manual

ran unit tests. Manually tested on a RP site and a non-RP site.
Since the asset paths are absolute, behavior is the same
assuming the RP is properly forwarding assets.


packages/pages/src/common/src/template/hydration.ts
```diff
@@ -82,6 +82,13 @@ export const getHydrationTemplate = (
     `;
 };
 
+const makeAbsolute = (path: string): string => {
```
jwartofsky-yext, line 85: path.resolve should turn this into an absolute path

jwartofsky-yext (approved):

mkilpatrick (approved):

packages/pages/src/common/src/template/hydration.ts
mancusi: These are url paths not system dependent paths and url paths all use forward-slashes.

jwartofsky-yext: You can path.resolve(convertToPosixPath(path)) too, but this works aswell

PR #405, opened 2023-10-11
Added jstest workflow - WIP

jwartofsky-yext:


.github/workflows/jstest.yml
```diff
@@ -0,0 +1,42 @@
...
+          node-version: 18
+          cache: "pnpm"
+
+      - name: Install deps
+        run: pnpm i
+
+      - name: Pack
+        run: pnpm pack --pack-destination latest
+
+      - name: Checkout JSTEST
```
mkilpatrick, line 33: This checks out jstest, but how does it run the tests of each package using the local packed version of pagesjs?

jwartofsky-yext: It doesn't yet, still WIP

packages/pages/src/dev/server/server.ts
```diff
@@ -51,7 +51,7 @@ export const createServer = async (
       projectStructure.config.envVarConfig.envVarPrefix
     ),
     optimizeDeps: {
-      include: ["react-dom", "react-dom/client"],
```
jwartofsky-yext, line 54: Will undo this

.github/workflows/jstest.yml
jwartofsky-yext: It installs and attempts to run playwright in all the packages now, but am running into some webserver errors that don't occur locally: https://github.com/yext/pages/actions/runs/6527694680/job/17722896253?pr=405

Debugging..

packages/pages/scripts/install_and_test.js
```diff
@@ -0,0 +1,24 @@
+import path from "path";
+import fs from "fs";
+import * as child_process from "child_process";
+
+// Gets the path to the packaged version of PagesJS
+const latestPath = path.resolve("latest");
```
mkilpatrick, line 6: Unused

packages/pages/scripts/install_and_test.js
```diff
@@ -0,0 +1,24 @@
...
+const packagePath = path.resolve(
+  "latest",
```
mkilpatrick, line 8: What is latest?

packages/pages/scripts/install_and_test.js
```diff
@@ -0,0 +1,24 @@
...
+  fs.readdirSync(latestPath.toString())[0]
+);
+
+// Iterate through sdks, install local pages package, and run playwright
+const sdks = fs.readdirSync(`../../jstest/sdks`);
```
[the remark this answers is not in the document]
jwartofsky-yext: This is currently only running on Ubuntu. Do we want to run on Windows aswell?

packages/pages/scripts/install_and_test.js
jwartofsky-yext: It's the directory that pages gets packed into

.github/workflows/jstest.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          ref: "main"
+          repository: yext/jstest
+          ssh-key: ${{ secrets.JSTEST_PRIVATE_SSH_KEY }}
+          path: jstest
+
+      - name: Test chat-ui-react
+        run: | 
+         cd $GITHUB_WORKSPACE/jstest/sdks/chat-ui-react \
+          && npm i && npx playwright install \
```
mkilpatrick, line 44: It playwright install necessary? I'd think `npm i` would cover it if the tests in the folder require it as it'd be an npm dep.

.github/workflows/jstest.yml
```diff
@@ -0,0 +1,53 @@
...
+          && npm install @yext/pages $GITHUB_WORKSPACE/packages/pages/latest/* \
+          && npm run test
+
+      - name: Test search-ui-react
```
mkilpatrick, line 48: Is there a way we can just loop through all folders rather than specifying each one manually?

scripts/install_and_test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+cd $1/jstest/sdks
+
+for D in ./*; do
+    if [ -d "$D" ]; then
+        cd "$D"
+        npm i && npx playwright install \
+          && npm install @yext/pages $1/packages/pages/latest/* \
```
mkilpatrick, line 8: Re-commenting this: Is playwright install necessary? I'd think npm i would cover it if the tests in the folder require it as it'd be an npm dep.

jwartofsky-yext: Playwright install is separate from node package.json and doesn't get included in `npm i`. `Npm i` does install playwright, but the playwright install command installs the browsers and has to be run separately.

mkilpatrick (approved):

PR #408, opened 2023-10-16
fix(dev): prevent duplicate stream definitions

briantstephan:
This prevents shared/duplicate stream definitions from appearing multiple times in features.json. If conflicting stream definitions exist for the same stream ID, an error is thrown. Verified that duplicate identical stream definitions are ignored and duplicate but differing stream definitions cause an error.


packages/pages/src/generate/templates/createTemplatesJson.ts
```diff
@@ -86,9 +87,20 @@ export const getTemplatesConfig = (
...
+      (stream) => stream.$id === streamConfig.$id
+    );
+    if (!matchingStreamConfig) {
       streams.push(streamConfig);
+      continue;
+    }
+    if (isEqual(matchingStreamConfig, streamConfig)) {
+      continue;
     }
+    throw `Conflicting configurations found for stream ID: ${streamConfig.$id}`;
```
mkilpatrick, line 103: throw new Error('Conflicting....

briantstephan: We decided to keep it like this and use logErrorAndExit so that the full stack trace isn't included.

mkilpatrick (approved):

PR #413, opened 2023-10-17
feat(dev): better error message for CLI not installed

jwartofsky-yext:
We can change the message to ask the user to install packages once the CLI package change is ready


packages/pages/src/util/checkInstalled.ts
```diff
@@ -0,0 +1,17 @@
+import { execSync } from "child_process";
+
+function checkInstalled(command: string): void {
```
mkilpatrick, line 3: Can you add a test for this?

mkilpatrick (approved):

PR #414, opened 2023-10-17
feat(plugin): add ts-morph for client templates

asanehisa:


packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -0,0 +1,170 @@
...
+      .organizeImports()
+      .fixUnusedIdentifiers();
+    this.sourceFile.addStatements('import * as React from "react";');
+  }
+
+  getFileName(): string {
+    return this.sourceFile.getBaseName();
+  }
+
+  clearAll() {
```
mkilpatrick, line 160: Let's delete the file before running the parser, opposed to clearing the file out.

packages/pages/src/common/src/parsers/templateParser.ts
```diff
@@ -0,0 +1,32 @@
+import SourceFileParser, { createTsMorphProject } from "./sourceFileParser.js";
+
+/**
+ * TemplateParser is a class for parsing the template.
+ */
+export default class TemplateParser {
```
mkilpatrick, line 6: This name is a little generic. We're doing something specific here. Can you name it accordingly?

packages/pages/src/common/src/parsers/templateParser.ts
```diff
@@ -0,0 +1,32 @@
+import SourceFileParser, { createTsMorphProject } from "./sourceFileParser.js";
+
+/**
+ * TemplateParser is a class for parsing the template.
```
mkilpatrick, line 4: Add tests.

packages/pages/src/common/src/parsers/sourceFileParser.ts
asanehisa: Sounds good.

packages/pages/src/common/src/parsers/clientTemplateParser.ts
```diff
@@ -1,17 +1,16 @@
 import SourceFileParser, { createTsMorphProject } from "./sourceFileParser.js";
 
 /**
- * TemplateParser is a class for parsing the template.
+ * ClientTemplateParser is a class for parsing the client template.
  */
-export default class TemplateParser {
+export default class ClientTemplateParser {
```
mkilpatrick, line 6: TemplateParser was fine. This is literally parsing the src/template files. Those files aren't client or server-specific.

packages/pages/src/common/src/parsers/sourceFileParser.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+describe("getDefaultExport", () => {
+  it("correctly gets default export's name when function", () => {
+    const parser = createParser(
+      `export const no = false; export default function test() {}`
+    );
+    const defaultExport = parser.getDefaultExport();
+    expect(defaultExport).toBe("test");
+  });
+  it("correctly gets default export's name when variable", () => {
```
mkilpatrick, line 13: empty lines between tests

packages/pages/src/common/src/parsers/sourceFileParser.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    expect(imports[1].moduleSpecifier).toBe("@yext/pages");
+  });
+  it("handles file having no imports", () => {
+    const parser = createParser(`const test = 5;`);
+    const imports = parser.getAllImports();
+    expect(imports).toEqual([]);
+  });
+  it("handles multiple named imports from one path", () => {
+    const parser = createParser(
+      `import {Template, TemplateConfig, TemplateProps} from "@yext/pages";`
```
mkilpatrick, line 70: Does this work when there's a default export too?

`import Foo, {Template, TemplateConfig, TemplateProps} from "@yext/pages";`

packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -150,21 +158,25 @@ export default class SourceFileParser {
...
-      this.sourceFile.getPos(),
-      this.sourceFile.getEnd()
-    );
-  }
-
   async save() {
     await this.sourceFile.save();
   }
+
+  private getAbsolutePath = (base: string, relative: string): string => {
```
mkilpatrick, line 171: Can you add tsdoc to this explaining what it does with an example? Also add tests for it. If it's specific to imports you might want to name it accordingly as well.

mkilpatrick: Also, any reason this is a private function in the class and not just a const function outside of it (though in the same file)?

packages/pages/src/util/createTestSourceFile.ts
```diff
@@ -0,0 +1,21 @@
+import { Project } from "ts-morph";
+
+export default function createTestSourceFile(
```
mkilpatrick, line 3: tsdoc

packages/pages/src/common/src/parsers/clientTemplateParser.ts
```diff
@@ -0,0 +1,31 @@
+import SourceFileParser, { createTsMorphProject } from "./sourceFileParser.js";
+
+/**
+ * ClientTemplateParser is a class for parsing the client template.
+ */
+export default class ClientTemplateParser {
+  constructor(private originalSfp: SourceFileParser) {}
+
+  async makeClientTemplate(path: string) {
```
mkilpatrick, line 9: Add doc for that this does

packages/pages/src/common/src/parsers/clientTemplateParser.ts
```diff
@@ -0,0 +1,31 @@
...
+    const childExpressions =
+      this.originalSfp.getExpressionsByName(childExpressionNames);
+    const imports = this.originalSfp.getAllImports();
+
+    if (defaultExportName !== "") {
+      newSfp.addExpressions(childExpressions.filter((s) => s));
+      newSfp.addDefaultExport(defaultExportName);
+      newSfp.setAllImports(imports);
+    }
+    newSfp.save();
```
mkilpatrick, line 29: Where is the file saved? I think we'd need an output path for the user to control where it's output.

packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -0,0 +1,182 @@
...
+    private filepath: string,
+    project: Project
+  ) {
+    if (!project.getSourceFile(filepath)) {
+      project.addSourceFileAtPath(filepath);
+    }
+    this.sourceFile = project.getSourceFileOrThrow(filepath);
+  }
+
+  getFunctions() {
```
mkilpatrick, line 32: Add docs for all of these functions

packages/pages/src/common/src/parsers/clientTemplateParser.ts
asanehisa: So makeClientTemplate(path) takes in a path to the directory, assuming it is made already. 
In my follow-up PR, I'll have a function that makes a client dir and files. This will save those files after making ts-morph changes.

packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -1,3 +1,4 @@
+import path from "path";
```
mkilpatrick, line 1: Prefer the node prefix from node imports:
`import path from "node:path";`

packages/pages/src/common/src/parsers/templateParser.ts
```diff
@@ -0,0 +1,42 @@
+import fs from "fs";
```
mkilpatrick, line 1: node prefixes for fs and path

packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -164,19 +201,20 @@ export default class SourceFileParser {
...
-    return stack.join("/");
-  };
 }
+
+/**
+ * @param base the file path
+ * @param relative the relative file path
+ * @returns an absolute file path
+ */
+const getAbsolutePath = (base: string, relative: string): string => {
```
mkilpatrick, line 217: This function is only used in one place and is relatively simply. Just combine it into one line in the place it's used.

packages/pages/src/util/createTestSourceFile.ts
```diff
@@ -1,5 +1,10 @@
 import { Project } from "ts-morph";
 
+/**
+ * @param code to insert into test file
+ * @param filepath
```
mkilpatrick, line 5: Looks like this is a filename and not a path. Add description as well.

packages/pages/src/common/src/parsers/templateParser.ts
mkilpatrick: Similar to my other comment, it'd be ideal if makeClientTemplate didn't save the change automatically, but instead returned the results. This way we can add tests that use actual templates that include a bunch of other cruft a user could add so that we can verify that the client output is what we'd expect. Right now all of the tests for the other file are very limited with their input. Adding a test for this would give us more confidence that it does exactly what we want.

packages/pages/src/common/src/parsers/sourceFileParser.test.ts
```diff
@@ -0,0 +1,157 @@
+import { describe, it, expect } from "vitest";
+import path from "path";
```
mkilpatrick, line 2: prefix

packages/pages/src/util/createTestSourceFile.ts
asanehisa: Ah this is a filepath just only providing the filename in this case. createSourceFile requires a file path for first param. I think leaving it as filepath makes sense so this function can be used with filepath as a param?

mkilpatrick: sg

packages/pages/src/common/src/parsers/templateParser.ts
mkilpatrick: So are you going to do this in the next PR?

asanehisa: Oh hmmm I thought I did it here? By creating templateParser.test.ts and changing how templateParser's makeClientTemplate returns

mkilpatrick: Gotcha. Can you add some tests that use full template files too? That will help us feel more confident everything works when there's more than just one line input.

packages/pages/src/common/src/parsers/templateParser.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    );
+    const testParser = createParser(``);
+    const templateParser = new TemplateParser(
+      sourceParser
+    ).makeClientTemplateFromSfp(testParser);
+    expect(templateParser.fileContents).not.toContain(` mock tsdoc comments`);
+  });
+});
+
+describe("complex makeClientTemplate usages", () => {
```
asanehisa, line 88: Cannot directly compare expected results to a file due to the imports being absolute paths after being processed.

packages/pages/src/common/src/parsers/templateParser.test.ts
```diff
@@ -68,10 +68,83 @@ describe("makeClientTemplate", () => {
...
+  it("correctly returns expected file contents", () => {
+    const sourceParser = new SourceFileParser(
+      "./tests/fixtures/sourceFileTemplates/static.tsx",
+      createTsMorphProject()
+    );
+    const expectedParser = new SourceFileParser(
```
mkilpatrick, line 94: You shouldn't need to use the parser against the client. You want to check that the regular template (static.tsx) returns exactly what is in your client file by just reading in the client and comparing it.

packages/pages/src/common/src/parsers/templateParser.test.ts
```diff
@@ -68,10 +68,83 @@ describe("makeClientTemplate", () => {
...
+
+const expectedStaticImports = [
+  `import { Template, TemplateRenderProps } from "@yext/pages";`,
+  `import Banner from `,
+  `import Card from `,
+  `import PageLayout from `,
+  `import { ExternalImage } `,
+];
+
+const unnecessaryStaticStrings = [
```
mkilpatrick: I don't think having this is necessary. This will naturally be checked when you compare (input -> ts-morph) == client.

packages/pages/src/common/src/parsers/templateParser.test.ts
asanehisa: No longer true. Now uses relative paths

packages/pages/src/common/src/parsers/templateParser.test.ts
```diff
@@ -68,10 +68,83 @@ describe("makeClientTemplate", () => {
...
+      sourceParser
+    ).makeClientTemplateFromSfp(testParser);
+    expect(templateParser.fileContents).not.toContain(` mock tsdoc comments`);
+  });
+});
+
+describe("complex makeClientTemplate usages", () => {
+  it("correctly returns expected file contents", () => {
+    const sourceParser = new SourceFileParser(
+      "./tests/fixtures/sourceFileTemplates/static.tsx",
```
mkilpatrick, line 91: Add a test for an entity template as well that also uses a function in the config that comes from another file. Something like:

```
import { dedupe } from 'foo';
export const config = {
  .......
 fields: dedupe(["id", "id", "name"])
}
```

asanehisa: Sounds good. Does import { isProduction } from "@yext/pages/util"  work in this case? Just added the state.tsx template with isProduction being called in the default export.

packages/pages/src/common/src/parsers/templateParser.test.ts
```diff
@@ -91,21 +92,32 @@ describe("complex makeClientTemplate usages", () => {
...
       sourceParser
     ).makeClientTemplateFromSfp(testParser);
-    expectedStaticImports.forEach((expectedImport) => {
-      expect(templateParser.fileContents).toContain(expectedImport);
-    });
-    unnecessaryStaticStrings.forEach((unnecessaryString) => {
-      expect(templateParser.fileContents).not.toContain(unnecessaryString);
-    });
-    expect(templateParser.fileContents).toContain(expectedParser.getAllText());
+    console.log(templateParser.fileContents);
```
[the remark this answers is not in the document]
asanehisa: oops

packages/pages/tests/fixtures/sourceFileTemplates/clientState.tsx
```diff
@@ -6,6 +6,7 @@ import DirectoryStateGrid from "../components/DirectoryStateGrid";
 import EditTool from "../components/EditTool";
 import PageLayout from "../components/PageLayout";
 import "../index.css";
+import { dedupe } from "./functions";
```
mkilpatrick, line 9: I'd call this file util.ts (also note .ts and not .tsx since it's not a react component) to avoid confusion with serverless functions.

mkilpatrick (approved):

PR #417, opened 2023-10-19
feat(util): pass optional list of domains into isProduction

briantstephan:
This updates isProduction to take a variable number of production domains. If no domains are provided, it returns true when the current window's hostname is not localhost or preview.pagescdn.com. If one or more domains are provided, it returns true only when the current window's hostname matches at least one of the provided domains.


asanehisa (approved):
lgtm

mkilpatrick (approved):

Issue #415, opened 2023-10-18
Issues using document data in `getPath` for static features locally


bhainesva:
### Describe the bug

I'm trying to include locale in the url for one of my static features. I have a `getPath` function that looks like this:

```
export const getPath: GetPath<TemplateProps> = (data) => {
  console.log(data);
  return `${data?.document?.locale}/robots.txt`;
};
```

When I serve the site with `npx pages dev` the sample urls look like this:
![image](https://github.com/yext/pages/assets/6308571/edb79543-7a3b-40d5-a418-5e6edd4ab334)

If I remove the `?` from the `data?.document?.locale` it errors because document is undefined. If I click on one of the urls, based on the logs it looks like `getPath` is called three times. Once with an empty object, once with most of the data but without `relativePrefixToRoot`, and a third time with `relativePrefixToRoot` added.

### System Info

```shell
System:
    OS: macOS 12.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1.25 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.4.0 - ~/.nvm/versions/node/v17.4.0/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v17.4.0/bin/npm
    pnpm: 7.9.0 - ~/.nvm/versions/node/v17.4.0/bin/pnpm
  Browsers:
    Chrome: 118.0.5993.70
    Edge: 118.0.2088.46
    Safari: 15.4
  npmPackages:
    @yext/pages: 1.0.0-rc.5 => 1.0.0-rc.5
```

### Used Package Manager

npm

### Used Browser

Chrome

### Validations

- [x] Check that there isn't [already an issue](https://github.com/yext/pages/issues) that reports the same bug to avoid creating a duplicate.

PR #421, opened 2023-11-03
fix(dev): include document support for static page `getPath` functions.

alextaing:
Fixes https://github.com/yext/pages/issues/415. 

In the past, if users tried to access `document` fields in a static page `getPath` function, it would show up as "undefined" in the page URL. This PR addresses this by giving static pages access to the `document` object.

A couple changes were made to accommodate this functionality.  We want to pass in a document in order correctly compute a `getPath` function.  This means that when we iterate through templates to find the corresponding template document to our URL, (see `findMatchingStaticTemplate`), we need to retrieve the documents which is an asynchronous action.  Therefore, the `criterion` function in `findTemplateModuleInternal` was changed to accept async functions.

For clarity, `urlToFeature` was also refactored.  It was also found that `defaultLocale` was not being used in `serverRenderRoute`.  Before, we hard-coded `en` to be the default Locale in `urlToFeature`. But in this refactor, we use `defaultLocale` instead.

J=SUMO-5417
TEST=manual


mkilpatrick:
This is a "fix" vs a "feat". Can you update the commit message? Also mention "fixes [link to issue]" so it auto closes when this is merged.

Also does the document get injected during prod?

alextaing:
> This is a "fix" vs a "feat". Can you update the commit message? Also mention "fixes [link to issue]" so it auto closes when this is merged.
> 
> Also does the document get injected during prod?

Based on what I tested, when we create a `getPath` function that accesses the `document` object in a static page, then build and render the site using `yext pages build`, `yext pages render`, etc, the `getPath` functions correctly access the document fields.  This led me to believe that the document also gets injected during this flow.  Is this what you mean by prod, or is there another flow?
```
export const getPath: GetPath<ExternalImageData> = ({document}) => {
  return `${document.locale}.html`;
};
```

image

mkilpatrick (approved):

Issue #409, opened 2023-10-16
`vite build` errors with "Cannot find module"


bhainesva:
### Describe the bug

I'm running `vite build`, and it appears to succeed except it outputs this error: 
```
⠙ Writing functionMetadata.jsonError: Cannot find module 'https://deno.land/x/yextpages@plugins@1.0.0-beta.3/mod.ts'
Require stack:
- /Users/bhaines/repo/site-starter-react-basic/node_modules/@yext/pages/dist/vite-plugin/build/closeBundle/v7BupgmAEXVrwE8QVAewx.js
```

I believe "https://deno.land/x/yextpages@plugins@1.0.0-beta.3/mod.ts" is a valid deno import.

As far as I can tell this doesn't cause any real problems when deploying the site, but it is confusing for users. 

### Reproduction

https://github.com/yextconsulting/site-starter-react-consulting

### System Info

```shell
System:
    OS: macOS 12.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1017.66 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.8.0 - ~/.nvm/versions/node/v18.8.0/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.18.0 - ~/.nvm/versions/node/v18.8.0/bin/npm
    pnpm: 8.8.0 - ~/.nvm/versions/node/v18.8.0/bin/pnpm
  Browsers:
    Chrome: 118.0.5993.70
    Edge: 118.0.2088.46
    Safari: 15.4
  npmPackages:
    @yext/pages: 1.0.0-rc.5 => 1.0.0-rc.5
```

### Used Package Manager

npm

### Used Browser

Chrome

### Validations

- [X] Check that there isn't [already an issue](https://github.com/yext/pages/issues) that reports the same bug to avoid creating a duplicate.

PR #424, opened 2023-11-06
fix(plugin): address "Cannot find module" for deno module imports

alextaing:
Fixes https://github.com/yext/pages/issues/409.

This PR addresses the "Cannot find module" errors that were found when using imports from `deno.land` modules.  In the past, we were using ESBuild to build user files in a temp dir, then using the built file to get the default import name.  However, ESBuild was not able to resolve these `deno.land` imports and would throw errors. In order to resolve `http`/`https` imports, we would need a plugin for ESBuild (see "[Why this is not Recommended](https://esbuild.github.io/getting-started/#deno)").  I looked for an npm package that would provide ESBuild with a plugin to resolve these imports, however none of them seemed well established/stable enough to use.

Here are two I considered:
- [esbuild-plugin-cache](https://www.npmjs.com/package/esbuild-plugin-cache)
- [esbuild-plugin-http-imports](https://github.com/lucsoft/esbuild-plugin-http-imports)

To get around this issue, we now use `ts-morph` to parse through its AST instead since all we need is the name of the default export.  Specifically, we now can get the default export in the following formats.

```
export default function foo() {
  // do something
}
```
```
export default foo;
```

**Note that the `ts-morph` way takes longer than the previous ESBuild way (500ms instead of 30ms for 5 function files).** 

J=SUMO-5418
TEST=manual

Tested that the above formats were properly parsed and their metadatas properly populated in `functionMetadata.json`.  Also tested that on `yext pages serve`, that `http` functions could still be called as expected.

### DIFF
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -33,6 +33,7 @@ const getFunctionMetadataMap = async (
     )
     .map((f) => path.resolve(f));
 
+  project.addSourceFilesAtPaths(filepaths);
   const results = await Promise.allSettled(
     filepaths.map(generateFunctionMetadata)
   );
@@ -60,7 +61,6 @@ const getFunctionMetadataMap = async (
 async function generateFunctionMetadata(
   filepath: string
 ): Promise<[string, FunctionMetadata]> {
-  project.addSourceFileAtPath(filepath);
   const defaultExportDeclaration = project
     .getSourceFile(filepath)
     ?.getDefaultExportSymbol()
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -72,9 +72,9 @@ async function generateFunctionMetadata(
     )[0];
     if (!exportIdenfitier) {
       throw new Error(
-        `${relativePath} contains a default export assignment that is improperly ` +
-          "formatted. It may have an anonymous arrow function as the default " +
-          "export which is not supported."
+        `${relativePath} contains an improper default export assignment. ` +
+          "The default export must be a function, and it must be formatted " +
+          "as `export default foo;` for function `foo`."
       );
     }
     return [relativePath, { entrypoint: exportIdenfitier.getText() }];
@@ -82,9 +82,9 @@ async function generateFunctionMetadata(
     const entrypoint = defaultExportDeclaration.getName();
     if (!entrypoint) {
       throw new Error(
-        `${relativePath} contains a default function declaration that is improperly ` +
-          "formatted. It may have an anonymous function as the default export " +
-          "which is not supported."
+        `${relativePath} contains an improper default function declaration. ` +
+          "The default export function must be a named function, exported as " +
+          "`export default function foo(){}` for function `foo`"
       );
     }
     return [relativePath, { entrypoint }];
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -72,7 +72,7 @@ async function generateFunctionMetadata(
       ?.getText();
     if (!entrypoint) {
       throw new Error(
-        `${relativePath} contains an improper default export assignment. ` +
+        `${relativePath} contains an unsupported default export assignment. ` +
           "The default export must be a function, and it must be formatted " +
           "as `export default foo;` for function `foo`."
       );
@@ -82,7 +82,7 @@ async function generateFunctionMetadata(
     const entrypoint = defaultExportDeclaration.getName();
     if (!entrypoint) {
       throw new Error(
-        `${relativePath} contains an improper default function declaration. ` +
+        `${relativePath} contains an unsupported default function declaration. ` +
           "The default export function must be a named function, exported as " +
           "`export default function foo(){}` for function `foo`"
       );
@@ -92,7 +92,7 @@ async function generateFunctionMetadata(
   throw new Error(
     `${relativePath} does not contain a properly formatted default export. ` +
       "The default export must be named and declared either in the function declaration " +
-      "or in an `export default ...;` expresion."
+      "or in an `export default ...;` expression."
   );
 }
 
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -82,7 +82,7 @@ async function generateFunctionMetadata(
     const entrypoint = defaultExportDeclaration.getName();
     if (!entrypoint) {
       throw new Error(
-        `${relativePath} contains a default export assignment that is improperly ` +
+        `${relativePath} contains a default function declaration that is improperly ` +
           "formatted. It may have an anonymous function as the default export " +
           "which is not supported."
       );
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -1,20 +1,19 @@
 import fs from "fs";
 import path from "path";
-import esbuild from "esbuild";
-import { importFromString } from "module-from-string";
+import { Project, Node, SyntaxKind } from "ts-morph";
 import { glob } from "glob";
 import chalk from "chalk";
-import os from "os";
 import { ProjectStructure } from "../../../common/src/project/structure.js";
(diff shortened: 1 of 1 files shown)


packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,39 @@ const getFunctionMetadataMap = async (
...
+        `${relativePath} contains a default export assignment that is ` +
+          "improperly formatted. It may have an anonymous function as the default export " +
+          "which is not supported."
+      );
+    }
+    return [relativePath, { entrypoint: exportIdenfitier.getText() }];
+  } else if (Node.isFunctionDeclaration(defaultExportDeclaration)) {
+    const entrypoint = defaultExportDeclaration.getName();
+    if (!entrypoint) {
+      throw new Error(
```
asanehisa, line 84: Do we know in what case this would happen?

asanehisa:
Oh [here](https://github.com/yext/pages/blob/main/.github/workflows/semantic-pull-request.yml) are the valid scopes as well

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
alextaing: updated the docs to more accurately describe this!

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,40 @@ const getFunctionMetadataMap = async (
...
-  if (!importedFile.default) {
-    throw new Error(`${relativePath} does not contain a default export.`);
+  if (Node.isExportAssignment(defaultExportDeclaration)) {
+    const exportIdenfitier = defaultExportDeclaration.getChildrenOfKind(
+      SyntaxKind.Identifier
+    )[0];
+    if (!exportIdenfitier) {
+      throw new Error(
+        `${relativePath} contains a default export assignment that is improperly ` +
+          "formatted. It may have an anonymous arrow function as the default " +
```
nmanu1, line 76: There are several ways the assignment could be invalid for this use case. Out of curiosity, is there a reason we specifically call out anonymous functions?

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,40 @@ const getFunctionMetadataMap = async (
...
+          "export which is not supported."
+      );
+    }
+    return [relativePath, { entrypoint: exportIdenfitier.getText() }];
+  } else if (Node.isFunctionDeclaration(defaultExportDeclaration)) {
+    const entrypoint = defaultExportDeclaration.getName();
+    if (!entrypoint) {
+      throw new Error(
+        `${relativePath} contains a default export assignment that is improperly ` +
```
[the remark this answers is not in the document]
alextaing: oops yes, forgot to change that

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,40 @@ const getFunctionMetadataMap = async (
...
+      SyntaxKind.Identifier
+    )[0];
+    if (!exportIdenfitier) {
+      throw new Error(
+        `${relativePath} contains a default export assignment that is improperly ` +
+          "formatted. It may have an anonymous arrow function as the default " +
+          "export which is not supported."
+      );
+    }
+    return [relativePath, { entrypoint: exportIdenfitier.getText() }];
```
nmanu1, line 80: What happens if the default export assignment is not a function?

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
alextaing: Talked offline about different possibilities that it could be invalid. Changed the error message!

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
alextaing: Unfortunately this current way does not check that the default export is a function.  I don't think that there's a way to easily check with ts-morph if something is a function or not since the exported function could be derived from something imported, in which case ts-morph has no way of knowing what it is:

```
import pluginFactory from 'some/module';
const plugin = pluginFactory("params");
export default plugin;
```

The approach we had before also did not check that the default export was a function, so I don't think any functionality is being lost.  But I do agree that a function check would be nice to have.

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,40 @@ const getFunctionMetadataMap = async (
...
-    write: false,
-    format: "esm",
-    bundle: true,
-  });
-  const importedFile = await importFromString(buildResult.outputFiles[0].text);
+  project.addSourceFileAtPath(filepath);
+  const defaultExportDeclaration = project
+    .getSourceFile(filepath)
+    ?.getDefaultExportSymbol()
+    ?.getDeclarations()[0];
```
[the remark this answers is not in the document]
alextaing: Not in any standard case that I could come up with.  If there is no default export, then it will returned `undefined`. Even if there is a weird case where the user mistakenly has `export default ;`, `getDeclarations` will still return a node.

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
alextaing: I'll think some more about this!

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,40 @@ const getFunctionMetadataMap = async (
...
 ): Promise<[string, FunctionMetadata]> {
-  const buildResult = await esbuild.build({
-    entryPoints: [filepath],
-    outdir: TEMP_DIR,
-    write: false,
-    format: "esm",
-    bundle: true,
-  });
-  const importedFile = await importFromString(buildResult.outputFiles[0].text);
+  project.addSourceFileAtPath(filepath);
```
nmanu1, line 63: Is there a way to point to the functions dir when initializing the ts-morph project instead of adding each file manually? IIRC, each time you add a file to the project like this, ts-morph recreates the whole tsc instance, which impacts performance

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,40 @@ const getFunctionMetadataMap = async (
...
-
-  if (!importedFile.default) {
-    throw new Error(`${relativePath} does not contain a default export.`);
+  if (Node.isExportAssignment(defaultExportDeclaration)) {
+    const entrypoint = defaultExportDeclaration
+      .getChildrenOfKind(SyntaxKind.Identifier)[0]
+      ?.getText();
+    if (!entrypoint) {
+      throw new Error(
+        `${relativePath} contains an improper default export assignment. ` +
```
nmanu1, line 75: nit: maybe "unsupported" instead of "improper"?

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -61,21 +60,40 @@ const getFunctionMetadataMap = async (
...
+      );
+    }
+    return [relativePath, { entrypoint }];
   }
-
-  return [relativePath, { entrypoint: importedFile.default.name }];
+  throw new Error(
+    `${relativePath} does not contain a properly formatted default export. ` +
+      "The default export must be named and declared either in the function declaration " +
+      "or in an `export default ...;` expresion."
```
nmanu1, line 95: nit: expres**s**ion

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
nmanu1: Got it, I'm fine with leaving it as is then

packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
alextaing: cool! that made it 2x faster!

alextaing: I did it a little differently-- instead of adding it all on project instantiation, I just add all the files in one function call.  If we wanted to do it on project instantiation, we would have to make a tsconfig file specifying the directory we wanted, and pass the file path to the function.  I think that having a random floating tsconfig file just for the `functions` dir would be weird, so I went with this approach.

mkilpatrick (approved):

PR #429, opened 2023-11-07
feat(dev/plugin)!: drop support for Node 17 and support Node 20

jwartofsky-yext:
Tested `npm run dev` and `npm run prod` on the starter using Node 20. Page generated without issue.

However, I had to remove references to `import { fetch } from "@yext/pages/util";` from the starter since that no longer exists in pages. With that removed, the starter uses the default fetch function which works fine with Node 18 and newer.


.github/workflows/unit_test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         working-directory: ./packages/${{ inputs.package }}
     strategy:
       matrix:
-        node-version: [17.x, 18.x]
+        node-version: [18.x, 19.x, 20.x]
```
asanehisa, line 19: does this one need 19? besides that lgtm

jwartofsky-yext: Maybe? It had two before and we are still supporting 18, so I thought we would test 18-20

asanehisa (dismissed):

.github/workflows/jstest.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Build with NodeJS 18
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20
```
mkilpatrick, line 24: Keep this at 18

.github/workflows/playwright.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        node-version: [17.x, 18.x, 19.x]
+        node-version: [18.x, 19.x, 20.x]
```
mkilpatrick, line 11: Remove 19, we only want 18 and 20

.github/workflows/publish.yml
```diff
@@ -18,10 +18,10 @@ jobs:
       - name: Install pnpm
         uses: pnpm/action-setup@v2.2.4
 
-      - name: Set node version to 17.x
+      - name: Set node version to 20.x
```
mkilpatrick, line 21: Do 18 since that's our min version. Everywhere else too.

packages/pages/package.json
```diff
@@ -32,7 +32,7 @@
     }
   },
   "engines": {
-    "node": "^17 || ^18 || ^19"
+    "node": "^18 || ^19 || ^20"
```
mkilpatrick, line 35: Remove 19

packages/pages/src/util/index.ts
```diff
@@ -1,3 +1,2 @@
-export { fetch } from "./fetch.js";
```
mkilpatrick, line 1: Remove the custom fetch function too

jwartofsky-yext: Already done

mkilpatrick (approved):

PR #430, opened 2023-11-08
feat(dev/plugin): add localization to static pages

alextaing:
This PR adds localization to static pages and adds an interface for `StaticTemplateConfig`.  Localization can now be specified for static pages by including the following in the `TemplateConfig`.

```
export const config: StaticTemplateConfig = {
  name: "robots",
  locales: ["en", "es"]
};
```

This config is included in the `templates.json`/`features.json` to be consumed by Spruce as:

```
{
  "features": [
    {
      "name": "robots",
      "templateType": "JS",
      "staticPage": {
        "locales": ["en", "es"]
      }
    },
  ],
  "streams": []
}
```

Additionally, it was found that since static pages no longer have a single `staticURL` per `featureName`, the `localDataManifest` had to be refactored.  This is because static pages now accept a `document` in their `getPath` function, meaning that when static pages may have different URLs depending on the document if the path uses the `document`'s `locale`.

Furthermore, it was found that when running `yext pages render` to generate pages, static pages with a constant `getPath` return value multiple locales would only generate one static page.  To deal with this behavior, pages will now throw an error when users have multiple static pages with identical paths.

J=SUMO-5451
TEST=manual

It was tested that you can specify `locale`s in the `TemplateConfig` for static pages, and that they property show up in the `templates.json`.  It was also tested that when `yext pages generate-test-data` is called, it generates a document for each locale specified in a static page.


packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -18,10 +18,12 @@ export interface LocalDataManifest {
     {
       // The featureName for a specific static template
       featureName: string;
-      // The return value of the template's getPath()
-      staticURL: string;
-      // The locale codes
-      locales: string[];
+      staticPages: {
```
alextaing, line 21: static page `getPath` return val could use the locale.  So, we can't just have one staticURL per feature name anymore.

packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
```diff
@@ -15,12 +15,16 @@ export default async function findMatchingStaticTemplate(
       if (!isStaticTemplateConfig(t.config)) {
         return false;
       }
-      const document = await getLocalDataForEntityOrStaticPage({
-        entityId: "",
-        locale,
-        featureName: t.config.name,
-      });
-      return slug === t.getPath({ document });
+      try {
```
alextaing, line 18: Reason for change: We use this `getLocalDataForEntityOrStaticPage` to get a document to pass to the `getPath` function in order to get the path of a static template.  Since this function just iterates through all the templates using a locale that might not be relevant to other pages, it might not find a document it was asked to find and error out. For example.

We have two static pages with different locales
```
myHomePage: locales["en"]
myTacosPage: locales["en", "zh"]
```
and we are iterating through the documents, trying to find the Chinese (zh) tacos page.  At one point, `getLocalDataForEntityOrStaticPage` may attempt to run with `locale=zh` and `featureName=myHomePage` which does not exist.  This is an expected part of the iteration process, but should not break since it just means that  a document with those characteristics is not the right document.

packages/pages/src/common/src/template/types.ts
```diff
@@ -98,6 +98,24 @@ export interface TemplateConfig {
...
+  /** Locales for a static page */
+  locales?: string[];
+}
+
+/**
+ * Shape of TemlateConfig for Static Pages
+ *
+ * @public
+ */
+export interface StaticTemplateConfig {
```
alextaing, line 110: @mkilpatrick Would we rather have this as a `type` that picks from `TemplateConfig` instead to reduce repetition? Or is something like this OK?

alextaing:
@mkilpatrick One thing of note that I ran into when testing: If we have a static page with more than one locale, but a constant `getPath` return value:
```
export const config: StaticTemplateConfig = {
  name: "robots",
  locales: ['en', 'zh']
};
export const getPath = () => {
  return `robots.txt`;
};
```
In dev, it populates 2 links. `robots.txt:?locale=en` and  `robots.txt:?locale=zh`. But, when we generate pages using `yext pages render`, it only has one page , `robots.txt` and is filled with the contents of the page with the `en` locale.  I think that this issue might be slightly outside the scope of this item since I'm not sure whether we should expect one of the locales to be overwritten, or if it should have two pages as we do in dev.

packages/pages/src/common/src/template/types.ts
```diff
@@ -98,6 +98,24 @@ export interface TemplateConfig {
   alternateLanguageFields?: string[];
   /** The name of the onUrlChange function to use. */
   onUrlChange?: string;
+  /** Locales for a static page */
+  locales?: string[];
+}
+
+/**
+ * Shape of TemlateConfig for Static Pages
```
asanehisa, line 106: TemplateConfig

asanehisa (commented):
lgtm

packages/pages/src/common/src/template/types.ts
mkilpatrick: This is ok in case they diverge a lot.

packages/pages/src/common/src/template/types.ts
```diff
@@ -98,6 +98,24 @@ export interface TemplateConfig {
   alternateLanguageFields?: string[];
   /** The name of the onUrlChange function to use. */
   onUrlChange?: string;
+  /** Locales for a static page */
+  locales?: string[];
```
mkilpatrick, line 102: I don't think we want this. A templateConfig (which now represents and entity template) doesn't have this option. Locales exist in the Stream definition.

packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
mkilpatrick: Great explanation! Thanks for providing this.

mkilpatrick:
> @mkilpatrick One thing of note that I ran into when testing: If we have a static page with more than one locale, but a constant `getPath` return value:
> 
> ```
> export const config: StaticTemplateConfig = {
>   name: "robots",
>   locales: ['en', 'zh']
> };
> export const getPath = () => {
>   return `robots.txt`;
> };
> ```
> 
> In dev, it creates two pages: `robots.txt?locale=en` and `robots.txt?locale=zh`. But, when we generate pages using `yext pages render`, it only has one page , `robots.txt` and is filled with the contents of the page with the `en` locale.
> 
> Is this something we want to address as a part of this item, or something that we should make an item for? As of now, I'm not sure whether we should expect one of the locales to be overwritten, or if it should have two pages as we do in dev.

This is a good question. Could you bring it up in the original Slack thread discussion? Perhaps we should instead throw an error that you've defined multiple locales but only one output path/html file can be created.

packages/pages/src/common/src/template/types.ts
alextaing: I see yes, that makes sense.  Both `TemplateConfig` and `StaticTemplateConfig` both eventually get turned into `TemeplateConfigInternal` which does have `locale`, and that is passed through and used thought Pages.  `TemplateConfig` is just a user facing interface. Removed `locale` from `TemplateConfig`!

mkilpatrick: Precisely.

mkilpatrick (approved):
🔥

PR #433, opened 2023-11-08
feat(dev/plugin): support secrets

mkilpatrick:
This adds support for truly secret environment variables. Before this change we only supported _public_ environment variables that were prefixed with `YEXT_PUBLIC`.

### DIFF
--- a/packages/pages/src/util/processEnvVariables.test.ts
+++ b/packages/pages/src/util/processEnvVariables.test.ts
@@ -4,30 +4,34 @@ import { processEnvVariables } from "./processEnvVariables.js";
 describe("processEnvVariables", () => {
   beforeAll(() => {
     global.process.env = {
-      NODE_ENV: "development",
+      NODE_ENV: "production",
       VITE_KEY: "pk.abcdefghij",
       YEXT_PUBLIC_KEY: "pk.0123456789",
+      SECRET: "secret",
     };
   });
 
-  it("filters .env for Vite specific variables", () => {
+  it("filters .env for Vite specific variables for production", () => {
     const env = processEnvVariables();
-
     expect(Object.keys(env).length).toEqual(1);
     expect(env.VITE_KEY).toEqual(`"pk.abcdefghij"`);
   });
 
-  it("filters .env for Vite specific variables with custom prefix", () => {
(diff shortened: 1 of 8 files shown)


packages/pages/src/util/processEnvVariables.ts
```diff
@@ -8,20 +8,21 @@ import { loadEnv } from "vite";
...
-    Object.entries(processEnv)
-      .filter(([env]) => env.startsWith(prefix))
-      .map(([key, value]) => [
-        key,
-        stringifyValues ? JSON.stringify(value) : value,
-      ])
-  );
 
-  return processEnv;
+  // If we're development, return all env var keys, otherwise use Vite's default
```
asanehisa, line 15: nit: if we're in development

asanehisa (approved):
lgtm

packages/pages/src/util/processEnvVariables.test.ts
```diff
@@ -4,30 +4,34 @@ import { processEnvVariables } from "./processEnvVariables.js";
...
 
     expect(Object.keys(env).length).toEqual(1);
     expect(env.YEXT_PUBLIC_KEY).toEqual(`"pk.0123456789"`);
   });
 
-  it("does not stringify values", () => {
-    const env = processEnvVariables("YEXT_PUBLIC", false);
+  it("returns all env vars for development", () => {
+    global.process.env.NODE_ENV = "development";
+    const env = processEnvVariables("YEXT_PUBLIC");
```
jwartofsky-yext, line 29: We could have an example with another key, like VITE

jwartofsky-yext (approved):
lgtm

packages/pages/src/util/processEnvVariables.test.ts
mkilpatrick: The key actually doesn't matter in this case as it returns everything. Ack.

asanehisa (dismissed):

mkilpatrick:
I've opened an [issue](https://github.com/vitejs/vite/issues/14929) with VIte re: the Ubuntu failures. Looks like it has to do with loadEnv and an empty string prefix.

asanehisa (approved):

PR #436, opened 2023-11-14
refactor: adjust how site-stream.json is migrated 

asanehisa:
If site-steam.json is `{"filter":{"entityIds":["1234"]},"fields":["id","meta","name"], "reverseProxy":{"displayUrlPrefix": "foo"}}`
then after migration, it'll show up as 
```
siteStream:
  entityId: "1234"
  serving:
    reverseProxyPrefix: foo
  fields:
    - id
    - meta
    - name

```
in the config.yaml. 

J=SUMO-5481
TEST=manual

### DIFF
--- a/packages/pages/src/upgrade/migrateConfig.ts
+++ b/packages/pages/src/upgrade/migrateConfig.ts
@@ -92,14 +92,38 @@ const migrateSiteStream = async (
     console.info(
       `migrating global data from ${siteStreamPath} to ${configYamlPath}`
     );
-    // this logic replaces $id with id and keeps id in the first position
-    const newSiteStream = {
-      id: sitesJson.$id,
-      ...sitesJson,
-    };
-    newSiteStream.$id = undefined;
+    const newSiteStream = formatSiteStream(sitesJson, siteStreamPath);
+    if (!newSiteStream) {
+      return null;
+    }
     writeYamlSync(configYamlPath, "siteStream", newSiteStream);
   }
+};
+
+const formatSiteStream = (sitesJson: any, siteStreamPath: string) => {
+  let entityId: string | undefined = undefined;
+  if (sitesJson.filter?.entityIds && sitesJson.filter?.entityIds.length === 1) {
+    entityId = sitesJson.filter.entityIds[0];
+  } else if (sitesJson.filter?.entityIds) {
+    console.error(
+      `unable to migrate ${siteStreamPath} due to multiple entityIds`
+    );
+    return null;
+  }
+
+  const siteStream = {
+    id: sitesJson.$id,
+    entityId: entityId,
+    serving: {
+      reverseProxyPrefix: sitesJson.reverseProxy?.displayUrlPrefix,
+    },
+    ...sitesJson,
+  };
+  siteStream.$id = undefined;
+  siteStream.reverseProxy = undefined;
+  siteStream.filter = undefined;
+
+  return siteStream;
 };
 
 const migrateRedirects = async (source: string, dest: string) => {


packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -92,16 +92,40 @@ const migrateSiteStream = async (
...
     writeYamlSync(configYamlPath, "siteStream", newSiteStream);
   }
 };
 
+const formatSiteStream = (sitesJson: any, siteStreamPath: string) => {
+  let entityId: string | undefined = undefined;
+  if (sitesJson.filter?.entityIds && sitesJson.filter?.entityIds.length === 1) {
+    entityId = sitesJson.filter.entityIds[0];
+  } else if (sitesJson.filter?.entityIds) {
+    console.error(
```
mkilpatrick, line 108: We should use the colorized version and also kill the update script. You can use `logErrorAndExit` for that.

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -92,16 +92,40 @@ const migrateSiteStream = async (
...
+    const newSiteStream = formatSiteStream(sitesJson, siteStreamPath);
+    if (!newSiteStream) {
+      return null;
+    }
     writeYamlSync(configYamlPath, "siteStream", newSiteStream);
   }
 };
 
+const formatSiteStream = (sitesJson: any, siteStreamPath: string) => {
+  let entityId: string | undefined = undefined;
```
mkilpatrick, line 104: I don't think you actually need to set undefined.

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -92,16 +92,40 @@ const migrateSiteStream = async (
...
+  if (sitesJson.filter?.entityIds && sitesJson.filter?.entityIds.length === 1) {
+    entityId = sitesJson.filter.entityIds[0];
+  } else if (sitesJson.filter?.entityIds) {
+    console.error(
+      `unable to migrate ${siteStreamPath} due to multiple entityIds`
```
mkilpatrick, line 109: Capitalize the first letter

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -92,16 +92,40 @@ const migrateSiteStream = async (
...
+    );
+    return null;
+  }
+
+  const siteStream = {
+    id: sitesJson.$id,
+    entityId: entityId,
+    serving: {
+      reverseProxyPrefix: sitesJson.reverseProxy?.displayUrlPrefix,
```
mkilpatrick, line 118: If displayUrlPrefix isn't set we shouldn't set serving at all.

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -92,16 +92,40 @@ const migrateSiteStream = async (
...
+    },
+    ...sitesJson,
```
mkilpatrick, line 120: You should do this first, otherwise it will overwrite any keys that came before it (like id).

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -92,16 +92,40 @@ const migrateSiteStream = async (
...
+  };
+  siteStream.$id = undefined;
```
mkilpatrick, line 122: Why are we clearing this?

asanehisa: That was part of earlier logic (in a diff commit) to replace $id with id

packages/pages/src/upgrade/migrateConfig.ts
asanehisa: Got it, moved it after id b/c in earlier commit, comments said it was purposefully after id such that id is first.

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -93,32 +94,30 @@ const migrateSiteStream = async (
...
     ...sitesJson,
+    entityId: entityId,
   };
+  if (siteStream.reverseProxy?.displayUrlPrefix) {
+    siteStream.serving = {
+      reverseProxyPrefix: sitesJson.reverseProxy?.displayUrlPrefix,
+    };
+  }
   siteStream.$id = undefined;
   siteStream.reverseProxy = undefined;
```
mkilpatrick, line 122: Use `delete`s for these instead, ie `delete siteStream.reverseProxy`

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -93,32 +94,30 @@ const migrateSiteStream = async (
...
-    console.error(
-      `unable to migrate ${siteStreamPath} due to multiple entityIds`
+    logErrorAndExit(
+      `Unable to migrate ${siteStreamPath} due to multiple entityIds`
     );
-    return null;
   }
 
   const siteStream = {
     id: sitesJson.$id,
```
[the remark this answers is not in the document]
asanehisa: Added comment instead.

mkilpatrick (dismissed):

packages/pages/src/upgrade/migrateConfig.ts
```diff
@@ -92,16 +93,39 @@ const migrateSiteStream = async (
...
-      id: sitesJson.$id,
-      ...sitesJson,
-    };
-    newSiteStream.$id = undefined;
+    const newSiteStream = formatSiteStream(sitesJson, siteStreamPath);
     writeYamlSync(configYamlPath, "siteStream", newSiteStream);
   }
 };
 
+const formatSiteStream = (sitesJson: any, siteStreamPath: string) => {
```
mkilpatrick, line 101: Add some tests for this method specifically.

packages/pages/src/upgrade/migrateConfig.test.ts
```diff
@@ -0,0 +1,33 @@
...
+const siteStreamPath = "foo/bar";
+
+describe("formatSiteStream", () => {
+  it("errors and exits when there are multiple entityIds", () => {
+    const testJson = { filter: { entityIds: ["1234", "123"] } };
+    const mockExit = vi
+      .spyOn(process, "exit")
+      .mockImplementation(() => undefined as never);
+    formatSiteStream(testJson, siteStreamPath);
+    expect(mockExit).toHaveBeenCalledWith(1);
```
mkilpatrick, line 13: Do you want toHaveBeenCalledTimes here? You're not passing a particular function param was called

asanehisa: Oh I meant toHaveBeenCalledWith(1) as it calls an exit 1.

mkilpatrick (approved):

PR #441, opened 2023-11-16
feat: auto generate markdown files 

asanehisa:


packages/pages/src/common/src/template/paths.ts
```diff
@@ -19,6 +19,7 @@ export const getRelativePrefixToRootFromPath = (path: string): string => {
 /**
  * Converts any path to a posix path delimited by "/". Useful for ensuring that a path will be dynamically importable as only posix-style
  * paths are supported with import.
+ * @public
```
mkilpatrick, line 22: Only things we export for users should be marked @public. So not this function since it's internal only.

packages/pages/src/dev/dev.ts
```diff
@@ -65,7 +65,9 @@ const handler = async ({
 
   if (openBrowser) await open(`http://localhost:${devServerPort}/`);
 };
-
+/**
+ * @public
```
mkilpatrick, line 69: remove

packages/pages/src/dev/server/server.ts
```diff
@@ -24,6 +24,9 @@ import { isUsingConfig } from "../../util/config.js";
 import yaml from "yaml";
 const __dirname = dirname(fileURLToPath(import.meta.url));
 
+/**
+ * @public
```
mkilpatrick, line 28: remove

.github/workflows/generate_markdown_files.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:    
+      - name: "Checkout"
+        uses: actions/checkout@v4
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2.2.4
+      - name: Install deps
+        run: pnpm i
+      - name: generate-docs
+        run: pnpm -r run api-extractor && pnpm -r run generate-docs && pnpm run fmt
```
mkilpatrick, line 16: I think you can just call `pnpm run docs` since the docs command already exists in the root package.json

packages/pages/src/common/src/template/paths.ts
asanehisa: Got it. api-extractor errors if it doesn't have one of the following: public / internal / beta / alpha

packages/pages/src/dev/dev.ts
asanehisa: replaced with internal

packages/pages/src/dev/server/server.ts
asanehisa: replaced with internal

packages/pages/src/common/src/template/paths.ts
mkilpatrick: Is that a new requirement? When I had it enabled in the past I never had to mark internal things specifically. Is it because the file is exported or something?

.github/workflows/generate_markdown_files.yml
```diff
@@ -13,4 +13,16 @@ jobs:
...
       - name: generate-docs
-        run: pnpm -r run api-extractor && pnpm -r run generate-docs && pnpm run fmt
\ No newline at end of file
+        run: pnpm run docs
+      - name: Auto Generate Markdown Files
+        id: generate-markdown
+        uses: EndBug/add-and-commit@v9
+        with:
+          message: "Auto generate markdown files and push"
+          add: "*"
```
mkilpatrick, line 22: Only add .md files

packages/pages/src/common/src/template/paths.ts
asanehisa: I think b/c it's exported since only a couple functions were causing the errors.

.github/workflows/generate_markdown_files.yml
```diff
@@ -13,4 +13,16 @@ jobs:
...
         run: pnpm i
       - name: generate-docs
-        run: pnpm -r run api-extractor && pnpm -r run generate-docs && pnpm run fmt
\ No newline at end of file
+        run: pnpm run docs
+      - name: Auto Generate Markdown Files
+        id: generate-markdown
+        uses: EndBug/add-and-commit@v9
+        with:
+          message: "Auto generate markdown files and push"
```
mkilpatrick, line 21: This is the commit message. Change it to "Auto generated documentation"

packages/pages/docs/api/pages.clientserverrendertemplates.iscustomrendertemplate.md
```diff
@@ -0,0 +1,13 @@
...
+[Home](./index.md) &gt; [@yext/pages](./pages.md) &gt; [ClientServerRenderTemplates](./pages.clientserverrendertemplates.md) &gt; [isCustomRenderTemplate](./pages.clientserverrendertemplates.iscustomrendertemplate.md)
+
+## ClientServerRenderTemplates.isCustomRenderTemplate property
+
+If the render templates are user-defined or built-in
+
+**Signature:**
+
+```typescript
+isCustomRenderTemplate: boolean;
```
mkilpatrick, line 12: Some of these generated files should not be public. Anything internal to PagesJS and not external functions a user could import should not be documented.

packages/pages/tsdoc-metadata.json
```diff
@@ -0,0 +1,11 @@
+// This file is read by tools that parse documentation comments conforming to the TSDoc standard.
+// It should be published with your NPM package.  It should not be tracked by Git.
```
mkilpatrick, line 2: This comment sounds like we need to add this to .gitignore

.github/workflows/generate_markdown_files.yml
asanehisa: Done! (had to force push to test and see what files got pushed by bot, it would've pushed nothing if nothing changed)

packages/pages/docs/api/pages.clientserverrendertemplates.iscustomrendertemplate.md
asanehisa: Agree. It makes docs for everything marked as public + any dependencies those functions have.

for example: 
```
// @public
export interface Site {
  /** some comment. */
  branchId: string;
  ...
  }
```

Then we'd have a .md file for Site as well as for branchId

packages/pages/tsdoc-metadata.json
asanehisa: Yes! I wasn't suppose to allow the workflow to add + commit that file

packages/pages/etc/pages.api.md
```diff
@@ -3,18 +3,56 @@
...
+import { Command } from "commander";
+import { default as React_2 } from "react";
 
 // @public
 export type Attributes = Record<string, string>;
 
+// Warning: (ae-internal-missing-underscore) The name "ClientServerRenderTemplates" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal
+export interface ClientServerRenderTemplates {
```
mkilpatrick, line 15: These 4 internal funcs shouldn't have docs. We shouldn't need to do the underscore thing it's suggesting. I think we just need to not export them at the top level wherever we are.

packages/pages/src/common/src/template/types.ts
```diff
@@ -1,5 +1,6 @@
 import { ProjectStructureConfig } from "../project/structure.js";
 import { HeadConfig } from "./head.js";
+import React from "react";
```
mkilpatrick, line 3: What this necessary? It's a .ts file that doesn't use react.

packages/pages/src/common/src/template/types.ts
```diff
@@ -241,6 +248,8 @@ export interface RenderTemplate {
 
 /**
  * Context of a page, which defines the template itself and its props.
+ *
+ * @internal
```
mkilpatrick, line 252: This should be public

packages/pages/src/common/src/template/types.ts
asanehisa: Ah you're right. I assumed we needed it for React.JSX.Element, but we don't.

asanehisa: Oh nvm, we do or else I get an error 
`│ ERROR: Internal Error: Unable to follow symbol for "React"`
from the api-extractor

packages/pages/tsdoc-metadata.json
mkilpatrick: Wait, we don't want the workflow to commit this either. The message says `It should not be tracked by Git` so this file should be added to .gitgnore.

packages/pages/src/common/src/template/types.ts
mkilpatrick: Hmm. This is dubious. 
1. React isn't used in this file at all
2. The code is correct (compiles np) so why does api-extractor need it?

packages/pages/tsdoc-metadata.json
asanehisa: It is being ignored by the workflow currently (not part of the files the workflow commits). This file shouldn't be present in current commit

mkilpatrick: Got it 👍

packages/pages/src/common/src/template/types.ts
asanehisa: Hmmm unfortunately the error doesn't give a whole lot of detail. 
Full error is: 
`ERROR: Internal Error: Unable to follow symbol for "React"
│ You have encountered a software defect. Please consider reporting the issue to the maintainers of this application.`

Someone else had this [error pop-up due to a missing import](https://github.com/microsoft/rushstack/issues/3612) which caused me to try and import React which fixed the error.

packages/pages/etc/pages.api.md
asanehisa: There are not a lot of [configurable settings](https://api-extractor.com/pages/configs/api-extractor_json/#api-report-section) available for how the api-extractor creates the API Report.

mkilpatrick: Are you saying that .md files are generated but documentation doesn't get added to doc/api? Actually I think that's ok. I think I was misremembering that md files were only for public things, but it's actually for everything and then only public things to to docs/api.

mkilpatrick (approved):

PR #444, opened 2023-11-21
feat: add node engines to upgrade script

mkilpatrick:
Adds supported engines to package.json

```
"engines": {
    "node": "^18.0.0 || >=20.0.0"
  },
```

### DIFF
--- a/packages/pages/src/upgrade/pagesUpdater.ts
+++ b/packages/pages/src/upgrade/pagesUpdater.ts
@@ -119,16 +119,18 @@ const installDependencies = async (
 };
 
 // Function to update package.json engines
+const NODE_ENGINES = "^18.0.0 || >=20.0.0";
 const updatePackageEngines = (targetDirectory: string) => {
   const packageJsonPath = path.resolve(targetDirectory, "package.json");
 
   try {
     const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
     let engines = packageJson.engines;
-    if (!engines) {
-      engines = { node: "" };
+    if (engines) {
+      engines.node = NODE_ENGINES;
+    } else {
+      engines = { node: NODE_ENGINES };
     }
-    engines.node = "^18.0.0 || >=20.0.0";
     packageJson.engines = engines;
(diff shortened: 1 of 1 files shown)


packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -118,6 +118,25 @@ const installDependencies = async (
...
+const updatePackageEngines = (targetDirectory: string) => {
+  const packageJsonPath = path.resolve(targetDirectory, "package.json");
+
+  try {
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
+    let engines = packageJson.engines;
+    if (!engines) {
+      engines = { node: "" };
+    }
+    engines.node = "^18.0.0 || >=20.0.0";
```
jwartofsky-yext, line 131: Kind of strange that we support 18 and 20+, but not 19

mkilpatrick: We are support what Vite supports. Odd versions of Node are quick betas. Once the next even version is released the prior odd becomes deprecated, with the previous even being in LTS.

mkilpatrick: https://nodejs.org/en/about/previous-releases

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -118,6 +118,25 @@ const installDependencies = async (
...
 };
 
+// Function to update package.json engines
+const updatePackageEngines = (targetDirectory: string) => {
+  const packageJsonPath = path.resolve(targetDirectory, "package.json");
+
+  try {
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
+    let engines = packageJson.engines;
+    if (!engines) {
```
jwartofsky-yext, line 128: Could also use:

NODE_ENGINE = "^18.0.0 || >=20.0.0";
if (engines) {
  engines.node = NODE_ENGINE;
else {
  engines = { node: NODE_ENGINE}
}

jwartofsky-yext (approved):

jwartofsky-yext (approved):
lgtm

PR #446, opened 2023-11-27
feat(deps): update to Vite 5

mkilpatrick:
Upgrades to Rollup 4 as well, which is a requirement for Vite 5.

We now expose Vite as a peerDependency where both Vite 4 and 5 are supported.

### DIFF
--- a/packages/pages/src/util/processEnvVariables.ts
+++ b/packages/pages/src/util/processEnvVariables.ts
@@ -12,7 +12,7 @@ export const processEnvVariables = (
 ): Record<string, string> => {
   const mode = process.env.NODE_ENV || "development";
 
-  // If we're development return all env var keys, otherwise use Vite's default
+  // If we're in development return all env var keys, otherwise use Vite's default
   // way of loading env vars in prod.
   // For prod, only public env vars are loaded since they are statically replaced in code.
   // Cog makes the non-public env vars available as global vars in the Deno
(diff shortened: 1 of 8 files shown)


asanehisa (dismissed):

packages/pages/src/util/processEnvVariables.ts
```diff
@@ -21,10 +21,43 @@ export const processEnvVariables = (
...
+      // The define key "BASH_FUNC_protosearch%%" must be a valid identifier
+      .filter(([env]) => isIdentifier(env))
       // The value must be stringified: https://vitejs.dev/config/shared-options.html#define
       .map(([key, value]) => [key, JSON.stringify(value)])
   );
 };
+
+const startRunes = "_$abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+const continueRunes =
+  "_$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
asanehisa, line 38: will these change? if not, enums?

mkilpatrick: They won't change. It's easier to have a list of strings to do an includes check than it is to define all as enums and switch case on them.

asanehisa (approved):

asanehisa (approved):
Lgtm! Just odd tests not running haha

PR #451, opened 2023-12-11
feat(dev): upgrade script updates Pages version

jwartofsky-yext:
Running the upgrade script will now upgrade Pages version.

This also adds a few new steps that will be executed, such as checking node version and breaking down old helper functions.


packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -43,38 +62,156 @@ const processDirectoryRecursively = (currentPath: string) => {
...
+      }
     }
   });
+}
+
+/**
+ * Update sitesComponents to the latest version
+ * @param targetDirectory
+ */
+export const updateSitesComponents = async (targetDirectory: string) => {
```
jwartofsky-yext, line 83: This will be removed and the updatePagesComponents step will be used instead

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -12,8 +25,20 @@ interface UpgradeArgs {
...
+  await updateDevDependencies(source);
+  checkLegacyMarkdown(source);
+  replacePagesSlashComponentsImport(source);
+  replaceSitesComponentsImports(source);
+  removeFetchImport(source);
+  updatePackageScripts(source);
+  updatePackageEngines(source);
+  checkNodeVersion();
+  await updatePages(source);
+  // await updatePagesComponents(source);
```
jwartofsky-yext, line 40: This will be enabled and the updateSitesComponent call removed once pages-components is released

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -135,28 +283,27 @@ const updatePackageEngines = (targetDirectory: string) => {
...
+    `Please install node ${NODE_ENGINES}.`;
+  try {
+    const nodeVersion = execSync("node -v");
+    if (!nodeVersion || nodeVersion.length < 4) {
+      console.log(ERR_MSG);
+      return;
+    }
+    const version = parseInt(nodeVersion.toString().split(".")[0].substring(1));
+    if (version != 18 && version != 20) {
+      console.warn(
```
mkilpatrick, line 302: error instead of warn

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -12,8 +25,20 @@ interface UpgradeArgs {
 
 const handler = async (args: UpgradeArgs) => {
   const scoped = { scope: args.scope || "" };
+  const source = path.resolve(scoped.scope);
   const projectStructure = await ProjectStructure.init(scoped);
-  await updatePages(projectStructure);
+  await updateSitesComponents(source);
```
mkilpatrick, line 30: Can you move this one right before `updatePages` since it's the same thing? Actually, it'd be better to combine `updatePages` and `updatePagesComponents` to each do the two things: update the imports and update the package.json, since it's really one thing we're trying to update overall. The method can call 2 sub-methods.

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -12,8 +25,20 @@ interface UpgradeArgs {
...
+  await updateDevDependencies(source);
+  checkLegacyMarkdown(source);
+  replacePagesSlashComponentsImport(source);
+  replaceSitesComponentsImports(source);
+  removeFetchImport(source);
+  updatePackageScripts(source);
+  updatePackageEngines(source);
+  checkNodeVersion();
+  await updatePages(source);
```
mkilpatrick, line 39: Give this a more descriptive name.

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -43,38 +62,156 @@ const processDirectoryRecursively = (currentPath: string) => {
...
+export const updateSitesComponents = async (targetDirectory: string) => {
+  await updatePackageDependency(
+    targetDirectory,
+    "@yext/sites-components",
+    null
+  );
+};
+
+/**
+ * Update pagesComponents to the latest version
```
mkilpatrick, line 92: Give this a better description with example (since this will combine the two methods now given my other comment).

packages/pages/src/upgrade/upgrade.ts
```diff
@@ -12,8 +25,20 @@ interface UpgradeArgs {
 
 const handler = async (args: UpgradeArgs) => {
   const scoped = { scope: args.scope || "" };
+  const source = path.resolve(scoped.scope);
   const projectStructure = await ProjectStructure.init(scoped);
-  await updatePages(projectStructure);
+  await updateSitesComponents(source);
+  await updateDevDependencies(source);
+  checkLegacyMarkdown(source);
+  replacePagesSlashComponentsImport(source);
```
mkilpatrick, line 33: This should be combined with updateSitesComponents

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -43,38 +62,156 @@ const processDirectoryRecursively = (currentPath: string) => {
...
+ * @param targetDirectory
+ */
+export const updatePagesComponents = async (targetDirectory: string) => {
```
mkilpatrick, line 95: This will need to update the imports. I don't think you're doing that anywhere.

packages/pages/src/upgrade/upgrade.ts
jwartofsky-yext: Would it be updateYextPages? This is the command that updates @yext/pages to the version that is called by npx pages upgrade

mkilpatrick: updatePagesJSToCurrentVersion

packages/pages/src/upgrade/upgrade.ts
mkilpatrick: I confused `updatePages` because the name was not clear. At any rate, I mean to combine the updating of the sites-components package and updating all of the imports under one function.

packages/pages/src/upgrade/pagesUpdater.ts
jwartofsky-yext: This command just updates the dependency.

replacePagesSlashComponentsImport replaces @yext/pages/components -> @yext/pages/sites-components
replaceSitesComponentsImports replaces @yext/sites-components -> @yext/pages-components

packages/pages/src/upgrade/upgrade.ts
jwartofsky-yext: In the future, I think we will likely want to be updating the import but not updating sites-components in package.json, because sites-components will no longer exist as it is replaced with pages-components

mkilpatrick: Right. This will need to be part of the (either): `updateToUseSitesComponents` (current) or `updateToUsePagesComponents` (latest one)

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -77,7 +77,8 @@ function processDirectoryRecursively(
 }
 
 /**
- * Update sitesComponents to the latest version
+ * Update sites-components to the latest version and replace pages/components imports with
+ * sites-components.
  * @param targetDirectory
  */
 export const updateSitesComponents = async (targetDirectory: string) => {
```
mkilpatrick, line 84: Let's rename to `updateToUseSitesComponents`

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -86,10 +87,13 @@ export const updateSitesComponents = async (targetDirectory: string) => {
...
+  replacePagesSlashComponentsImports(targetDirectory);
 };
 
 /**
- * Update pagesComponents to the latest version
+ * Update pages-components to the latest version and replace sites-components imports with
+ * pages-components.
  * @param targetDirectory
  */
 export const updatePagesComponents = async (targetDirectory: string) => {
```
mkilpatrick, line 99: Let's rename to `updateToUsePagesComponents`

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -98,13 +102,17 @@ export const updatePagesComponents = async (targetDirectory: string) => {
     "@yext/pages-components",
     null
   );
+  // update imports from sites-components to pages-components
+  replaceSitesComponentsImports(targetDirectory);
```
mkilpatrick, line 106: This func should call this, but also do replacePagesSlashComponentsImports for this new package. We need to handle both.

mkilpatrick (dismissed):

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -106,6 +107,8 @@ export const updateToUsePagesComponents = async (targetDirectory: string) => {
   replacePagesSlashComponentsImports(targetDirectory);
   // update imports from sites-components to pages-components
   replaceSitesComponentsImports(targetDirectory);
+  // update imports from react-components to pages-components
+  replaceReactComponentsImports(targetDirectory);
```
mkilpatrick, line 111: We need to remove react-components and sites-components from package.json too

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -33,17 +35,58 @@ async function updatePackageDependency(
     process.exit(1);
   }
   try {
+    const dependencyType = devDependency ? "devDependencies" : "dependencies";
```
mkilpatrick, line 38: I don't think we can assume where they've installed it. Just look in both places. Same for the remove.

jwartofsky-yext: We can look in both places for upgrade or remove, but if we are installing we have to pick one. I guess this would default to devDependency then?

All of these packages live under devDependencies on the starter

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -35,10 +37,13 @@ async function updatePackageDependency(
...
   }
   try {
-    const dependencyType = devDependency ? "devDependencies" : "dependencies";
+    let dependencyType = DEV_DEPENDENCIES;
     const packageJson = JSON.parse(fs.readFileSync(packagePath, "utf-8"));
-    const toVersion = version || (await latestVersion(packageName));
+    if (packageJson.dependencies[packageName]) {
+      dependencyType = DEPENDENCIES;
+    }
     const currentVersion = packageJson[dependencyType][packageName];
```
mkilpatrick, line 45: Don't we need to check if it's in both, otherwise this could fail?

jwartofsky-yext: It defaults to devDependencies. It then checks if it is in dependencies. If it is in dependencies, it will use that. Otherwise will use devDependencies

mkilpatrick: Right, but it checks if the package is in devDeps, if not, it then assumes it's in deps but doesn't actually verify that the package actually exists there.

jwartofsky-yext: This covers all cases.

For upgrade it will:

1. install in devDependencies if not found
2. upgrade in devDependecies if it is there
3. upgrade in dependencies if it is there

For remove it will:
1. remove if it is in devDependencies
2. remove if it is in dependencies

jwartofsky-yext: It will also show a unique log 'Installing ${packageName' if the package doesn't exist at all, and 'Upgrading to' if it is found and is upgrading

jwartofsky-yext: Made some changes

If user has sites-components or react-components in their package.json, we will install pages-components. If they have pages/components imports, sites-components imports, or react-components imports, we also install pages-components. If they already have pages-components, we upgrade it regardless.

All other packages we only upgrade if they already exist in devDependencies or dependencies

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -6,13 +6,13 @@ import { execSync } from "child_process";
...
-const sitesComponentsRegex = /"@yext\/sites-components"/g;
-const sitesComponentsReplacement = '"@yext/sites-components"';
-const pagesComponentsReplacement = '"@yext/pages-components"';
-const reactComponentsRegex = /"@yext\/react-components"/g;
-const fetchImportRegex = /\nimport { fetch } from "@yext\/pages\/util";/g;
-const markdownRegex = 'Markdown.{1,10}from "@yext\\/react-components';
+const pagesSlashComponentsRegex = /@yext\/pages\/components/g;
+const sitesComponentsRegex = /@yext\/sites-components/g;
+const reactComponentsRegex = /@yext\/react-components/g;
+const fetchImportRegex = /\nimport { fetch } from ["']@yext\/pages\/util["'];/g;
```
mkilpatrick, line 12: Not from this change, but this doesn't catch no spaces around `fetch`

`import {fetch} from '@yext/pages/util';`

or more than one import

`import { fetch, foo } from '@yext/pages/util';`

also what's the \n for? I think that shouldn't be there.

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -48,6 +49,9 @@ async function updatePackageDependency(
       return;
     }
     if (!currentVersion) {
+      if (!install) {
```
mkilpatrick, line 52: Make this check before getting the latestVersion. It's a network call that is unnecessary if we're not forcing the install.

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -132,19 +139,34 @@ function processDirectoryRecursively(
...
   // update imports from react-components to pages-components
-  replaceReactComponentsImports(targetDirectory);
+  const hasReactComponentsImports =
+    replaceReactComponentsImports(targetDirectory);
+
+  await updatePackageDependency(
+    targetDirectory,
+    "@yext/pages-components",
+    null,
+    hasSitesComponents ||
```
mkilpatrick, line 164: We actually only want to add pages-components if one has been replaced, not if the previous packages were there too. Like if they were installed but not used, no point in adding an new unused package.

packages/pages/src/upgrade/pagesUpdater.ts
jwartofsky-yext: Without the \n it wasn't clearing the line properly

I can look into making that check smarter

This works on the starter repo

mkilpatrick: What do you mean by _this_? All 3 cases? I was trying in regex101.com

jwartofsky-yext: The check that is there currently works for fetch as it appears on the starter repo.

I just updated the PR to cover all cases however

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -296,14 +279,32 @@ export const replaceReactComponentsImports = (source: string): boolean => {
...
  */
 export const removeFetchImport = (source: string) => {
   const operation = async (filePath: string) => {
+    let hasReplaced = false;
     const fileContent = fs.readFileSync(filePath, "utf8");
-    if (fileContent.match(fetchImportRegex)) {
-      const modifiedContent = fileContent.replace(fetchImportRegex, "");
-      fs.writeFileSync(filePath, modifiedContent, "utf8"); // Update the file content
+    let lines = fileContent.split("\n");
+    for (let i = 0; i < lines.length; i++) {
```
mkilpatrick, line 289: Seems pretty inefficient to loop through every line of every file. Is there a better way we can get just the import lines, either through a regex or ts-morph?

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -296,14 +279,32 @@ export const replaceReactComponentsImports = (source: string): boolean => {
...
+      if (
+        lines[i].substring(0, 7).includes("import") &&
```
mkilpatrick, line 291: Someone could technically have some commented line somewhere that's like `// fetch fella import guy`

I suppose the substring part helps

packages/pages/src/upgrade/pagesUpdater.ts
mkilpatrick: You could also stop search as soon as you find a line that doesn't have `import`

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -296,14 +279,32 @@ export const replaceReactComponentsImports = (source: string): boolean => {
...
+        lines[i].includes("fetch")
```
mkilpatrick, line 292: This doesn't handle multi-line imports

```
import {
  fetch,
  foo,
} from "@yext/pages/util";
```

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -296,14 +279,32 @@ export const replaceReactComponentsImports = (source: string): boolean => {
 
 /**
  * Removes old fetch import that is no longer used
- * @param source
+ * @param source the src folder
  */
 export const removeFetchImport = (source: string) => {
```
mkilpatrick, line 284: We're at the point we really need tests for this stuff. You can do just this one for now though.

packages/pages/src/upgrade/pagesUpdater.ts
jwartofsky-yext: Upgraded to use ts-morph which handles all cases of imports

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, it, expect } from "vitest";
+import { removeFetchImport } from "./pagesUpdater.js";
+import { Project } from "ts-morph";
+import path from "path";
+import fs from "fs";
+
+describe("test pages updater steps", () => {
+  it("remove fetch imports", () => {
+    const testFilePath = "src/upgrade/test/src/templates/location.tsx";
```
mkilpatrick, line 9: Can you instead have some consts defined so you don't need ts-morph in the test and then tests all of those different fetch import cases I mentioned before?

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, it, expect } from "vitest";
+import { removeFetchImport } from "./pagesUpdater.js";
+import { Project } from "ts-morph";
+import path from "path";
+import fs from "fs";
+
+describe("test pages updater steps", () => {
+  it("remove fetch imports", () => {
```
mkilpatrick, line 8: removes

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -43,46 +119,217 @@ const processDirectoryRecursively = (currentPath: string) => {
...
+};
+
+/**
+ * Removes old fetch import that is no longer used
+ * @param root folder that contains package.json and tsconfig.json
+ */
+export const removeFetchImport = (root: string) => {
+  let hasRemoved = false;
+  const project = new Project({
+    tsConfigFilePath: path.resolve(root, "tsconfig.json"),
```
mkilpatrick, line 288: Does this read the tsconfig to find the source files? Wouldn't we only want to do the source files defined by the projectStructure?

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -43,46 +119,217 @@ const processDirectoryRecursively = (currentPath: string) => {
...
+  });
+  const sourceFiles = project.getSourceFiles();
+  for (let i = 0; i < sourceFiles.length; i++) {
```
mkilpatrick, line 291: You need to check that fetch is coming from @yext/pages/util specifically. There could be other `fetch`es from other libs we don't want to remove.

mkilpatrick: Add a test case for that too.

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -43,46 +119,217 @@ const processDirectoryRecursively = (currentPath: string) => {
...
+export const updatePackageScripts = (root: string) => {
+  const packageJsonPath = path.resolve(root, "package.json");
 
   try {
     const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf-8"));
     const scripts = packageJson.scripts;
     scripts.dev = "pages dev";
     scripts.prod = "pages prod";
     scripts["build"] = "pages build";
+    scripts["build:local"] = undefined;
```
mkilpatrick, line 332: Doing this means we need to update their buildCmd to use build instead of build:local. Do we do that already?

jwartofsky-yext: Yup
https://github.com/yext/pages/blob/main/packages/pages/src/upgrade/migrateConfig.ts#L48

packages/pages/src/upgrade/pagesUpdater.ts
jwartofsky-yext: Yes,

ts-morph wants a ts-config, but we can also set a bunch of constants instead

This gives us all of the source files we are looking for however

packages/pages/src/upgrade/pagesUpdater.test.ts
jwartofsky-yext: The consts would have to be defined in the fetch function. Do we want to use constants rather than read the ts-config?

packages/pages/src/upgrade/pagesUpdater.ts
jwartofsky-yext: Updated to use project structure instead

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -3,25 +3,54 @@ import { removeFetchImport } from "./pagesUpdater.js";
...
+    // modify test file
+    removeFetchImport(source);
+
+    // validate changes
     const afterContent = fs.readFileSync(path.resolve(testFilePath), "utf-8");
-    expect(beforeContent).toContain("fetch");
-    expect(afterContent).not.toContain("fetch");
+    expect(beforeContent).toEqual(expectedBefore);
+    expect(afterContent).toEqual(expectedAfter);
     sourceFile.delete();
```
mkilpatrick, line 54: Do you still need to save and sync?

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -3,25 +3,54 @@ import { removeFetchImport } from "./pagesUpdater.js";
...
     const testFilePath = "src/upgrade/test/src/templates/location.tsx";
-    const testProject = new Project();
+    const source = path.resolve("src/upgrade/test/src");
+    const testProject = new Project({
+      compilerOptions: {
+        jsx: typescript.JsxEmit.ReactJSX,
+        sourceRoot: source,
+      },
+    });
+    testProject.addSourceFilesAtPaths([
```
mkilpatrick, line 26: You probably only want to add the one specific file you're testing against

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -3,25 +3,54 @@ import { removeFetchImport } from "./pagesUpdater.js";
 import { Project } from "ts-morph";
 import path from "path";
 import fs from "fs";
+import typescript from "typescript";
 
 describe("test pages updater steps", () => {
-  it("remove fetch imports", () => {
+  it("removes fetch imports", () => {
```
mkilpatrick, line 9: Case for multi-line

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -280,19 +281,31 @@ export const replaceReactComponentsImports = (source: string): boolean => {
...
   let hasRemoved = false;
   const project = new Project({
-    tsConfigFilePath: path.resolve(root, "tsconfig.json"),
+    compilerOptions: {
+      jsx: typescript.JsxEmit.ReactJSX,
+      sourceRoot: source,
+    },
   });
+  project.addSourceFilesAtPaths([
+    `${source}/**/*.ts`,
```
mkilpatrick, line 295: Does `${source}/**/*.{ts,js,tsx,jsx}` work?

jwartofsky-yext: Yes, I've tested on the starter repo and verified that the fetch import in static.tsx is removed.

packages/pages/src/upgrade/pagesUpdater.test.ts
jwartofsky-yext: We actually don't need this addSourceFiles call here at all since we have one below. I'll remove it

packages/pages/src/upgrade/pagesUpdater.test.ts
jwartofsky-yext: Yes, this last delete/save is to cleanup the test file.

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -1,57 +1,56 @@
...
+): boolean {
+  const testFolder = path.resolve("src/upgrade/test");
+  const source = path.resolve(testFolder, "src");
+  const templates = path.resolve(source, "templates");
+  fs.mkdirSync(templates, {
+    recursive: true,
+  });
+  const filePath = path.resolve(templates, fileName);
+  fs.writeFileSync(filePath, beforeContent);
+  removeFetchImport(source);
```
mkilpatrick, line 19: Can you not pass ts-morph a string directly instead of having it read an actual file? It's a little wonky to be creating and deleting files for the tests.

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -1,57 +1,56 @@
...
-    expect(beforeContent).toEqual(expectedBefore);
-    expect(afterContent).toEqual(expectedAfter);
-    sourceFile.delete();
-    testProject.saveSync();
+describe("test pages updater steps", () => {
+  it("removes fetch import", () => {
+    const beforeContent = 'import { fetch } from "@yext/pages/util";';
+    expect(testFetchRemoval("location.tsx", beforeContent, "")).toEqual(true);
+  });
+  it("removes fetch import multi-line", () => {
```
mkilpatrick, line 30: empty lines between

packages/pages/src/upgrade/pagesUpdater.test.ts
jwartofsky-yext: I wasn't able to find a way to enforce line length. Multiple imports all ended up on one line.

The previous method with ts-morph was also creating and deleting files.

jwartofsky-yext: The function we are testing is also just given a source path, then it creates a ts-morph project to read it. We would have to change the function in order to pass it a string without creating a file.

mkilpatrick: That's what I was thinking we'd do, but it's not all that important. We can leave how it is.

mkilpatrick (approved):

PR #453, opened 2023-12-14
fix(plugin): allow "main" function exports and actually fail the build

mkilpatrick:
The platform allows "main" exports and reads them specifically from the /functions folder (meaning that they shouldn't be defined in functionMetadata.json...only default exports go in there). Since platform allows it, we should too.

Additionally, function errors will now properly fail the build instead of continuing on.

### DIFF
--- a/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
+++ b/packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
@@ -92,7 +92,7 @@ async function generateFunctionMetadata(
       throw (
         `${relativePath} contains an unsupported default function declaration. ` +
         "The default export function must be a named function, exported as " +
-        "`export default function foo(){}` for function `foo`"
+        "`export default function foo(){}` for function `foo`."
       );
     }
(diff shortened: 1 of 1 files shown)


packages/pages/src/vite-plugin/build/closeBundle/functionMetadata.ts
```diff
@@ -37,62 +36,71 @@ const getFunctionMetadataMap = async (
...
   } else if (Node.isFunctionDeclaration(defaultExportDeclaration)) {
     const entrypoint = defaultExportDeclaration.getName();
     if (!entrypoint) {
-      throw new Error(
+      throw (
         `${relativePath} contains an unsupported default function declaration. ` +
-          "The default export function must be a named function, exported as " +
-          "`export default function foo(){}` for function `foo`"
+        "The default export function must be a named function, exported as " +
+        "`export default function foo(){}` for function `foo`"
```
asanehisa, line 95: nit: no period here (earlier has period)

asanehisa (commented):
lgtm

asanehisa (approved):

PR #457, opened 2024-01-06
fix(plugin): polyfill node for serverless functions

mkilpatrick:
This changes how serverless components are built. Namely, we are moving back to Vite (and Rollup) instead of using esbuild directly -- effectively a revert of https://github.com/yext/pages/pull/384. The reasoning for this is to keep consistency between the main Generator plugin build and everything else we build. It also allows us to polyfill the same way using https://github.com/davidmyersdev/vite-plugin-node-polyfills, rather than using a separate esbuild plugin for polyfilling, which could have inconsistencies between the two.

Additionally, I noticed that some existing validation was actually doing anything because some previous PRs started validating a different way. I removed the useless code.


asanehisa:
two checks failing

.github/workflows/generate_markdown_files.yml
```diff
@@ -8,9 +8,6 @@ jobs:
     steps:    
       - name: "Checkout"
         uses: actions/checkout@v3
-        with:
```
asanehisa, line 11: oh without these the bot won't auto run tests after committing files (if markdown files or 3rd party notices are added by github)

jwartofsky-yext (approved):
lgtm

asanehisa (approved):

PR #459, opened 2024-01-09
feat(dev): add pageUrlField to generated configs

briantstephan:
Adds the `pageUrlField` to each feature's `entityPageSet` object in features.json and templates.json, if applicable.

Tested by adding a `pageUrlField` to a template's `TemplateConfig` and then running `pages generate features`, which made `pageUrlField` appear as a field within the feature's `entityPageSet` in features.json. Running `pages generate templates` made `pageUrlField` appear as a field within the feature's `entityPageSet` in templates.json, as expected.


packages/pages/src/common/src/feature/features.ts
```diff
@@ -93,7 +94,9 @@ export const convertTemplateConfigToFeatureConfig = (
   } else {
     featureConfig = {
       ...featureConfigBase,
-      entityPageSet: {},
+      entityPageSet: config.pageUrlField
+        ? { pageUrlField: config.pageUrlField }
```
mkilpatrick, line 98: Check with Spruce if they require not setting the key. If not, we don't need this ternary.

briantstephan: The field is actually omitted automatically if `pageUrlField` is `null` even without the ternary, so I removed it.

mkilpatrick (approved):

PR #462, opened 2024-01-10
feat(plugin): create plugin to integrate modules

dsurducan:
Allows clients to create "modules". See my test modules here: https://github.com/asanehisa/widgets-test-site/tree/main/src/modules
Hosted here: https://www.yext.com/s/4059983/yextsites/134396/branch/96516/deploys/recent?b=4059983

example use case: to allow clients to put react Search SDK on their non react pages (similar to how we used to iframe the Search Theme in the past)

All a user would need to do to have this work is:
1. create a separate folder under the "modules" subfolder in the pages-starter for each module with the name of the folder corresponding to the name of the module (ex. "`search`") which will be what is used when packaging into a umd module (ex. `search.umd.js`)

In each folder they can have:
- The main module file (ex `search.tsx`) (this will contain all of the logic and configuration that pages allows you to provide, some variable of type Module is required. If ModuleConfig is configured, it'll use the provided name.)
- `tailwind.config.ts file`
- `postcss.config.cjs file` that specifies the path of the tailwind config file they want the module to use
`module.exports = {
  plugins: {
    "tailwindcss": {config: './tailwind-config.ts' },
    "autoprefixer": {},
  },
};`
Also ensure the css code (if used) is scoped properly to the widget. It can affect code in the non-react site if not. 

This also adds response headers to the `config.yaml` such that different sites can import the widget. This can be manually changed to only allow certain sites to use the widget.

Here is a test site to see how modules appear. 
https://vibrantly-plain-lark.pgsdemo.com/test.html

To inject a module into non-react site the following is required: 
```
<div id="moduleName"> </div>
<script type="module" src="moduleUrl"></script>
```

Here's my most recent instructions / setup guide: 
https://snapdragon-hat-825.notion.site/Widgets-in-PagesJS-fc01b4e61005482e92ec4ebee462e603?pvs=4


packages/pages/src/common/src/widget/types.ts
```diff
@@ -0,0 +1,72 @@
...
+  document: T;
+  /** Additional metadata added by the toolchain */
+  __meta: {
+    /** Specifies if the data is returned in development or production mode */
+    mode: "development" | "production";
+  };
+}
+
+/**
+ * The shape of the data passed directly to the template's render function.
```
mkilpatrick, line 54: Doc incorrect. Still has "template" wording.

packages/pages/src/common/src/widget/types.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Extends the {@link TemplateProps} interface and has the additions of a path
+ * and a relativePrefixToRoot field.
+ *
+ * @public
+ */
+export interface WidgetRenderProps<T = any> extends WidgetProps<T> {
```
mkilpatrick, line 60: I don't think this is necessary, at least not yet. There's no support you added for users defining the widget path. It just goes by the name, correct?

packages/pages/src/common/src/widget/types.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @public
+ */
+export interface WidgetConfig {
+  /** The name of the widget feature. If not defined uses the widget filename (without extension) */
+  name?: string;
+}
+
+/**
+ * The shape of the data passed directly to the different template functions
```
mkilpatrick, line 39: different widgets

packages/pages/src/common/src/widget/types.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @public
+ */
+export interface WidgetProps<T = Record<string, any>> {
```
mkilpatrick, line 43: You're not supporting streams yet, so this might not be necessary. Or at least you can remove the `document: T` param.

packages/pages/src/vite-plugin/build/build.ts
```diff
@@ -21,6 +22,12 @@ export const build = async (
   return {
     name: "vite-plugin:build",
     apply: "build",
+    writeBundle: {
```
mkilpatrick, line 25: Rebase on main

packages/pages/src/vite-plugin/widgets/plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+      emptyOutDir: false,
+      outDir: outdir,
+      minify: false,
+      lib: {
+        entry: filepaths,
+        formats: ["es"],
+      },
+      rollupOptions: {
+        output: {
+          format: "umd",
```
mkilpatrick, line 65: I thought building multiple files in umd format doesn't work. Did you test with more than one widget? I'm pretty sure you'll need to loop over each widget and do a build, rather than a single build with all entries.

packages/pages/src/vite-plugin/widgets/plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+          entryFileNames: `[name]/search.umd.js`,
```
mkilpatrick, line 66: Don't hardcode this. The output filename should just be [name].js, where name will be the widget's name defined by the config or the filename if the config.name isn't defined.

packages/pages/src/vite-plugin/widgets/plugin.ts
```diff
@@ -0,0 +1,93 @@
...
+    ],
+  });
+};
+
+export const shouldBundleWidgets = (projectStructure: ProjectStructure) => {
+  const { rootFolders, subfolders } = projectStructure.config;
+  return fs.existsSync(path.join(rootFolders.source, subfolders.widgets));
+};
+
+const getFilePathName = (filepath: string) => {
```
mkilpatrick, line 89: This should use the widget config.name. I recommend creating an internal type for the widget config which handles that translation and then using it here. See `convertTemplateConfigToTemplateConfigInternal` in `packages/pages/src/common/src/template/internal/types.ts`

package-lock.json
```diff
@@ -0,0 +1,12425 @@
+{
```
mkilpatrick, line 1: This file shouldn't exist. You probably used npm instead of pnpm at once point. You can delete this.

packages/pages/src/vite-plugin/widgets/plugin.ts
dsurducan: I tried it with two widgets and it packaged them into two separate umd modules but I didn't try serving it locally so maybe I should do that too

packages/pages/src/common/src/template/internal/types.ts
```diff
@@ -71,6 +72,14 @@ export interface TemplateConfigInternal {
   slugField?: string;
 }
 
+/**
+ * The exported `config` function's definition.
+ */
+export interface WidgetConfigInternal {
```
mkilpatrick, line 78: Can you make a src/widgets/internal/types.ts file instead of putting it in the template one?

packages/pages/src/common/src/widget/types.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @public
+ */
+export interface WidgetConfig {
+  /** The name of the widget feature. If not defined uses the widget filename (without extension) */
+  name?: string;
+}
+
+/**
+ * The shape of the data passed directly to the different widget functions
```
mkilpatrick, line 37: add period

pnpm-lock.yaml
```diff
@@ -7654,7 +7663,7 @@ packages:
       - rollup
     dev: false
 
-  /vite@5.0.2:
+  /vite@5.0.2(@types/node@20.10.6):
```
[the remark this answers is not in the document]
asanehisa: ignoring for now

packages/pages/src/common/src/module/types.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * The type definition for the module's default function.
+ *
+ * @public
+ */
+export type Module = () => React.JSX.Element;
+
+/**
+ * The exported `config` function's definition.
```
mkilpatrick, line 11: Is the "function's" part a copy/paste from serverless functions? I would change this to be more meaningful for a user as this is an exposed type: "The configuration options for a Module."

packages/pages/src/common/src/module/types.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @public
+ */
+export interface ModuleConfig {
+  /** The name of the module feature. If not defined uses the module filename (without extension) */
```
mkilpatrick, line 16: "feature" seems weird here. Just the "name of the module" is fine.

packages/pages/src/common/src/module/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  name?: string;
+}
+
+/**
+ * The shape of the data passed directly to the different module functions.
```
mkilpatrick, line 21: The shape of the data passed directly to the module's default export.

packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -207,4 +208,115 @@ export default class SourceFileParser {
   getAllText(): string {
     return this.sourceFile.getFullText();
   }
+
+  /**
+   * For example, we can do getVariablePropertyByType("ModuleConfig", "name")
```
mkilpatrick, line 213: I think this will be brittle to use. There's nothing forcing a user to actual type their config. They should, but they don't have to. So if we're solely relying on the type to find the config then this will fail. The config name is specific though, so we can just look for that.

packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -207,4 +208,115 @@ export default class SourceFileParser {
...
+          const value = variableProperty
+            .getFirstChildByKind(SyntaxKind.StringLiteral)
+            ?.getText();
+          if (value) {
+            return value;
+          } else {
+            return variableProperty;
+          }
+        } else {
+          console.log(type + " does not have a '" + property + "' property.");
```
mkilpatrick, line 249: Use string interpolation. It's easier to read.

packages/pages/src/common/src/parsers/sourceFileParser.ts
```diff
@@ -207,4 +208,115 @@ export default class SourceFileParser {
...
+          return;
+        }
+      } else {
+        console.log(
+          "Initializer for " + type + "variable is not an object literal."
```
mkilpatrick, line 254: Same

packages/pages/src/common/src/parsers/sourceFileParser.ts
mkilpatrick, line 249: Do we just console.log in other places or do we error? I think we'd want to error out and fail the build.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,173 @@
...
+import { processEnvVariables } from "../../util/processEnvVariables.js";
+import { nodePolyfills } from "vite-plugin-node-polyfills";
+import pc from "picocolors";
+import { addResponseHeadersToConfigYaml } from "../../util/editConfigYaml.js";
+import SourceFileParser, {
+  createTsMorphProject,
+} from "../../common/src/parsers/sourceFileParser.js";
+
+const wrappedCode = (moduleName: string, containerName: string): string => {
+  return `\nconst moduleContainerForBuildUseOnly = document.getElementById('${containerName}');
```
mkilpatrick, line 16: This would be easier to read if formatted using triple backticks.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,173 @@
...
+  <React.StrictMode>
+    <${moduleName}/>
+  </React.StrictMode>,
+  moduleContainerForBuildUseOnly
+);
+  `;
+};
+
+const moduleResponseHeader = {
+  pathPattern: "^modules/.*",
```
mkilpatrick, line 31: I'm not sure if this needs to configurable. I can imagine someone wanting to change this somehow. Ask in the project channel.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,173 @@
...
+};
+
+export const buildModules = async (
+  projectStructure: ProjectStructure
+): Promise<void> => {
+  if (!shouldBundleModules(projectStructure)) {
+    return;
+  }
+
+  addResponseHeadersToConfigYaml(projectStructure, moduleResponseHeader);
```
mkilpatrick, line 43: Also verify if we should be updating the config.yaml automatically.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,173 @@
...
+      envPrefix: envVarConfig.envVarPrefix,
+      resolve: {
+        conditions: ["worker", "webworker"],
+      },
+      publicDir: false,
+      css: {
+        postcss: path.join(
+          rootFolders.source,
+          subfolders.modules,
+          `${moduleName}/postcss.config.cjs`
```
mkilpatrick, line 88: A user could technically use .cjs or .mjs. or .js. or .ts.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,173 @@
...
+      const filepath = path.resolve(f);
+      const moduleName = getModuleName(filepath);
+      filepaths[moduleName] = filepath;
+    });
+
+  const logger = createLogger();
+  const loggerInfo = logger.info;
+
+  for (const [moduleName, modulePath] of Object.entries(filepaths)) {
+    const index = addExtraModuleCode(modulePath, moduleName);
```
mkilpatrick, line 66: Why do we need to add the moduleCode and then remove it?

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,173 @@
...
+  if (declaration === undefined) {
+    throw new Error(`Cannot find variable Module in ${modulePath}`);
+  }
+  const moduleName = declaration.getName();
+  const insertIndex = sfp.getEndPos();
+  const afterInsertIndex = sfp.insertStatement(
+    wrappedCode(moduleName, name),
+    insertIndex
+  );
+  sfp.addReactImports();
```
mkilpatrick, line 158: Do we actually need this? Is the react stuff because of the ReactDOM.render we're using? Or is it just if the user doesn't add it to their template? If the latter I don't think we need to do it for them automatically.

packages/pages/src/vite-plugin/modules/plugin.ts
mkilpatrick, line 43: We'll definitely need to test this manually against a reverse proxy setup, meaning the module is hosted at a RP domain.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: >  Is the react stuff because of the ReactDOM.render we're using?


Yes! We do need it for that.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Ah I remove it so it doesn't show up to user, just added for the bundling into umd file

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Confirmed with Aaron.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Ok! Do we want to support all of them or just .js?

packages/pages/src/common/src/parsers/sourceFileParser.ts
asanehisa: Oh I see. By "The config name is specific" do you mean, they'll always name the const `config` and if so why / how can we guarantee this?

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Added support for all!

packages/pages/src/util/editConfigYaml.ts
```diff
@@ -33,4 +41,12 @@ export const addResponseHeadersToConfigYaml = (
   }
 
   fs.writeFileSync(configYamlPath, YAML.stringify(yamlDoc));
+
+  if (comment) {
+    const yaml = YAML.parseDocument(fs.readFileSync(configYamlPath, "utf-8"));
+    const responseHeader: any = yaml.get("responseHeaders");
+    responseHeader.comment = comment;
```
mkilpatrick, line 48: Doesn't this add the comment to the top of the responseHeaders? If you have multiple modules it's going to have multiple comments in a row saying the same thing. I think we want it above each new entry we auto add. Is it possible to add nodes to the yaml doc so you can add comments at the same time or can you only add via json above?

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -56,14 +64,24 @@ export const buildModules = async (
...
     });
 
   const logger = createLogger();
   const loggerInfo = logger.info;
 
-  for (const [moduleName, modulePath] of Object.entries(filepaths)) {
-    const index = addExtraModuleCode(modulePath, moduleName);
+  if (tailwindBaseExists()) {
+    // TODO add link to recommended implementation for user.
+    loggerInfo(
```
mkilpatrick, line 76: Use logWarning from packages/pages/src/util/logError.ts

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -72,6 +90,18 @@ export const buildModules = async (
...
+    // For each module, add response header to config.yaml.
+    // Users can manually adjust headerKey and headerValue in their config.yaml.
+    // As long as pathPattern matches, it won't be overwitten.
+    addResponseHeadersToConfigYaml(
+      projectStructure,
+      {
+        pathPattern: `^modules/${moduleName}.*`,
+        ...moduleResponseHeaderProps,
+      },
+      " This response header allows access to your modules from other sites"
```
mkilpatrick, line 102: Remove space before `This`

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -171,3 +205,60 @@ const removeAddedModuleCode = (modulePath: string, index: number) => {
...
+  }
+
+  for (const extension of postcssExtensions) {
+    const filePath = path.join(`postcss.config.${extension}`);
+    if (fs.existsSync(filePath)) {
+      return filePath;
+    }
+  }
+
+  throw new Error(`Cannot find a postcss.config file for ${filename}`);
```
mkilpatrick, line 242: A user doesn't need to use TW/PostCss at all.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -171,3 +205,60 @@ const removeAddedModuleCode = (modulePath: string, index: number) => {
...
+ * Returns the postcss.config filepath if it exists in module.
+ * Else returns the root postcss.config filepath.
+ * If there is none, throws error.
+ *
+ * @param rootFolders
+ * @param subfolders
+ * @param filename of module
+ * @returns string
+ */
+const getPostcssConfigFilepath = (
```
mkilpatrick, line 219: getPostCssConfigFilepath

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -171,3 +205,60 @@ const removeAddedModuleCode = (modulePath: string, index: number) => {
...
+  rootFolders: any,
+  subfolders: any,
+  filename: string
+): string => {
+  for (const extension of postcssExtensions) {
```
mkilpatrick, line 224: You can just glob for the file and use the extension selector - https://stackoverflow.com/questions/43118216/how-to-include-multiple-file-extensions-with-glob-sync-in-nodejs

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -171,3 +205,60 @@ const removeAddedModuleCode = (modulePath: string, index: number) => {
...
+      return filePath;
+    }
+  }
+
+  throw new Error(`Cannot find a postcss.config file for ${filename}`);
+};
+
+/**
+ * Looks at all css files in src and returns true
+ * if there is an unwrapped tailwind base
```
mkilpatrick, line 247: period

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -171,3 +205,60 @@ const removeAddedModuleCode = (modulePath: string, index: number) => {
...
+ *
+ * @returns boolean
+ */
+const tailwindBaseExists = (): boolean => {
+  const files = glob.sync("./src/**/*.css");
+  for (const filePath of files) {
+    try {
+      const data = fs.readFileSync(filePath, "utf8");
+      if (data.includes(`@tailwind base`) && !data.includes(`.tailwind {`)) {
```
mkilpatrick, line 256: You can't assume someone is wrapping in .tailwind can you? Like if I have 2 different modules and I want to put both on the same page then they'd need to have the base namespaced differently, right?

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -171,3 +205,60 @@ const removeAddedModuleCode = (modulePath: string, index: number) => {
...
+        return true;
+      }
+    } catch (err) {
+      throw new Error(`Cannot read file at ${filePath} due to: ${err}`);
```
mkilpatrick, line 260: I think we ignore the error and return false. A user could have some weird file we effectively just want to skip over. Comment about this in the catch though because it's intentional we're not handling.

packages/pages/src/common/src/parsers/sourceFileParser.ts
mkilpatrick: Yes. We can guarantee it because it's convention for how to use a module. I mean it's technically not required because name is the only property which is optional, so you'd have to look for an export named config, otherwise it doesn't exist and we end up using the filename.

packages/pages/src/util/editConfigYaml.ts
asanehisa: This just adds one comment! 

I tried to add it via nodes but it got messier than I'd prefer and required weird formatting due to the fact that responseHeaders is an array of objects. Therefore in the end I decided keeping the simple JSON method was cleaner + had less room for error.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Ah oops.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: You're right. Changed it to check if it is inside any rule.

packages/pages/src/util/editConfigYaml.ts
mkilpatrick: The issue is that there could be other responseHeaders the user adds.

Also, the text is "This response header allows access to your modules from other sites". Let's verify with Aaron that we actually want this at all. If so then we should change the text.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Done + tested.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: I don't think that is possible in this code. I get 
```
This expression is not callable.
Type 'String' has no call signatures.
```

mkilpatrick: Woops, triple backticks is Go 😅.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -8,12 +8,19 @@ import { processEnvVariables } from "../../util/processEnvVariables.js";
...
   createTsMorphProject,
 } from "../../common/src/parsers/sourceFileParser.js";
+import { logWarning } from "../../util/logError.js";
+import postcss from "postcss";
+import nested from "postcss-nested";
 
 const wrappedCode = (moduleName: string, containerName: string): string => {
-  return `\nconst moduleContainerForBuildUseOnly = document.getElementById('${containerName}');
+  return `
+// The following code is added during build and removed after build is completed.
```
mkilpatrick, line 22: You should be able to use rollup's `outro` option to handle this instead of modifying the original file.

mkilpatrick: You can use `intro` for the react imports.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -171,3 +203,71 @@ const removeAddedModuleCode = (modulePath: string, index: number) => {
   sfp.removeUnusedImports();
   sfp.save();
 };
+
+/**
+ * Returns the postcss.config filepath if it exists in module.
```
mkilpatrick, line 208: in the module

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+      css: {
+        postcss: getPostCssConfigFilepath(
+          rootFolders,
+          subfolders,
+          fileInfo.name
+        ),
+      },
+      esbuild: {
+        logOverride: {
+          "css-syntax-error": "silent",
```
mkilpatrick, line 122: I don't think we want this. Also esbuild shouldn't be used at all so I don't think we need the whole esbuild block.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -0,0 +1,273 @@
...
+        },
+      },
+      build: {
+        chunkSizeWarningLimit: 2000,
```
mkilpatrick, line 126: remove

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Was trying to suppress this warning: 
```
warnings when minifying css:
▲ [WARNING] Unexpected "{" [css-syntax-error]

    <stdin>:630:1:
      630 │  {
          ╵  ^
```
The exact line it is warning you about shifts, but is always there when minimize: true.

packages/pages/src/vite-plugin/modules/plugin.ts
mkilpatrick: Oic. You can keep it then.

packages/pages/src/vite-plugin/modules/plugin.ts
asanehisa: Oh that is really weird. It did properly suppress the warning when running npm run prod

asanehisa: https://github.com/vitejs/vite/issues/9196
https://github.com/vitejs/vite/pull/9210

asanehisa: Can also add a catch error in the logger for this instead if we prefer.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -117,13 +112,7 @@ export const buildModules = async (
           fileInfo.name
         ),
       },
-      esbuild: {
-        logOverride: {
-          "css-syntax-error": "silent",
```
mkilpatrick: We can keep this if necessary. We're you still getting the warnings after I couldn't reproduce yesterday?

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -117,13 +112,7 @@ export const buildModules = async (
...
-        },
-      },
       build: {
-        chunkSizeWarningLimit: 2000,
```
mkilpatrick: I responded to the last comment that we can keep this since the default is 500. Didn't realize that.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -148,10 +138,42 @@ export const buildModules = async (
...
   }
 };
 
+export default function addWrappedCodePlugin(
+  path: string,
+  moduleName: string
+): Plugin {
+  return {
+    name: "wrapped-code-plugin",
+    apply: "build",
```
mkilpatrick, line 150: Try `pre` based on our discussion.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -19,19 +20,14 @@ import nested from "postcss-nested";
 
 const wrappedCode = (moduleName: string, containerName: string): string => {
```
mkilpatrick, line 21: Can you move this down closer to the functions that are using it?

mkilpatrick (dismissed):

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -8,9 +8,6 @@ import { processEnvVariables } from "../../util/processEnvVariables.js";
 import { nodePolyfills } from "vite-plugin-node-polyfills";
 import pc from "picocolors";
 import { addResponseHeadersToConfigYaml } from "../../util/editConfigYaml.js";
-// https://github.com/dolanmiu/vite-plugin-scope-tailwind/issues/5
-// @ts-expect-error due to type any
-import scopeTailwind from "vite-plugin-scope-tailwind";
```
mkilpatrick: Remove from package.json

mkilpatrick (approved):

PR #463, opened 2024-01-10
feat(dev): pass slugField to generate-test-data if provided

briantstephan:
This adds slugField to entityPageSet to that it can be read by the CLI in dynamic dev mode.


packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -52,16 +55,30 @@ export const generateTestDataForSlug = async (
...
+    );
+    const streamConfig = convertTemplateConfigToStreamConfig(
+      templateModuleInternal.config
+    );
+    const featuresConfigForEntityPages: FeaturesConfig = {
+      features: "entityPageSet" in featureConfig ? [featureConfig] : [],
+      streams: streamConfig ? [streamConfig] : [],
+    };
+    const args = getCommonArgs(featuresConfigForEntityPages, projectStructure);
+    args.push("--slug", templateModuleInternal.config.slugField ?? slug);
```
mkilpatrick, line 72: `slug` represents the slug value, which is coming from the url request. It's not "what slug field to use." We'll need new functionality in the CLI function.

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -52,16 +55,30 @@ export const generateTestDataForSlug = async (
...
-    features: featuresConfig.features.filter((f) => "entityPageSet" in f),
-    streams: featuresConfig.streams,
-  };
-  const args = getCommonArgs(featuresConfigForEntityPages, projectStructure);
-  args.push("--slug", slug);
 
-  const parsedData = await spawnTestDataCommand(stdout, "yext", args);
-  return getDocumentByLocale(parsedData, locale);
+  const allParsedData: any[] = [];
+  templateModuleCollection.forEach(async (templateModuleInternal) => {
```
mkilpatrick, line 60: You shouldn't need to do this. You can just pass the slug off of the templateModuleInternal into generateTestDataForPage

briantstephan: Would that be better than simply putting it into entityPageSet and then allowing the CLI to pick it up from there?

mkilpatrick:
Did you forget to stage something? The new field isn't passed to generateTestData at all.

mkilpatrick (approved):

asanehisa (approved):

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -60,6 +60,25 @@ export const generateTestDataForSlug = async (
   const args = getCommonArgs(featuresConfigForEntityPages, projectStructure);
   args.push("--slug", slug);
 
+  const nonStaticFeatureNames = new Set<string>(
```
mkilpatrick, line 63: I don't think this is necessary. You can just loop over the templateModuleCollection and add slugField if it's defined, otherwise add entity.slug. It doesn't matter if we loop over static templates too.

packages/pages/src/dev/server/ssr/getLocalData.ts
```diff
@@ -74,21 +74,23 @@ export const getLocalDataManifest = async (
...
+      vite,
+      async (t) => featureName === t.config.name,
+      templateFilepaths
+    );
+
     const uid = data.uid?.toString();
     const entityId = data.id?.toString();
-    const slug = data.slug?.toString();
+    const slugField = templateModuleInternal?.config?.slugField;
+    const slug = slugField ? data[slugField] : data.slug?.toString();
```
mkilpatrick, line 86: Do you need to .toString() data[slugField]? Or is toString not necessary on data.slug?

briantstephan: We shouldn't to toString on `data[slugField]` so that it is properly passed as undefined in the case that it isn't there.

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -60,6 +60,23 @@ export const generateTestDataForSlug = async (
...
   args.push("--slug", slug);
 
+  const slugFields = new Set<string>();
+  let shouldAddDefaultSlugField = false;
+  templateModuleCollection.forEach((templateModule) => {
+    const slugField = templateModule?.config?.slugField;
+    if (slugField) {
+      slugFields.add(`entity.${slugField}`);
+    } else {
+      shouldAddDefaultSlugField = true;
```
mkilpatrick, line 70: This else could just add entity.slug and then you can can rid of lines 73-78 (and always do the args.push now).

mkilpatrick (approved):

PR #465, opened 2024-01-12
feat(dev): update serverless function types in upgrade script

mkilpatrick:


asanehisa (dismissed):

packages/pages/src/upgrade/pagesUpdater.test.ts
```diff
@@ -1,31 +1,34 @@
...
+  updateFunction: (input: string) => void
 ): boolean {
   const testFolder = path.resolve("src/upgrade/test");
   const source = path.resolve(testFolder, "src");
   const templates = path.resolve(source, "templates");
   fs.mkdirSync(templates, {
     recursive: true,
   });
-  const filePath = path.resolve(templates, fileName);
+  const filePath = path.resolve(templates, "foo.tsx");
```
jwartofsky-yext, line 20: Why is foo.tsx here? Is that a placeholder?

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -407,3 +407,138 @@ export const checkNodeVersion = () => {
...
+      }
+      const namedImports = importDeclaration.getNamedImports();
+      for (let k = 0; k < namedImports.length; k++) {
+        const namedImport = namedImports[k];
+        if (!oldServerlessFunctionTypes.includes(namedImport.getName())) {
+          continue;
+        }
+
+        switch (namedImport.getName()) {
+          case "SitesHttpRequest":
```
jwartofsky-yext, line 463: I think a map would be a more appropriate structure here, then you could have a map of 'SitesHttpResponse' -> 'PagesHttpResponse' instead of needing a switch case for each replacement.

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -407,3 +407,138 @@ export const checkNodeVersion = () => {
...
+    const functions = sourceFile.getFunctions();
+    for (let j = 0; j < functions.length; j++) {
+      // update parameter types
+      functions[j].getParameters().forEach((param) => {
+        const paramType = param.getType().getText();
+        if (oldServerlessFunctionRequestTypes.includes(paramType)) {
+          param.removeType();
+
+          switch (paramType) {
+            case "SitesHttpRequest":
```
jwartofsky-yext, line 501: same map comment

packages/pages/src/upgrade/pagesUpdater.test.ts
mkilpatrick: It's just a random filename. It doesn't matter what it is. Before you had every test take in a filename, but it doesn't actually matter what it's called because it creates and deletes the files per test.

PR #469, opened 2024-01-17
fix(dev): index page urls using `--no-prod-url`

mkilpatrick:
Prior to this change the index page urls would show the slug when using local urls which was confusing since the actual urls have the format /[template-name]/[external-id] (and optional ?locale= query param if not ‘en’). We now show the actual url.


jwartofsky-yext (approved):
lgtm

PR #471, opened 2024-01-18
fix(dev): filter out static templates when determining slug fields

mkilpatrick:
Adds a templateType option to TemplateModuleInternal to make it easier to filter by certain template types.

### DIFF
--- a/packages/pages/src/common/src/template/internal/types.ts
+++ b/packages/pages/src/common/src/template/internal/types.ts
@@ -45,8 +45,6 @@ export interface TemplateModuleInternal<
   render?: Render<U>;
   /** The exported default template function */
   default?: Template<U>;
-  /** The type of template */
-  templateType: "entity" | "static";
 }
 
 /**
@@ -74,6 +72,8 @@ export interface TemplateConfigInternal {
   pageUrlField?: string;
   /** The field to use as the slug for dynamic dev mode */
   slugField?: string;
+  /** The type of template */
+  templateType: "entity" | "static";
 }
 
 /**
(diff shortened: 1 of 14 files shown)


asanehisa (dismissed):

packages/pages/src/common/src/template/internal/types.ts
```diff
@@ -133,23 +135,35 @@ export const parse = (
   };
 };
 
+export const isStaticTemplateConfig = (
```
briantstephan, line 138: Could you import this from packages/pages/src/common/src/feature/features.ts instead of making a copy?

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -52,6 +52,13 @@ export const generateTestDataForSlug = async (
     vite,
     templateFilepaths
   );
+
+  for (const [key, templateModule] of templateModuleCollection.entries()) {
+    if (templateModule.templateType !== "entity") {
+      templateModuleCollection.delete(key);
+    }
+  }
```
briantstephan, line 60: If readability is a concern here, you could do something like [this](https://stackoverflow.com/a/53065133), although it's technically less efficient (shouldn't matter for the map sizes we're dealing with here though).

mkilpatrick: Ack. The map way has just as many lines and really isn't easy to read, also at the cost of performance.

briantstephan (approved):

PR #475, opened 2024-01-23
fix(dev/plugin): static templates not working for alternate locales

mkilpatrick:
The issue was originally found due to builds failing trying to load localData because of some validation to test for slug uniqueness. localData should never be checked in, so this validation was removed completely.

Additionally, it was noticed that the static template urls were not properly rendering on the index page when multiple locales were used due to a bug in the mapping logic.

Static templates also now properly work in true "dynamic" mode where the template and document are found based solely on the slug.

Apart from the fixes, there is a fair bit of refactoring to streamline codepaths. We now also only find localData documents once instead of twice.


playground/locations-site/sites-config/features.json
```diff
@@ -21,13 +21,25 @@
...
+        "meta",
+        "address",
+        "slug"
+      ],
       "localization": {
-        "locales": ["en"],
+        "locales": [
+          "en"
+        ],
         "primary": false
```
jwartofsky-yext, line 39: Not necessarily part of this PR, but I thought we were removing primary from localization?

mkilpatrick: We did, but it doesn't hurt anything and this PR isn't about cleaning up the playgrounds.

jwartofsky-yext (approved):
lgtm

briantstephan (approved):

PR #476, opened 2024-01-24
fix(plugin): serverless function shared components

mkilpatrick:
Bundling all serverless components in a single vite/rollup step does not allow you to create single outputs per entry. Chunks are always created for shared code. This changes the logic to perform a build per serverless function so that any shared code gets bundled in each corresponding output mod.ts.


asanehisa (approved):

briantstephan (approved):

jwartofsky-yext (approved):
lgtm

PR #477, opened 2024-01-25
fix(upgrade): name of serving config in config.yaml

mkilpatrick:
The keys need to be changed from
```
reverseProxy:
  displayUrlPrefix: foo
```
to
```
serving:
  reverseProxyPrefix: foo
```


jwartofsky-yext (approved):
lgtm

PR #478, opened 2024-01-25
feat(upgrade): move imports from components-tsx-maps/geo

mkilpatrick:
Moves all named imports from both `@yext/components-tsx-maps` and `@yext/components-tsx-geo` to `@yext/pages-components`. Also removes those imports from package.json.

### DIFF
--- a/packages/pages/src/upgrade/pagesUpdater.ts
+++ b/packages/pages/src/upgrade/pagesUpdater.ts
@@ -143,6 +143,8 @@ function processDirectoryRecursively(
 export const updateToUsePagesComponents = async (source: string) => {
   await removePackageDependency("@yext/sites-components");
   await removePackageDependency("@yext/react-components");
+  await removePackageDependency("@yext/components-tsx-maps");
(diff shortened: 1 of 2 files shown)


packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -184,7 +188,7 @@ export const updatePagesJSToCurrentVersion = async () => {
  */
 export const updateDevDependencies = async () => {
   await updatePackageDependency("@vitejs/plugin-react", null);
-  await updatePackageDependency("vite", "^5.0.10");
+  await updatePackageDependency("vite", null);
```
asanehisa, line 191: not sure why this changed to null, besides that lgtm

mkilpatrick: It's technically out of scope of my changes but it's a small enough one. It's so the latest version of Vite is pulled instead of a specific one.

asanehisa (approved):

PR #480, opened 2024-01-25
fix(upgrade): check for Markdown before updating imports

mkilpatrick:
The markdown warning check was never happening because we rewrote the imports that it's looking for before this check was executed.


asanehisa (approved):

jwartofsky-yext (approved):
lgtm

PR #486, opened 2024-02-02
fix(dev)!: rework _server.tsx

mkilpatrick:
This reworks the exports of _server.tsx. According to [this](https://github.com/vitejs/vite/pull/15345#issuecomment-1855550194) transformIndexHtml should be called on the index.html (which is provided automatically) and not the entire result of calling render (after the server code is injected). It hasn’t actually posed a problem but I’d like to get ahead of it and make sure we’re doing it the right way. It really only affects local dev but it’s breaking since what’s exported from _server.tsx has changed.

_server.tsx is now expected to export 3 things:
```
export const render = async (pageContext: PageContext<any>) => {
  const { Page, pageProps } = pageContext;

  return ReactDOMServer.renderToString(<Page {...pageProps} />);
};

export const replacementTag = "<!--YEXT-SERVER-->";

export const indexHtml = `<!DOCTYPE html>
    <html lang="<!--app-lang-->">
      <head></head>
      <body>
        <div id="reactele">${replacementTag}</div>
      </body>
    </html>`;
```

Specifically, `indexHtml` is now split out from the server `render`. It also requires a `replacementTag` to know where the server html should be injected into the indexHtml.

### DIFF
--- a/packages/pages/etc/pages.api.md
+++ b/packages/pages/etc/pages.api.md
@@ -10,6 +10,11 @@ import { default as React_2 } from "react";
 export type Attributes = Record<string, string>;
 
 // @internal
+export interface ClientRenderTemplate {
+  render(pageContext: PageContext<any>): Promise<string>;
+}
+
+// @internal
 export interface ClientServerRenderTemplates {
   clientRenderTemplatePath: string;
   isCustomRenderTemplate: boolean;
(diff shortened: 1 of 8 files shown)


asanehisa:
lgtm but some tests failing

PR #494, opened 2024-03-12
feat(scaffold): scaffold modules

asanehisa:


packages/pages/src/scaffold/modules/generate.ts
```diff
@@ -0,0 +1,120 @@
...
+  );
+  while (moduleExists(moduleName)) {
+    console.error(
+      `Error: Module "${moduleName}" already exists or is invalid.`
+    );
+    moduleName = await query("Please enter a different name for your Module: ");
+  }
+
+  // Ensure user input is yes, no, y, or n (ignores capitalization)
+  const useTailwind: string = await query(
```
mkilpatrick, line 35: Use prompts package. See autoInit.ts

packages/pages/src/scaffold/modules/generate.ts
```diff
@@ -0,0 +1,105 @@
...
+    {
+      type: "confirm",
+      name: "useTailwind",
+      message: "Would you like to use Tailwind CSS?",
+      initial: true,
+    },
+  ];
+  const response = await prompts(questions);
+
+  const modulePath = path.join("src", "modules", response.moduleName);
```
mkilpatrick, line 39: This should use the projectStructure

packages/pages/src/scaffold/modules/generate.ts
```diff
@@ -0,0 +1,105 @@
...
+  process.removeListener("SIGINT", handleCancel);
+  console.log(
+    `\nModule "${response.moduleName}" created successfully at ${modulePath}`
+  );
+};
+
+// Ensures moduleName isn't used already in a modulePath and the name starts with
+// an alphabetic character
+const validateModuleName = (moduleName: string): boolean => {
+  const modulePath = path.join("src", "modules", moduleName);
```
mkilpatrick, line 76: Same projectStructure comment throughout this file

packages/pages/src/scaffold/modules/templates.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  return `import { Module, ModuleConfig, ModuleProps } from "@yext/pages/*";
+import { AnalyticsProvider } from "@yext/pages-components";
+import "./index.css";
+
+const templateData: ModuleProps = {
+  document: {
+    businessId: "REPLACE_ME",
+    siteId: "REPLACE_ME",
+    apiKey: "REPLACE_ME",
```
mkilpatrick, line 17: Remove, this isn't part of document. The apiKey I was referring to is a top-level prop of AnalyticsProvider once my update is out.

Here are the updated props:
```
constructor(
    private apiKey: string,
    private defaultCurrency: string,
    private templateData: TemplateProps,
    requireOptIn?: boolean | undefined,
    private productionDomains: string[] = [],
    disableSessionTracking?: boolean | undefined,
    private enableDebugging: boolean = false
  )
```

We should collect (aka set "REPLACE_ME"s):
* apiKey
* defaultCurrency
* templateData (with this data of course)
* productionDomains

packages/pages/src/scaffold/modules/templates.ts
```diff
@@ -0,0 +1,85 @@
...
+    __: {
+      name: "${moduleName}",
+      staticPage: true
+    }
+  }, 
+  __meta: {
```
mkilpatrick, line 23: I don't think this is used for anything.

packages/pages/src/scaffold/modules/templates.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+}
+
+export const config: ModuleConfig = {
+  name: "${moduleName}"
+}
+
+const ${formattedModuleName}: Module = () => {
+  return(
+    <AnalyticsProvider templateData={templateData}>
```
mkilpatrick, line 34: This is where you'd add the props of the previous comment

packages/pages/src/scaffold/modules/templates.ts
```diff
@@ -0,0 +1,85 @@
...
+@tailwind utilities;  
+`
+    : ``;
+};
+
+export const tailwindCode = () => {
+  return `import type { Config } from 'tailwindcss';
+
+export default {
+  content: ["./**/*.{js,ts,tsx}"],
```
mkilpatrick, line 78: Just checking that this stays scoped to its current folder.

packages/pages/src/scaffold/modules/templates.ts
asanehisa: Hmmm I was getting some error w/o __meta, let me recreate

packages/pages/src/scaffold/modules/templates.ts
asanehisa: Ah gotcha. Should this item not be released until after ur analytics then? Since apiKey doesn't exist rn?

packages/pages/src/scaffold/modules/templates.ts
asanehisa: Ah I get an error b/c `Property '__meta' is required in type 'TemplateProps<Record<string, any>>'` and `AnalyticsProviderProps.templateData: TemplateProps<Record<string, any>>`

packages/pages/src/scaffold/modules/templates.ts
mkilpatrick: You can just add them. It fails TS validation but that's only in your editor. It still builds fine. Try it out.

packages/pages/src/scaffold/modules/templates.ts
mkilpatrick: Ah ok. You can leave it then.

packages/pages/src/scaffold/modules/templates.ts
asanehisa: Yup tested + confirms works fine, just fails IDE TS validation.

packages/pages/src/scaffold/scaffold.ts
```diff
@@ -0,0 +1,12 @@
+import { Command } from "commander";
+import { modulesCommand } from "./modules/modules.js";
+
+export const scaffoldCommand = (program: Command) => {
+  const scaffold = program
+    .command("scaffold")
```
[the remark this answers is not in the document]
asanehisa: With the generate command, the subcommands are the one that have .option(--scope), not the generate command. In this case, would it be better to have the subcommand, modulesCommand, have the .option(--scope)

packages/pages/src/scaffold/modules/generate.ts
```diff
@@ -110,9 +113,12 @@ function handleCancel(moduleName: string, projectStructure: ProjectStructure) {
     fs.rmdirSync(modulePath);
   }
 
-  const modulesPath = path.join("src", "modules");
-  if (fs.readdirSync(modulesPath).length === 0) {
-    fs.rmdirSync(modulesPath);
+  const moduleDirPath = path.join(
```
mkilpatrick, line 116: Won't this delete all modules if you cancel the scaffold? We'd only want to clean up THE module folders that were created by this command.

packages/pages/src/scaffold/scaffold.ts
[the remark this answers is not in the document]
mkilpatrick: If you add it to scaffold does it bubble down to the subcommands? Any scaffold command will most likely need to be scoped.

packages/pages/src/scaffold/modules/generate.ts
asanehisa: In this case, it only removes the `modules/` dir if it is empty. But yeah I can remove it + have it only delete the `modules/` dir if was created by this command.

packages/pages/src/scaffold/scaffold.ts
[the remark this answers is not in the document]
asanehisa: Talked offline + resolved

packages/pages/src/scaffold/modules/generate.ts
```diff
@@ -0,0 +1,125 @@
...
+    {
+      type: "confirm",
+      name: "useTailwind",
+      message: "Would you like to use Tailwind CSS?",
+      initial: true,
+    },
+  ];
+  const response = await prompts(questions);
+
+  const modulePath = path.join(
```
mkilpatrick, line 43: Add a function getModulesPath to structure.ts (see others at the bottom) and then use it here. You could also have it optionally take in a moduleName which will return it with the module in the path (but still scoped).

getModulesPath() => src/modules/[scope] (if scope exists)
getMoudlesPath(myModule) => src/modules/[scope]/myModules

asanehisa: Great idea. Thanks

mkilpatrick (approved):

PR #496, opened 2024-03-14
fix: removes warning about `--experimental-loader`

asanehisa:
This error shows up consistently in yext-pages-1.1.0-beta.0.tgz

```
(node:8401) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("/Users/asanehisa/pages/packages/pages/dist/bin/loader.js", pathToFileURL("./"));'
```

If we don't want to use register(), we can mute the error instead?


asanehisa:
Draft rn. Windows doesn't like the file path right now.

PR #497, opened 2024-03-14
feat(dev): add widgets to dev mode

jwartofsky-yext:
This change allows the user to load widgets in dev mode.

The user must use the --module flag in order for it to work.

To load a module named "search", the user can use "npx pages dev --module search". Can also use the folder name instead of the module.config.name to select a module to start. If there is no module.config.name, then the folder name must be used instead.


packages/pages/src/common/src/module/internal/_server.tsx
```diff
@@ -0,0 +1,21 @@
+// Environment: server
+
```
mkilpatrick, line 2: Are these any different from the template ones? Doesn't look like it. We could just reuse those. But in reality we shouldn't put things on the page the user hasn't coded. All we really need is a div that mounts to the name of the module.

```
<div id="moduleName"><!--YEXT-SERVER--></div>
```

packages/pages/src/common/src/module/internal/validateModuleInternal.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+};
+
+export const validateConfig = (
+  filename: string,
+  moduleConfigInternal: ModuleConfigInternal
+) => {
+  if (!moduleConfigInternal.name) {
+    throw new Error(
+      `Module ${filename} is missing a "name" in the config function.`
```
mkilpatrick, line 19: This should never happen as moduleConfigInternal has a name based on the filename if the name in the config isn't set.

packages/pages/src/common/src/module/types.ts
```diff
@@ -1,10 +1,21 @@
...
  * The type definition for the module's default function.
  *
  * @public
  */
 export type Module = () => React.JSX.Element;
 
+/**
+ * The type to include in any module file.
+ */
+export interface ModuleModule {
```
mkilpatrick, line 13: Do we need this type at all? It's not something someone would use since it defines the whole file which you can't actually use in any practical sense. We could just remove it to avoid this weird name.

packages/pages/src/dev/server/middleware/serverRenderModule.ts
```diff
@@ -0,0 +1,42 @@
...
+  vite: ViteDevServer;
+  projectStructure: ProjectStructure;
+};
+
+export const serverRenderModule =
+  ({ vite, projectStructure }: Props): RequestHandler =>
+  async (req, res, next): Promise<void> => {
+    try {
+      const moduleName = req.originalUrl
+        .substring(req.originalUrl.lastIndexOf("modules/") + 8)
```
mkilpatrick, line 19: You can do lastIndexOf("/") + 1

packages/pages/src/dev/server/middleware/serverRenderModule.ts
```diff
@@ -0,0 +1,42 @@
...
+        .split("/")[0]
```
mkilpatrick, line 20: This isn't necessary

packages/pages/src/common/src/module/internal/getModuleFilepaths.ts
```diff
@@ -0,0 +1,89 @@
...
+ * @returns a list of module filepaths
+ */
+export const getModuleFilepaths = (paths: Path[]): string[] => {
+  const moduleFilePaths: string[] = [];
+  const addedFilenames: Set<string> = new Set();
+  paths.forEach((p) => {
+    readdirSync(p.getAbsolutePath(), { withFileTypes: true })
+      .filter((dir) => dir.isDirectory())
+      .map((dir) => dir.name)
+      .forEach((dir) => {
```
mkilpatrick, line 28: I don't think we need to do this validation here. It can happen at request time if a file doesn't exist when we try to load the module at a path that has no tsx file.

packages/pages/src/common/src/project/structure.ts
```diff
@@ -244,6 +244,27 @@ export class ProjectStructure {
     return [new Path(templatesRoot)];
   };
 
+  /**
+   * @returns the list of of src/modules, taking scope into account. If a scope is defined then
+   * both the scoped and non-scoped module paths are returned.
+   */
+  getModulePaths = () => {
```
mkilpatrick, line 251: Alexis [wrote](https://github.com/yext/pages/pull/494/files#diff-9065153815b3fabd28e0778d526a6430d8985add32f659f18d19293b9d316901R278) one too but yours is more correct. When you merge take yours instead and update her code that was using her version.

packages/pages/src/dev/server/middleware/serveModule.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const html = transformedIndexHtml.replace(
+    serverRenderTemplateModule.replacementTag,
+    await getServerHtml()
+  );
+
+  // Send the rendered HTML back.
+  res
+    .status(200)
+    .type(lookup(moduleInternal.path) || "text/html")
```
mkilpatrick, line 63: You wrote `getContentType` but don't use it. Isn't it required that files are always text/html though? We can't serve non-html modules.

packages/pages/src/dev/server/ssr/findMatchingModule.ts
```diff
@@ -0,0 +1,32 @@
+import { ViteDevServer } from "vite";
+import {
+  convertModuleToModuleInternal,
+  ModuleInternal,
+} from "../../../common/src/module/internal/types.js";
+import { ModuleModule } from "../../../common/src/module/types.js";
+import { loadViteModule } from "./loadViteModule.js";
+
+// Determines the module to load from a given feature name (from the exported config)
```
mkilpatrick, line 9: from a given module name

packages/pages/src/common/src/module/types.ts
jwartofsky-yext: We use this in the call to SSRLoadModule which populates it with everything from the .tsx file. It's only used in one place so we might be able to just use <any> and parse instead.

packages/pages/src/dev/server/middleware/serverRenderModule.ts
mkilpatrick: That's fine. We're not allowing multi-level deep urls like that. Maybe that will be a feature someday but not something to worry about now.

packages/pages/src/common/src/loader/vite.ts
```diff
@@ -23,6 +23,29 @@ export const getViteServerConfig = (
...
 
+export const getViteServerConfigWithPostCssPath = (
+  projectStructure: ProjectStructure,
+  postCssPath: string
+): InlineConfig => {
+  return {
+    css: {
+      postcss: postCssPath,
+    },
+    server: {
```
mkilpatrick, line 34: Don't duplicate all of this. Just do this where you need it:

```
{
  ...getViteServerConfig(ps),
  css: {...}
}
```

packages/pages/src/common/src/module/types.ts
mkilpatrick: Ok maybe rename it to something like `ModuleDefinition`

packages/pages/src/dev/dev.ts
```diff
@@ -104,5 +106,6 @@ export const devCommand = (program: Command) => {
     .option("--noInit", "Disables automatic yext init with .yextrc file")
     .option("--noGenFeatures", "Disable feature.json generation step")
     .option("--port <number>", "The port to use for the dev server")
+    .option("--widget <string>", "Name of the widget to load postcss for.")
```
mkilpatrick, line 109: We're not loading postcss for it. We're just loading the widget.

packages/pages/src/dev/server/server.ts
```diff
@@ -40,9 +45,18 @@ export const createServer = async (
 
   // initialize the default project structure and use to help configure the dev server
   const projectStructure = await ProjectStructure.init({ scope });
+  const postCssPath = await getPostCssPathFromModuleName(
```
mkilpatrick, line 48: All of the widget stuff should be contained. We shouldn't even load the index page. We should actually just pop open a tab at the widget path.

```
if (widget) {
 // do the widget things

 // we need some of the app.use stuff, but only for vite, none of our stuff
  app.use(vite.middlewares);
  app.use(errorMiddleware(vite));
  app.listen(devServerPort, () =>
    process.stdout.write(`listening on :${devServerPort}\n`)
  );
 return; // idk if return is required or app.listen() just stops further execution
}

// everything else we have
```

packages/pages/src/dev/server/ssr/findMatchingModule.ts
```diff
@@ -30,3 +35,86 @@ export const findModuleInternal = async (
...
+      const exportedDeclarations = sourceFile.getExportedDeclarations();
+      for (const [name, declarations] of exportedDeclarations) {
+        if (name == "config") {
+          declarations.forEach((declaration) => {
+            for (const child of declaration.getChildren()) {
+              const text = child.getText();
+              if (text.includes("name:")) {
+                const parsedName = text
+                  .substring(
+                    text.indexOf('"') + 1,
```
mkilpatrick, line 75: What if someone uses single quotes? I think we've done a less hacky way of this before. Can't we just get all of the config and convert it to json?

packages/pages/src/dev/server/ssr/findMatchingModule.ts
```diff
@@ -30,3 +35,86 @@ export const findModuleInternal = async (
...
+      }
+    }
+  }
+
+  // Use file path that matches given name if none match the config.name, if possible
+  if (targetModuleFilePath == "") {
+    for (const moduleFilePath of moduleFilePaths) {
+      const fileName = moduleFilePath.split("/").slice(-1)[0];
+      if (
+        fileName.substring(0, fileName.indexOf(".")).toLowerCase() ==
```
mkilpatrick, line 96: We don't care about filename, we care about folder name

packages/pages/src/dev/server/ssr/findMatchingModule.ts
```diff
@@ -30,3 +35,86 @@ export const findModuleInternal = async (
...
+
+  const postCssPath = path.resolve(
+    targetModuleFilePath,
+    "../postcss.config.js"
+  );
+
+  if (fs.existsSync(postCssPath)) {
+    return postCssPath;
+  }
+  console.log(`No postcss.config.js file found for module ${moduleName}`);
```
mkilpatrick, line 118: Don't log. TW/PostCSS is not a requirement.

packages/pages/src/dev/server/ssr/findMatchingModule.ts
```diff
@@ -30,3 +35,86 @@ export const findModuleInternal = async (
...
+  return "";
```
mkilpatrick, line 119: Just `return` here and make the type string | undefined

packages/pages/src/dev/server/ssr/findMatchingModule.ts
jwartofsky-yext: In Diana's instructions it says that the filename and folder name have to match. I can dig a bit more and see if that's actually enforced or just recommended.

I'll change this to check the folder name instead though.

packages/pages/src/dev/server/ssr/findMatchingModule.ts
jwartofsky-yext: It doesn't convert to json very well. The name we're looking for looks something like this and there's also other code in the file:

export const config: ModuleConfig = {
  name: "prefix"
}

I'll update to handle all quote types: " ` and '

packages/pages/src/common/src/module/internal/getModuleFilepaths.ts
jwartofsky-yext: As in don't check if there's only 1 source file? We already have the number of source files though. We're also looking for the singular source file in each module folder and wouldn't want to return multiple for one folder.

For templates we do some validation in this similar step:
https://github.com/yext/pages/blob/ba4647aa17ae53fa744f81530b2012a618c2fc58/packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts

packages/pages/src/common/src/module/internal/_server.tsx
jwartofsky-yext: They don't have PageProps like the template ones do, but otherwise they are the same.

Do the template clients/servers also have extra stuff we shouldn't have?

packages/pages/src/dev/server/ssr/findMatchingModule.ts
mkilpatrick: I thought there was still a better. Way. Talk to Alexis as she may know. If there's not then we can go with this.

packages/pages/src/common/src/module/internal/getModuleFilepaths.ts
mkilpatrick: Ok we can leave it.

packages/pages/src/common/src/module/internal/_server.tsx
mkilpatrick: We don't need the head tags stuff I don't think (unless it's necessary for the Vite css code to inject into). So if we don't need head then it'd just be like

```
<html>
  <div id="moduleName"><!--YEXT-SERVER--></div>
</html>
```

jwartofsky-yext: I get an error No head tag is defined if I remove the head tags.

Going to keep this as is since it works and matches what we do for templates.

packages/pages/src/common/src/project/structure.ts
jwartofsky-yext: Ack for now

Her version returns just one path while mine can return multiple. I will update hers in another PR.

packages/pages/src/dev/server/server.ts
```diff
@@ -40,6 +43,27 @@ export const createServer = async (
   // initialize the default project structure and use to help configure the dev server
   const projectStructure = await ProjectStructure.init({ scope });
 
+  if (widget) {
+    const postCssPath = await getPostCssPathFromModuleName(
```
mkilpatrick, line 47: We're searching for the module stuff twice, once here and again in serverRenderModule. Make a function that we call first that basically determines the module path and returns an object containing all module info we'd need. That's be the path to the module folder and the module name.

We can use the above to get the postCSS file as well as load the module in serverRenderModule. The function above can do the validation if more than 1 is found.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -25,6 +25,32 @@ type FileInfo = {
   name: string;
 };
 
+/**
+ * Returns a custom vite logger with nested tailwind warnings hidden.
+ * @return Logger
+ */
+export const createModuleLogger = (): Logger => {
+  const ignoredMessages: string[] = [
```
mkilpatrick, line 33: Move outside of func

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -75,10 +88,10 @@ export const buildModules = async (
   for (const [moduleName, fileInfo] of Object.entries(filepaths)) {
     logger.info = (msg, options) => {
       if (msg.includes("building for production")) {
-        loggerInfo(pc.green(`\nBuilding ${moduleName} module...`));
+        logger.info(pc.green(`\nBuilding ${moduleName} module...`));
```
mkilpatrick, line 91: Why change this from the loggerInfo helper?

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -198,7 +211,7 @@ const shouldBundleModules = (projectStructure: ProjectStructure) => {
  * @param modulePath
  * @returns name of module if set by user via ModuleConfig
  */
-const getModuleName = (modulePath: string): string | undefined => {
+export const getModuleName = (modulePath: string): string | undefined => {
```
mkilpatrick, line 214: If this is used by the plugin and dev then it should be moved to a common folder

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -25,6 +25,32 @@ type FileInfo = {
   name: string;
 };
 
+/**
+ * Returns a custom vite logger with nested tailwind warnings hidden.
+ * @return Logger
+ */
+export const createModuleLogger = (): Logger => {
```
mkilpatrick, line 32: Move this to a common place too

packages/pages/src/vite-plugin/modules/plugin.ts
jwartofsky-yext: there were two variables created loggerInfo and loggerWarn. LoggerWarn was not needed outside of the createModuleLogger() function so to clean up a little I also removed loggerInfo which was only used in these two lines.

mkilpatrick: Ah ok. I thought this was something from logError.ts.

packages/pages/src/vite-plugin/modules/plugin.ts
mkilpatrick: This wasn't moved

jwartofsky-yext: Fixed

mkilpatrick (approved):

PR #498, opened 2024-03-15
fix(dev): find correct document by slug

mkilpatrick:
There were some recent changes in the Yext CLI generate-test-data command. Beforehand we were blindly returning the first document returned from GTD, which led to inconsistencies because the wrong document could have been selected. Since the actual document we care about cannot be determined on the CLI side, we need additional logic to figure out which document to return based on the slug value and slug field used based on the templates available.

### DIFF
--- a/packages/pages/src/dev/server/ssr/generateTestData.ts
+++ b/packages/pages/src/dev/server/ssr/generateTestData.ts
@@ -89,8 +89,12 @@ export const generateTestDataForSlug = async (
 
   const parsedData = await spawnTestDataCommand(stdout, "yext", args);
 
-  // handle documents coming back as an array just in case
-  return parsedData?.length > 0 ? parsedData[0] : parsedData;
+  return getDocumentBySlug(
+    parsedData,
+    slug,
+    Array.from(slugFields),
+    templateModuleCollection
+  );
 };
 
 const loadTemplateModuleCollectionUsingVite = async (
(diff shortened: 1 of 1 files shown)


packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -294,3 +298,62 @@ const getDocumentByLocale = (parsedData: any, locale: string): any => {
...
+      `Multiple documents have the same slugField (${matchingSlugFields.join(
+        ", "
+      )}) with value: "${slug}"`
+    );
+  }
+
+  // Find the template that uses the slugfield
+  const matchingTemplateModules: TemplateModuleInternal<any, any>[] = [];
+  templateModuleCollection.forEach((templateModule) => {
+    const slugField = templateModule.config?.slugField || "slug";
```
asanehisa, line 337: random but do we prefer ?? over ||

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -294,3 +298,62 @@ const getDocumentByLocale = (parsedData: any, locale: string): any => {
...
+    if (slugField === matchingSlugFields[0]) {
+      matchingTemplateModules.push(templateModule);
+    }
+  });
+
+  if (matchingTemplateModules.length === 0) {
+    throw new Error(`Could not find template for slug: "${slug}"`);
+  } else if (matchingTemplateModules.length > 1) {
+    throw new Error(`Multiple templates found for slug: "${slug}"`);
```
asanehisa, line 346: do we want to list the multiple templates here (if easy)

packages/pages/src/dev/server/ssr/generateTestData.ts
mkilpatrick: `??` does nullish coalescing, so empty string will be taken as truthy and returned (if it happens to be empty string). We don't want that.

https://stackoverflow.com/a/61481631

asanehisa (approved):

jwartofsky-yext (approved):
lgtm

briantstephan (approved):

PR #502, opened 2024-03-19
feat: support scope with modules

asanehisa:
`npm run prod -- --scope myScope` works w/ myScope in templates, modules, and both. npm run build and dev work as expected as well. Double checked scaffolding supported scope still too. 

scope means files in both myScope and src/... are built and used. Without the scope being set, the files in myScope are not used. If a file is in myScope and in src/... (with same config name), the one in myScope should be used (if scope is set).

### DIFF
--- a/packages/pages/src/common/src/project/structure.ts
+++ b/packages/pages/src/common/src/project/structure.ts
@@ -1,8 +1,10 @@
 import pathLib from "node:path";
 import merge from "lodash/merge.js";
+import fs from "node:fs";
 import { Path } from "./path.js";
 import { determineAssetsFilepath } from "../assets/getAssetsFilepath.js";
 import { determinePublicFilepath } from "../assets/getPublicFilepath.js";
+import { Manifest } from "../template/types.js";
 
 /**
  * All important folders defined at the root of the project.
@@ -127,6 +129,8 @@ export interface ProjectStructureConfig {
    *
    * The subfolder path inside src/templates and sites-config
    * to scope a build to a subset of templates using a specific sites-config folder.
+   *
+   * Modules scoping is also supported.
    */
   scope?: string;
 }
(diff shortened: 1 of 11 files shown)


packages/yext-function/manifest.ts
```diff
@@ -128,6 +128,8 @@ export interface ProjectStructureConfig {
    *
    * The subfolder path inside src/templates and sites-config
    * to scope a build to a subset of templates using a specific sites-config folder.
+   *
+   * Modules scoping is also supported.
```
jwartofsky-yext, line 132: Describe testing process in the PR description

packages/pages/src/common/src/project/structure.ts
```diff
@@ -224,47 +228,37 @@ export class ProjectStructure {
...
     );
 
     if (this.config.scope) {
-      return [
-        new Path(pathLib.join(templatesRoot, this.config.scope)),
-        new Path(templatesRoot),
-      ];
+      // src/templates/[scope]
+      const scopedPath: string = pathLib.join(templatesRoot, this.config.scope);
+      if (fs && fs.existsSync(scopedPath)) {
```
mkilpatrick, line 245: Use optional chaining, aka the ? instead of &&

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -77,7 +83,9 @@ export const buildModules = async (
     addResponseHeadersToConfigYaml(
       projectStructure,
       {
-        pathPattern: `^modules/${moduleName}.*`,
+        pathPattern: scope
```
mkilpatrick, line 86: This shouldn't be dependent on scope, but where the module lives. If you define a scope but the module lives in root, this will output it to scope won't it?

asanehisa: Ah good catch. I removed that portion entirely since in [SUMO-5833](https://yexttest.atlassian.net/browse/SUMO-5833) we don't want it there at all anyways.

mkilpatrick (approved):

PR #504, opened 2024-03-20
fix(dev): adjust GTD logic and pass scope

mkilpatrick:
Fixes the following issues:
1. Not passing the --scope flag to the CLI GTD which is necessary for config v2 (we were only passing --hostname which works the same but only for config v1)
2. `slug` wasn't added to the --slugFields flag if no other alternate slugFields were used
3. Filters out static templates so we can actually find the correct document. Currently, the --featuresConfig flag is not honored by the CLI so static template document data is coming back when it shouldn't. This is a workaround for that.

### DIFF
--- a/packages/pages/src/dev/server/ssr/generateTestData.ts
+++ b/packages/pages/src/dev/server/ssr/generateTestData.ts
@@ -71,6 +71,10 @@ export const generateTestDataForSlug = async (
 
   const slugFields = new Set<string>();
   templateModuleCollection.forEach((templateModule) => {
+    // We don't want to add the default "slug" field when no entity templates need it
+    if (templateModule.config.templateType == "static") {
+      return;
+    }
     const slugField = templateModule?.config?.slugField;
     if (slugField) {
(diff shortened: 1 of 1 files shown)


packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -70,22 +70,16 @@ export const generateTestDataForSlug = async (
...
 
   const slugFields = new Set<string>();
-  let shouldAddDefaultSlugField = false;
   templateModuleCollection.forEach((templateModule) => {
     const slugField = templateModule?.config?.slugField;
     if (slugField) {
       slugFields.add(slugField);
     } else {
-      shouldAddDefaultSlugField = true;
+      slugFields.add("slug");
```
briantstephan, line 82: We only want to add this default slug field if we encounter a dynamic template without slugField set. This will add it if we find a static template without slugField (which is every static template).

briantstephan (approved):

asanehisa (approved):

PR #509, opened 2024-03-25
perf(dev): optimize loading index page

mkilpatrick:
The template modules were all being loaded for each localData file. For repos with many templates/documents this was extremely slow. We now only load the template modules once.


packages/pages/src/common/src/template/loader/loader.ts
```diff
@@ -55,3 +58,28 @@ export type TemplateModuleCollection = Map<
   string,
   TemplateModuleInternal<any, any>
 >;
+
+/**
+ * Simlar to loadTemplateModules above but reuses and existing Vite dev server.
```
asanehisa, line 63: ? reuses an?

asanehisa (approved):

PR #510, opened 2024-03-27
feat(dev): upgrade script uses engine 18.4.0

jwartofsky-yext:
Verified that upgrade script works correctly and sets the engine to 18.4.0

### DIFF
--- a/packages/pages/src/upgrade/pagesUpdater.ts
+++ b/packages/pages/src/upgrade/pagesUpdater.ts
@@ -440,7 +440,7 @@ export const installDependencies = async () => {
 };
 
 // Note that Node 20 <20.2.0 leads to build errors: `Unexpected early exit.`
-const NODE_ENGINES = "^18.40.0 || >=20.2.0";
+const NODE_ENGINES = "^18.4.0 || >=20.2.0";
 /**
  * Update package engines to latest supported node versions.
  */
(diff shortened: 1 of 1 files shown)


asanehisa (dismissed):

packages/pages/src/upgrade/pagesUpdater.ts
```diff
@@ -440,7 +440,7 @@ export const installDependencies = async () => {
 };
 
 // Note that Node 20 <20.2.0 leads to build errors: `Unexpected early exit.`
-const NODE_ENGINES = "^18.0.0 || >=20.2.0";
+const NODE_ENGINES = "^18.20.0 || >=20.2.0";
```
mkilpatrick, line 443: As I mentioned in Slack, I don't think this is the correct fix. This wasn't a problem until the dep upgrade PR and searching the error has results that it's related to types/node.

jwartofsky-yext: I tried reverting the @types/node version as used by pages back to ^20.10.4 for the branch/commit here:
https://www.yext.com/s/950642/yextsites/138238/branch/102492/deploys/recent

Reverting @types/node version did not resolve the issue, but changing the node version to 18.20 does resolve the issue. I also confirmed that 18.19 also works.

mkilpatrick: Wrong version. You put 18.40.0

mkilpatrick: Also update the title

asanehisa (approved):

mkilpatrick (approved):

PR #512, opened 2024-04-01
perf(dev): speed up page loads and alleviate memory pressure

mkilpatrick:
This change speeds up the dev server page load performance on subsequent page refreshes/HMR changes by caching a mapping of templateName to templateFilepath. This allows us to find the templates much more quickly. Beforehand, we were iterating through and loading every template on each requests.

This caching also has the added benefit of making the Node memory less spikey (and with less GC) because less modules are loaded now.

### DIFF
--- a/packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
+++ b/packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
@@ -1,25 +1,30 @@
 import { ViteDevServer } from "vite";
 import {
   getAllLocalData,
-  getLocalData,
+  readLocalDataFile,
   staticPageCriterion,
 } from "./getLocalData.js";
-import {
-  TemplateModuleInternal,
-  convertTemplateModuleToTemplateModuleInternal,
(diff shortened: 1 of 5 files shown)


packages/pages/src/dev/server/ssr/findMatchingStaticTemplate.ts
```diff
@@ -1,26 +1,31 @@
...
+};
+
+// A cache of slug to template/localData filepaths to avoid loading everything on every lookup.
+const slugToModuleAndLocalDataPaths = new Map<
+  string,
+  TemplateAndLocalDataPaths
+>();
+
 /**
  * Loops through all static templates. For each, finds all localData static files
```
asanehisa, line 30: is this comment still accurate

mkilpatrick: Yes. Really all that changed here is using a cache to do lookups quicker.

jwartofsky-yext (dismissed):
lgtm

jwartofsky-yext (approved):

PR #514, opened 2024-04-02
fix: handle static assets in modules

asanehisa:
Static assets worked fine before this if they were under 4 kb. 

Now, if the user provides their hosted domain (in the ModuleConfig), static assets will be imported from the hosted domain's path.

Tested with 10 mb svg and works with new method. Other static assets are also handled. 

Example of what can occur in the umd.js file. If domain is set to "http://localhost:8000", then "/assets/test.svg" is replaced with "http://localhost:8000/modules/assets/test.svg"


packages/pages/etc/pages.api.md
```diff
@@ -112,6 +112,7 @@ export interface ModuleClientServerRenderTemplates {
 
 // @public
 export interface ModuleConfig {
+  domain?: string;
```
mkilpatrick, line 115: Can you move this after name?

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -117,9 +123,44 @@ export const buildModules = async (
...
+/**
+ * Replaces asset paths with absolute paths based on hosted domain.
+ * Ex. if the asset path is "src/assets/foo.svg", it'll be changed to
+ * "<hostedDomain>/modules/assets/foo.svg".
+ *
+ * @param umdPath dist/modules/<moduleName>.umd.js
+ * @param domain if given by user
+ * @param subfolders of the ProjectStructure
+ */
+const editAssetsPaths = (
```
mkilpatrick, line 143: Let's try Vite's [renderBuiltUrl](https://vitejs.dev/guide/build.html#advanced-base-options). I believe this is exactly what we want instead of building and parsing/rewriting the final output.

packages/pages/etc/pages.api.md
asanehisa: odd, it is in the types.ts file but not in that documentation.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -97,6 +94,15 @@ export const buildModules = async (
           "css-syntax-error": "silent",
         },
       },
+      experimental: {
+        renderBuiltUrl(filename, { type }) {
+          const domain = getModuleDomain(fileInfo.path);
+          if (type === "asset" && domain) {
+            return `${domain}/${subfolders.modules}/` + filename;
```
mkilpatrick, line 101: Add filename to the string interpolation instead of using +

jwartofsky-yext (approved):
lgtm

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -96,9 +93,15 @@ export const buildModules = async (
       },
       experimental: {
         renderBuiltUrl(filename, { type }) {
-          const domain = getModuleDomain(fileInfo.path);
+          let domain = `http://localhost:8000`;
+          if (typeof process.env.YEXT_SITE_ARGUMENT !== "undefined") {
+            domain = new URL(
+              "http://" +
+                JSON.parse(process.env.YEXT_SITE_ARGUMENT).productionDomain
```
mkilpatrick, line 100: I think we need to be careful here in the case the parse fails and handle that. Also, I think this should be https. It might work with just // which means "use whatever protocol the user is on" but you'd have to test that out.

asanehisa: Added try / catch. Tested w/ // but didn't work so am using https

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -95,10 +95,14 @@ export const buildModules = async (
...
-              "http://" +
-                JSON.parse(process.env.YEXT_SITE_ARGUMENT).productionDomain
-            ).toString();
+            try {
+              domain = new URL(
+                "https://" +
+                  JSON.parse(process.env.YEXT_SITE_ARGUMENT).productionDomain
+              ).toString();
+            } catch (_) {
+              logger.error("Cannot parse YEXT_SITE_ARGUMENT");
```
mkilpatrick, line 104: This should probably fail the build.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -101,7 +101,7 @@ export const buildModules = async (
                   JSON.parse(process.env.YEXT_SITE_ARGUMENT).productionDomain
               ).toString();
             } catch (_) {
-              logger.error("Cannot parse YEXT_SITE_ARGUMENT");
+              throw new Error("Cannot parse YEXT_SITE_ARGUMENT");
```
mkilpatrick, line 104: Use the logErrorAndExit function.

mkilpatrick (approved):

PR #515, opened 2024-04-02
feat(plugin): add support for statusCode to getRedirects

jwartofsky-yext:
Allows the user to pass in either RedirectSource {source: string, statusCode: int}[] or string[] for GetRedirects

The Yext CLI has not been updated to accept the new type yet, so this currently only works for string[]

### DIFF
--- a/packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
+++ b/packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
@@ -11,6 +11,7 @@ import {
   convertTemplateModuleToTemplateModuleInternal,
   TemplateModuleInternal,
 } from "../../../../common/src/template/internal/types.js";
+import { RedirectSource } from "../../../../common/src/redirect/types.js";
 import { ProjectStructure } from "../../../../common/src/project/structure.js";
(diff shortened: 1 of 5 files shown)


packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.ts
```diff
@@ -97,7 +98,7 @@ const importRenderTemplate = async (
 export type GeneratedPage = {
   path: string;
   content?: string;
-  redirects: string[];
+  redirects: (RedirectSource | string)[];
```
mkilpatrick, line 101: You need to update the redirects field in the return of generateResponses and map them to the new type. For any _just string_ items in our array we need to convert to the new object with a default status of 301.

jwartofsky-yext: Based on the comment here: https://yext.slack.com/archives/C06PKRZN4AW/p1712170611016879?thread_ts=1712167837.746649&cid=C06PKRZN4AW

We no longer want to do this and want to keep passing through the redirects as is?

mkilpatrick (approved):

PR #519, opened 2024-04-12
refactor: remove manifest check from getTemplatePaths

mkilpatrick:
Manifest is no longer being used after the changes in https://github.com/yext/pages/pull/517.


jwartofsky-yext (approved):
lgtm

PR #520, opened 2024-04-22
fix(dev): hide nested css warning

asanehisa:


packages/pages/src/dev/server/server.ts
```diff
@@ -96,7 +99,10 @@ export const createServer = async (
   }
 
   // create vite using ssr mode
-  const vite = await createViteServer(getViteServerConfig(projectStructure));
+  const vite = await createViteServer({
+    ...getViteServerConfig(projectStructure),
+    customLogger: createModuleLogger(),
```
mkilpatrick, line 104: If we're doing this in every instance we createViteServer we should move customLogger to getViteServerConfig instead.

mkilpatrick (approved):

PR #524, opened 2024-05-13
fix(plugin): use the user's vite.config

asanehisa:
Tested and confirmed it does use the user's build.rollupOptions.external when I add that to the `vite.config.js`
Tested w/ both the problem build (a serverless func repo) and a modules repo.

`configFile: false` should remain or else it'll get stuck in infinite loop of yext plugin.

### DIFF
--- a/packages/pages/src/util/viteConfig.ts
+++ b/packages/pages/src/util/viteConfig.ts
@@ -18,9 +18,10 @@ export const scopedViteConfigPath = (scope?: string) => {
 };
 
 export const removePluginFromViteConfig = (config: any) => {
-  if (config?.plugins) {
+  if (config?.plugins?.[0]) {
     config.plugins = config.plugins?.[0]?.filter(
-      (obj: any) => obj.name !== "vite:react-refresh"
+      (obj: any) =>
+        obj.name !== "vite:react-refresh" || obj.name !== "vite:react-babel"
     );
   }
   return config;
--- a/packages/pages/src/util/viteConfig.ts
+++ b/packages/pages/src/util/viteConfig.ts
@@ -18,11 +18,8 @@ export const scopedViteConfigPath = (scope?: string) => {
 };
 
 export const removePluginFromViteConfig = (config: any) => {
-  if (config?.plugins?.[0]) {
-    config.plugins = config.plugins?.[0]?.filter(
-      (obj: any) =>
-        obj.name !== "vite:react-refresh" || obj.name !== "vite:react-babel"
-    );
+  if (config?.plugins) {
+    config.plugins = "";
   }
   return config;
 };
--- a/packages/pages/src/util/viteConfig.ts
+++ b/packages/pages/src/util/viteConfig.ts
@@ -16,3 +16,12 @@ export const scopedViteConfigPath = (scope?: string) => {
     return viteConfigPath;
   }
 };
+
+export const removePluginFromViteConfig = (config: any) => {
+  if (config?.plugins) {
+    config.plugins = config.plugins?.[0]?.filter(
+      (obj: any) => obj.name !== "vite:react-refresh"
+    );
+  }
+  return config;
+};
--- a/packages/pages/src/vite-plugin/modules/plugin.ts
+++ b/packages/pages/src/vite-plugin/modules/plugin.ts
@@ -65,7 +65,10 @@ export const buildModules = async (
   const viteConfig = await import(
     scopedViteConfigPath(projectStructure.config.scope) ?? ""
   );
-  const rollupOptions = viteConfig?.default?.build?.rollupOptions;
+  const config = viteConfig?.default;
+  const filteredPlugins = config?.plugins?.[0]?.filter(
+    (obj: any) => obj.name !== "vite:react-refresh"
+  );
 
   for (const [moduleName, fileInfo] of Object.entries(filepaths)) {
     logger.info = (msg, options) => {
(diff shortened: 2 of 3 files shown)


packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -61,6 +62,14 @@ export const buildModules = async (
     );
   }
 
+  const viteConfig = await import(
+    scopedViteConfigPath(projectStructure.config.scope) ?? ""
+  );
+  const config = viteConfig?.default;
+  const filteredPlugins = config?.plugins?.[0]?.filter(
```
mkilpatrick, line 69: Don't you only want to remove the pages plugin? I think we can use mergeConfig instead and you set the below build config to an object and pass it in as the overrides - https://vitejs.dev/guide/api-javascript#mergeconfig

asanehisa: yay used that. also prevented the looping issue

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -70,7 +75,7 @@ export const buildModules = async (
       loggerInfo(msg, options);
     };
 
-    await build({
+    const override = {
```
mkilpatrick, line 77: Call this moduleBuildConfig

packages/pages/src/vite-plugin/serverless-functions/plugin.ts
```diff
@@ -53,7 +58,7 @@ export const buildServerlessFunctions = async (
       loggerInfo(msg, options);
     };
 
-    await build({
+    const override = {
```
mkilpatrick, line 60: Call this serverlessFunctionBuildConfig

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -135,7 +140,8 @@ export const buildModules = async (
           },
         }),
       ],
-    });
+    };
+    await mergeConfig(viteConfig, override);
```
mkilpatrick, line 143: Don't you need to pass isRoot as false? It says
> isRoot represents the level within the Vite config which is being merged. For example, set false if you're merging two build options.

mkilpatrick, line 143: You're not executing build anymore. You're only executing the merging of the configs. You need to merge them and then pass that config to build.

asanehisa: Yes ur right. Didn't notice since npm run build passed (but the serverless functions weren't built). fixed!

packages/pages/src/util/viteConfig.ts
```diff
@@ -16,3 +16,12 @@ export const scopedViteConfigPath = (scope?: string) => {
     return viteConfigPath;
   }
 };
+
+export const removePluginFromViteConfig = (config: any) => {
+  if (config?.plugins) {
+    config.plugins = config.plugins?.[0]?.filter(
+      (obj: any) => obj.name !== "vite:react-refresh"
```
asanehisa, line 23: If I console.log it, vite.config.js has two plugins 
```
  "plugins": [
    [
      {
        "name": "vite:react-babel",
        "enforce": "pre"
      },
      {
        "name": "vite:react-refresh",
        "enforce": "pre"
      }
    ],
    {}
  ],
```

my vite.config.js looks like:
```
export default defineConfig({
  resolve: {
    alias: {
      src: path.resolve("src/"),
    },
  },
  plugins: [react(), yextSSG()],
  build: {
    rollupOptions: {
      external: ['ssh2']
    }
  }
});
```

upon removing the `vite:react-refresh` plugin, the infinite looping doesn't occur. will look into exactly why tmrw.

asanehisa: talked offline.

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -135,7 +142,14 @@ export const buildModules = async (
           },
         }),
       ],
-    });
+    };
+    await build(
+      mergeConfig(
+        removePluginFromViteConfig(viteConfig.default),
+        moduleBuildConfig,
+        false
```
mkilpatrick, line 150: remove this, it can be true since they are root

packages/pages/src/util/viteConfig.ts
```diff
@@ -16,3 +16,13 @@ export const scopedViteConfigPath = (scope?: string) => {
     return viteConfigPath;
   }
 };
+
+export const removePluginFromViteConfig = (config: any) => {
```
mkilpatrick, line 20: I think we can just remove all plugins

packages/pages/src/vite-plugin/modules/plugin.ts
```diff
@@ -92,7 +99,7 @@ export const buildModules = async (
         },
       },
       experimental: {
-        renderBuiltUrl(filename, { type }) {
+        renderBuiltUrl(filename: any, { type }: any) {
```
mkilpatrick, line 102: Can you give this the real types? `any` is a catchall and you generally don't want to use it if possible

asanehisa: double checked types w/ vite's documentation

mkilpatrick (approved):

PR #526, opened 2024-06-12
fix(dev): don't lookup mime type unless there's an extension

mkilpatrick:
The mime-type library returns "application/json" when passed a value of "map" (also true for "json"). Don't use the library if there is no extension and instead default to "text/html".

This bug presented itself when someone used a slug value of "map" and the page rendered as raw html.


jwartofsky-yext (approved):
lgtm

PR #529, opened 2024-07-23
feat(scaffold): add prompts for scaffold template

asanehisa:


packages/pages/src/scaffold/template/generate.ts
```diff
@@ -0,0 +1,117 @@
...
+      inactive: "Static",
+    },
+    {
+      type: (prev) => (prev ? "select" : null),
+      name: "entityScope",
+      message:
+        "How would you like you to define the entity scope for your template?",
+      choices: [
+        { title: "Entity Type", value: "entityTypes" },
+        { title: "Static", value: "static" },
```
mkilpatrick, line 42: Static isn't a scope option. Aaron wrote "Static Filter" but he fixed after I commented. It should be "Saved Filter".

packages/pages/src/scaffold/template/generate.ts
```diff
@@ -0,0 +1,117 @@
...
+  ];
+
+  const response = await prompts(questions);
+
+  // TODO (SUMO-5251): handle generating VE templates
+  // TODO (SUMO-5252): handle generating non-VE templates
+};
+
+// Returns true if templateName can be formatted into valid filename and that filename isn't being used.
+const validateTemplateName = (
```
mkilpatrick, line 80: Don't we already have code that does this we can reuse?

packages/pages/src/scaffold/template/generate.ts
```diff
@@ -0,0 +1,117 @@
...
+    formattedFileName
+  );
+  if (fs.existsSync(templatePath)) {
+    return false;
+  }
+
+  return true;
+};
+
+const formatFileName = (templateName: string): string => {
```
mkilpatrick, line 102: Do we have this already too? If not, add docs for what it does.

asanehisa: The code for it is in alpha unfortunately.

packages/pages/src/scaffold/template/generate.ts
asanehisa: Not quite, modules has a decent amount of diff naming rules than templates

mkilpatrick (approved):

PR #530, opened 2024-07-24
feat: add puck to ProjectStructureConfig

asanehisa:


packages/pages/src/common/src/project/structure.ts
```diff
@@ -31,6 +31,8 @@ export interface Subfolders {
   redirects: string;
   /** The Node functions folder */
   serverlessFunctions: string; // Node functions
+  /** The puck folder */
+  puck: string;
```
mkilpatrick, line 35: Mentioned this in your other PR, but we shouldn't need to do this. We don't want puck-specific things in PagesJS other than creating the new ve templates via the new cmd.

PR #531, opened 2024-07-24
feat(dev): add pageSets to global variable in header in dev mode

jwartofsky-yext:
Adds the templatesJson to the header so that we can access it in the Visual Editor.

Tested locally and verified that the templatesJson was present in a script tag in the header.

Also verified that `npm run prod` still works as expected and the templatesJson is not present.

### DIFF
--- a/packages/pages/src/common/src/template/hydration.ts
+++ b/packages/pages/src/common/src/template/hydration.ts
@@ -1,6 +1,7 @@
 import { HeadConfig, renderHeadConfigToString } from "./head.js";
 import { convertToPosixPath } from "./paths.js";
 import { TemplateRenderProps } from "./types.js";
+import { FeaturesConfig } from "../feature/features.js";
 
 /**
  * Imports the custom hydration template and entrypoint template as modules and calls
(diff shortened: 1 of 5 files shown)


packages/pages/src/common/src/template/hydration.ts
```diff
@@ -134,21 +135,29 @@ const getCommonInjectedIndexHtml = (
...
-  return getCommonInjectedIndexHtml(
+  let commonIndex = getCommonInjectedIndexHtml(
     clientHydrationString,
     indexHtml,
     appLanguage,
     headConfig
   );
+  commonIndex = injectIntoEndOfHead(
+    commonIndex,
+    `<script id="templatesJsonDev">${JSON.stringify(templatesConfig, null)}</script>`
```
mkilpatrick, line 157: Don't use an id and set the json object to a `const pageSets`

jwartofsky-yext: Done

mkilpatrick (approved):

PR #532, opened 2024-07-24
feat(scaffold): create ve template on scaffold

asanehisa:


packages/pages/src/common/src/parsers/puckConfigParser.ts
```diff
@@ -0,0 +1,52 @@
...
+import { SyntaxKind } from "ts-morph";
+
+/**
+ * ConfigParser is a class for parsing the puck config file.
+ */
+export default class ConfigParser {
+  /**
+   * Adds variables to the puckConfig file and adds the new config to
+   * the exported map.
+   * @param templateName as recieved by user input
```
mkilpatrick, line 13: I don't think we need all 3 of these. We should just take the user's input for templateName, cleanse it, and pass/use that one thing for everything: the filename, and template name in the TemplateConfig. Technically the name isn't even necessary in the TemplateConfig because the filename will just be used, but we can add it for completeness.

packages/pages/src/common/src/parsers/puckConfigParser.ts
```diff
@@ -0,0 +1,52 @@
+import fs from "node:fs";
+import SourceFileParser, { createTsMorphProject } from "./sourceFileParser.js";
+import { newConfig } from "../../../scaffold/template/sampleTemplates.js";
+import { SyntaxKind } from "ts-morph";
+
+/**
+ * ConfigParser is a class for parsing the puck config file.
+ */
+export default class ConfigParser {
```
mkilpatrick, line 9: Any reason this a class and not just a function?

packages/pages/src/common/src/parsers/puckConfigParser.ts
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Adds variables to the puckConfig file and adds the new config to
+   * the exported map.
+   * @param templateName as recieved by user input
+   * @param formattedTemplateName capitalized first letter of fileName
+   * @param fileName removed invalid chars from templateName
+   * @param filepath /src/puck/ve.config.ts
+   */
+  addDataToPuckConfig(
```
mkilpatrick, line 18: Can you add tests for this?

packages/pages/src/common/src/project/structure.ts
```diff
@@ -31,6 +31,8 @@ export interface Subfolders {
   redirects: string;
   /** The Node functions folder */
   serverlessFunctions: string; // Node functions
+  /** The puck folder */
+  puck: string;
```
mkilpatrick, line 35: We shouldn't have anything specific to Puck in the pagesjs code (other than creating these VE templates/puckConfig). "puck" as a folder in project structure shouldn't be necessary. Also, everything in there now will be part of the new visual-editor lib and a user would just have an edit.tsx template that uses the <Edit> component we export from that lib. I know it doesn't exist yet, but that's what this PR should make use of.

packages/pages/src/scaffold/template/sampleTemplates.ts
```diff
@@ -0,0 +1,128 @@
...
+    isDraft: true,
+  }
+};
+
+export const transformProps = async (data) => {
+  const { document } = data;
+  const entityConfigurations = document.c_visualConfigurations ?? [];
+  const entityLayoutConfigurations = document.c_pages_layouts ?? [];
+  const siteLayoutConfigurations = document._site?.c_visualLayouts;
+  try {
```
mkilpatrick, line 56: All of this will end up being a one line function call to a function from the ve lib

packages/pages/src/scaffold/template/sampleTemplates.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export default ${formattedTemplateName};
+`;
+};
+
+export const newConfigFile = (templateName: string, fileName: string) => {
+  const formattedTemplateName =
+    fileName.charAt(0).toUpperCase() + fileName.slice(1);
+
+  return `import type { Config } from "@measured/puck";
```
mkilpatrick, line 108: We will need to add puck and the new ve lib to the package.json if it's not already there (and probably npm install?)

packages/pages/src/scaffold/template/sampleTemplates.ts
```diff
@@ -0,0 +1,128 @@
+// TODO: Use new resolveData (or similar function) in transformProps
+export const visualEditorTemplateCode = (
```
mkilpatrick, line 2: Can you add tests for this?

packages/pages/src/common/src/parsers/puckConfigParser.ts
asanehisa: Oh it's cuz the other parsers were classes w/ multiple functions. Made it a func.

packages/pages/src/scaffold/template/sampleTemplates.ts
asanehisa: Oops good catch, I forgot

packages/pages/src/scaffold/template/sampleTemplates.ts
asanehisa: using it from this [PR](https://github.com/yext/visual-editor/pull/6/), will update if changed

packages/pages/src/scaffold/template/sampleTemplates.ts
asanehisa: Ah rn, visual-editor can't be npm installed b/c it is an internal repo, will we change this?

packages/pages/src/scaffold/template/generate.ts
```diff
@@ -152,13 +157,14 @@ const addVETemplateToConfig = (
...
-      configPath
-    );
+    addDataToPuckConfig(fileName, configPath);
   } else {
-    fs.writeFileSync(configPath, newConfigFile(templateName, fileName));
+    fs.writeFileSync(configPath, newConfigFile(fileName));
   }
 };
+
+const getDependencies = async () => {
```
mkilpatrick, line 166: addVEDependencies?

packages/pages/src/scaffold/template/sampleTemplates.ts
mkilpatrick: Yes, so you can assume they will exist for this PR.

mkilpatrick (approved):

PR #534, opened 2024-07-29
feat(dev): override document visual configuration

briantstephan:
This provides a post request option for `serverRenderRoute` which overrides the c_visualLayouts field of the pulled document with the overrides that are passed in from the request.

### DIFF
--- a/packages/pages/src/dev/server/middleware/serverRenderRoute.ts
+++ b/packages/pages/src/dev/server/middleware/serverRenderRoute.ts
@@ -88,9 +88,12 @@ export const serverRenderRoute =
         return;
       }
 
+      const overrides = JSON.parse(req.body.overrides);
+      const documentWithOverrides = overrideDocument(document, overrides.data);
+
       const props = await propsLoader({
         templateModuleInternal,
-        document,
+        document: documentWithOverrides,
       });
       await sendAppHTML(
         res,
@@ -133,4 +136,24 @@ const getDocument = async (
       entityPageCriterion(entityId, templateModuleInternal.config.name, locale)
     )
   )?.document;
+};
+
+const overrideDocument = (document: any, overrideData: any) => {
+  if (!overrideData) {
+    return document;
+  }
+  return {
+    ...document,
+    _site: {
+      ...document._site,
+      c_visualLayouts: [
+        {
+          c_visualConfiguration: {
+            data: JSON.stringify(overrideData),
+            template: document.__.name,
+          },
+        },
+      ],
+    },
+  };
 };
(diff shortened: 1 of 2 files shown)


packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -88,9 +88,12 @@ export const serverRenderRoute =
         return;
       }
 
+      const overrides = JSON.parse(req.body.overrides);
```
mkilpatrick, line 91: Is this safe if there are no overrides or body?

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -134,3 +137,23 @@ const getDocument = async (
...
+
+const overrideDocument = (document: any, overrideData: any) => {
+  if (!overrideData) {
+    return document;
+  }
+  return {
+    ...document,
+    _site: {
+      ...document._site,
+      c_visualLayouts: [
```
mkilpatrick, line 149: Mentioned in the other CR but the overrides should come through as a map of key to overridden value so that it's generic. Right now it's very specific to VE and not reusable.

packages/pages/src/dev/server/server.ts
```diff
@@ -191,9 +191,18 @@ export const createServer = async (
     }
   });
 
+  app.post(
```
mkilpatrick, line 194: If you're not passing the override as a param, does app.use encompass app.post too? Basically, do we need this?

playground/multibrand-site/sites-config/sunglasses.oakley.com/features.json
```diff
@@ -15,21 +15,54 @@
       "name": "robots",
       "templateType": "JS",
       "staticPage": {}
+    },
+    {
```
mkilpatrick, line 19: Weird that this got added

packages/pages/src/dev/server/server.ts
briantstephan: We need it separate so that it's forced to use `serverRenderRoute` instead of `serverRenderSlugRoute`.

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
briantstephan: Updated, should be safe now

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
briantstephan: Done

mkilpatrick (approved):

PR #536, opened 2024-08-07
fix: removes warning about --experimental flags

asanehisa:
Tested locally with node 18 and 20

Tested in platform with a starter site as well. 

[PR](https://github.com/yext/pages/commit/366d2931fccbf6b2cf4b011e41eb2963a9913fe2) that added this code initially


asanehisa:
@mkilpatrick lmk if there's more specific testing I should be doing

PR #542, opened 2024-08-27
feat(dev): make site stream available to Visual Editor via global variable

benlife5:
- Reads the site stream config from sites-config or config.yaml
- injects it into the page head as global variable `siteStream`
- together with [https://github.com/yext/visual-editor/pull/31](https://github.com/yext/visual-editor/pull/31)
- future update to Storm will take this information from Visual Editor to return `_site` object on documents in dev mode


asanehisa (dismissed):
lgtm but wait for duval's +2

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -208,26 +209,27 @@ async function spawnTestDataCommand(
...
   const siteStreamPath = path.resolve(
     projectStructure.getSitesConfigPath().path,
     projectStructure.config.sitesConfigFiles.siteStream
   );
 
   if (fs.existsSync(siteStreamPath)) {
-    return prepareJsonForCmd(
-      JSON.parse(fs.readFileSync(siteStreamPath).toString())
-    );
+    return JSON.parse(fs.readFileSync(siteStreamPath).toString());
```
mkilpatrick, line 221: I think this wouldn't return data in the same format. We have a migration command in migrateConfig.ts call `formatSiteStream` which removes the $ from id. So if someone was on the old sites-config version then the format wouldn't work. I think we need to run it through this.

I'm thinking we move this function to src/common/src/feature/stream.ts and rename it to readSiteStream. We should define another type in there for this version of a SiteStream, which is different from StreamConfig. Add a comment about it. Additionally, if using sites-config, we should use that formatSiteStream on it. Move that function to this file as well. Then update all callers. Basically a little bit of refactoring work.

packages/pages/src/common/src/template/types.ts
```diff
@@ -177,6 +177,39 @@ export interface Stream {
   };
 }
 
+/**
+ * The site stream config.
+ *
+ * @public
+ */
+export interface SiteStreamConfig {
```
mkilpatrick, line 185: Any reason you added this?

packages/pages/src/common/src/feature/stream.test.ts
```diff
@@ -45,3 +52,101 @@ describe("stream", () => {
...
+    expect(siteStream).toEqual({
+      id: "site-stream",
+      entityId: "site-stream-from-yaml",
+      fields: ["c_visualLayouts.c_visualConfiguration"],
+      localization: { locales: ["en"] },
+    });
+    fs.rmSync(path);
+  });
+
+  it("reads siteStream from config.yaml", () => {
```
mkilpatrick, line 134: "from site-stream.json" ... love that you added tests though! +2

packages/pages/src/common/src/feature/stream.test.ts
```diff
@@ -45,3 +52,101 @@ describe("stream", () => {
...
+      `
+    );
+    const siteStream = readSiteStream(projectStructure);
+    expect(siteStream).toEqual({
+      id: "site-stream",
+      entityId: "site-stream-from-yaml",
+      fields: ["c_visualLayouts.c_visualConfiguration"],
+      localization: { locales: ["en"] },
+    });
+    fs.rmSync(path);
```
mkilpatrick: If the tests fail will this still delete the file? I'm guessing no and we need to do some sort of after call or try/catch

packages/pages/src/common/src/feature/stream.ts
```diff
@@ -75,3 +115,54 @@ export const convertRedirectConfigToStreamConfig = (
...
+  }
+
+  return;
+};
+
+/**
+ * Converts the deprecated format of a siteStream specified in site-stream.json into
+ * the format of a siteStream specified in config.yaml
+ */
+export const formatSiteStream = (sitesJson: any, siteStreamPath: string) => {
```
mkilpatrick, line 152: Can we create a type for the site-stream.json format?

packages/pages/src/common/src/feature/stream.test.ts
```diff
@@ -96,6 +96,16 @@ describe("formatSiteStream", () => {
 });
 
 describe("readSiteStream", () => {
+  afterEach(() => {
+    if (fs.existsSync("config.yaml")) {
+      fs.rmSync("config.yaml");
+    }
+    if (fs.existsSync("sites-config/site-stream.json")) {
+      fs.rmSync("sites-config/site-stream.json");
```
mkilpatrick, line 104: You can delete the folder and everything in it in one line with `fs.rmSync(dir, { recursive: true, force: true })`

packages/pages/src/common/src/feature/stream.ts
benlife5: That should just be the Stream type right?

Also this formatter supported site stream $id's that were not "site-stream", is there any situation where that would be the case?

mkilpatrick (approved):

mkilpatrick (approved):

PR #547, opened 2024-10-23
fix(pages-plugins): prevent making api request with unbounded entities

jballschneider:
See issue identified at https://yexttest.atlassian.net/browse/TECHOPS-12762?focusedCommentId=1586488

### DIFF
--- a/packages/plugins/slugManager/manager.ts
+++ b/packages/plugins/slugManager/manager.ts
@@ -54,7 +54,9 @@ export function createManager(config: InternalSlugManagerConfig) {
     const entitiesResponse = await api.listEntities(params);
 
     if (!entitesResponse.entities) {
-      return JSON.stringify({});
+      return JSON.stringify({
+        data: []
+      });
     }
 
     const profileParams = new URLSearchParams({
--- a/packages/plugins/slugManager/manager.ts
+++ b/packages/plugins/slugManager/manager.ts
@@ -53,6 +53,10 @@ export function createManager(config: InternalSlugManagerConfig) {
     }
     const entitiesResponse = await api.listEntities(params);
 
+    if (!entitesResponse.entities) {
+      return JSON.stringify({});
+    }
+
     const profileParams = new URLSearchParams({
       fields: ["meta", ...slugGeneratorFnFields].join(","),
       filter: JSON.stringify({


mkilpatrick (approved):

packages/plugins/slugManager/manager.ts
```diff
@@ -53,6 +53,10 @@ export function createManager(config: InternalSlugManagerConfig) {
     }
     const entitiesResponse = await api.listEntities(params);
 
+    if (!entitesResponse.entities) {
+      return JSON.stringify({});
```
bhainesva, line 57: Is the connector happy with this not having a `data` property?

bhainesva (dismissed):

packages/plugins/slugManager/manager.ts
jballschneider: I have no idea. Do you know how connectors work?

msiddique-yext: This does not concern the connector but the [listLanguageProfiles api](https://hitchhikers.yext.com/docs/managementapis/content/entities/#operation/listAllLanguageProfiles)
In this case, api.listLanguageProfiles(line 69) will return all languageProfiles if the filter (line 62() is empty. 
The filter is empty when entitiesResponse.entities is empty. Since it results in incorrect behavior, we should return as Joe did in this CR. I don't know the effect of {} though.

jballschneider: Yeah, I was wondering about the returning the `{}` as the `connector` here, as that is ultimately returned via the exported `mod.ts` function.

bhainesva: I don't think it works: https://www.yext.com/s/3405717/addData/?stage=sourceSettings&source=function try using the DataSource>empty function and then using the "View Raw Response" button.

msiddique-yext: It depends on the connector code. I don't know how the connector method is handled. 
Slater owns the connector platform but we don't own how a specific connector will utilize this method.

bhainesva: yeah https://hitchhikers.yext.com/docs/connectors/function-source/?target=returning-data

I'd do `return JSON.stringify({"data":[]});`

mkilpatrick (approved):

bhainesva (approved):

PR #552, opened 2024-11-11
feat(dev/plugin): build without templates directory 

benlife5:
Also succeeds with a templates directory but no files in the directory

### DIFF
--- a/packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
+++ b/packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
@@ -91,7 +91,6 @@ export const getGlobalClientServerRenderTemplates = (
 
 /**
  * @param templatesRootPath the path where the templates live, typically src/templates
- * @param scopedTemplatePath the subfolder path inside templatesRoot to scope to - used in multibrand setups
  * @param globalFilename the file to find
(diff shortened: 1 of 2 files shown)


packages/pages/src/common/src/template/internal/getTemplateFilepaths.ts
```diff
@@ -91,7 +91,6 @@ export const getGlobalClientServerRenderTemplates = (
 
 /**
  * @param templatesRootPath the path where the templates live, typically src/templates
- * @param scopedTemplatePath the subfolder path inside templatesRoot to scope to - used in multibrand setups
```
benlife5, line 94: This param was not implemented -- not sure if it should be removed from the doc or implemented

mkilpatrick: I think it used to exist but was removed. Good to remove the docs.

mkilpatrick:
Can you verify that dev, build, and prod all work with this change?

benlife5:
> Can you verify that dev, build, and prod all work with this change?

👍

Tested with templates and with no templates but a serverless function in dev, local prod, platform prod

asanehisa (approved):

colton-demetriou (approved):

mkilpatrick (approved):

PR #553, opened 2024-11-14
fix(modules): update scaffold tailwind 

asanehisa:
Updating how scoping tailwind logic for modules work. Confirmed it works in local dev, local prod, and platform. Tested with multiple modules on one page as well ensuring that tailwind styles don't leak into inserted page.

Instead of needing 

```
.tailwind {
   @tailwind base;
}
```

in the index.css we are using the `tailwindcss-scoped-preflight` library in the `tailwind.config.ts ` to scope it to `tailwind tw-${moduleName}`.

So the module component will need to be wrapped by  

```
<div className="tailwind tw-${moduleName}">
  Module
</div>
```

### DIFF
--- a/packages/pages/src/scaffold/modules/templates.ts
+++ b/packages/pages/src/scaffold/modules/templates.ts
@@ -10,7 +10,7 @@ export const moduleCode = (
   moduleName: string,
   useTailwind: boolean
 ): string => {
-  const tailwind = useTailwind ? ` className="tailwind tw-${moduleName}"` : ``;
+  const tailwind = useTailwind ? ` className="tw-${moduleName}"` : ``;
   const formattedModuleName = formatModuleName(moduleName);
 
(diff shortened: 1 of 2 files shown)


packages/pages/src/scaffold/modules/templates.ts
```diff
@@ -10,7 +10,7 @@ export const moduleCode = (
   moduleName: string,
   useTailwind: boolean
 ): string => {
-  const tailwind = useTailwind ? ` className="tailwind"` : ``;
+  const tailwind = useTailwind ? ` className="tailwind tw-${moduleName}"` : ``;
```
mkilpatrick, line 13: Do we need the tailwind class too? Ideally it can just be the tw-{moduleName}

mkilpatrick (approved):

PR #556, opened 2024-11-25
fix: migrate eslint docs per version changes

asanehisa:
[Bug](https://github.com/eslint/eslint/issues/19134) introduced in eslint v9.15.0. Ended up not downgrading, but adding additonal rules to eslint docs per Jake's comment

Also followed this [migration guide](https://eslint.org/docs/latest/use/configure/migration-guide) so eslint would work


eslint.config.mjs
```diff
@@ -0,0 +1,71 @@
...
+  recommendedConfig: js.configs.recommended,
+  allConfig: js.configs.all,
+});
+
+export default [
+  {
+    ignores: [
+      "**/dist",
+      "packages/pages/tests/fixtures",
+      "playground/locations-site",
```
mkilpatrick, line 25: Can you just just ignore everything under playground/?

eslint.config.mjs
```diff
@@ -0,0 +1,71 @@
+import react from "eslint-plugin-react";
```
mkilpatrick, line 1: Does the json version no longer work?

jwartofsky-yext (approved):
lgtm

mkilpatrick (approved):

PR #561, opened 2025-01-02
fix(upgrade): pin Vite to version 5

mkilpatrick:


jwartofsky-yext (approved):
lgtm

Should we add latestMajorVersion for other packages like yext/search-ui-react?

mkilpatrick:
> lgtm
> 
> Should we add latestMajorVersion for other packages like yext/search-ui-react?

I think those are fine to leave but we can reassess if it ever becomes an issue.

benlife5 (approved):

PR #563, opened 2025-01-15
fix(scaffold): update VE template generation

mkilpatrick:
Realigns the VE template generation output against the latest changes we've made in the starter. This includes updating the main.tsx template, adding the theme.config.ts, adding utils/buildSchema.ts, updating tailwind.config.ts to use the themeConfig, and adding to .template-manifest.json.

Even though this updates the command, it also removes the VE template generation option until hybrid mode is better supported.


jwartofsky-yext (commented):
Could you add more info into the commit message about what the purpose of this change is?

packages/pages/src/scaffold/template/sampleTemplates.ts
```diff
@@ -153,6 +150,7 @@ export const getPath: GetPath<TemplateProps> = ({ document }) => {
 };
 
 const ${formattedTemplateName}: Template<TemplateRenderProps> = ({ document }) => {
+  console.log(document);
```
benlife5, line 153: remove

mkilpatrick: I added this on purpose so that the user sees they have access to the document and there isn't a warning that the document is unused.

packages/pages/src/scaffold/template/sampleTemplates.ts
```diff
@@ -1,74 +1,71 @@
...
-  additionalProperties: {
-    isVETemplate: true,
-  }
-};
-
-export const transformProps = async (data: TemplateRenderProps) => {
-  return resolveVisualEditorData(data, "${templateName}");
-};
+import { applyTheme, VisualEditorProvider } from "@yext/visual-editor";
+import { themeConfig } from "../../theme.config";
```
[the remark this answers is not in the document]
mkilpatrick: Good point. Adding them both in now if they don't exist. I'm sure the defaults in theme.config.ts will change but we can update that again later.

benlife5 (approved):

PR #564, opened 2025-01-17
fix(plugin): ignore templates marked for in platform page sets

benlife5:
If a template is listed in the .template-manifest.json, do not add it to the templates.json file we generate.
This will prevent a static page being generated for in-platform page sets.


packages/pages/src/common/src/project/structure.ts
```diff
@@ -313,6 +316,18 @@ export class ProjectStructure {
     );
   };
 
+  /**
+   * @returns the {@link Path} to the .template-manifest.json file, taking scope into account.
+   */
+  getTemplateManifestPath = () => {
+    return new Path(
+      pathLib.join(
+        this.config.scope ?? "",
```
mkilpatrick, line 325: Idk that we need to worry about scope for template-manifest. We've been treating it as root only and would only become a thing if we want to support multi-brand hybrid VLE use cases which I highly doubt. That being said, it won't hurt to be here.

mkilpatrick (approved):

packages/pages/src/common/src/project/structure.ts
benlife5: Yeah, that makes sense

PR #566, opened 2025-02-18
chore: upgrade vitest

asanehisa:


jwartofsky-yext (approved):

packages/pages/src/scaffold/template/sampleTemplates.test.ts
```diff
@@ -33,7 +33,7 @@ describe("newConfigFile", () => {
         fs.unlinkSync("test.tsx");
       }
     }
-  });
+  }, 10000);
```
mkilpatrick, line 36: 10s is pretty long. Do they need to be this long? 2s would be better if that works.

asanehisa: hmmm so they take 2-3s locally, but in the github workflow they were taking like 6-7 seconds.

mkilpatrick: That's wild, but ok.

mkilpatrick (approved):

PR #569, opened 2025-02-27
fix: revert filtering of static templates from template manifest

benlife5:
Partial revert of https://github.com/yext/pages/commit/4fe7b8eb376e2aebf9eacded510781e89ba446c6.

Filtering was causing page generation to fail in platform


jwartofsky-yext:
Could you include the error message that you were getting in platform before this revert?

Ideally we could get this working

benlife5:
Resolved by https://github.com/yext/pages/pull/571

PR #570, opened 2025-03-04
chore: update module-from-string

jwartofsky-yext:
This updates module-from-string to ^3.3.1

This doesn't mitigate the vulnerability however.

There isn't a newer version of this package that addresses the vulnerability.


mkilpatrick (commented):
There's no point to making this change since it doesn't fix the issue. I created an issue with the lib - https://github.com/exuanbo/module-from-string/issues/24

PR #571, opened 2025-03-04
fix: correctly ignore templates marked for in platform page sets

benlife5:
In https://github.com/yext/pages/pull/564, the in-platform page sets were filtered in loader.ts, which removed them from both the manifest.json and features.json. They are still needed in manifest.json so page generation was failing. This change only removes them from features.json. 

Tested in-platform and confirmed that in platform page set pages generate successfully and the static pages are no longer generated.


mkilpatrick (approved):

asanehisa (approved):
yay!

jwartofsky-yext (approved):
lgtm

PR #572, opened 2025-03-05
feat(dev): support visual editor local dev mode

benlife5:
Adds support for in-platform page sets to pages js dev sever

- When the dev server starts with the `--siteId` flag passed, the in-platform page sets are fetched for that site.
- When the index page is loaded, each up to 10 documents are fetched for each page set and listed in a table.
- When `--prod-url` is true (default), requests to slugs first try to look up the in-repo page. If that fails, then it attempts to fetch an in-platform document with that slug from each page set until a match is found.
- When `--no-prod-url` is true and when POST requests are made (occurs for the VE preview), the url is first parsed as `pageSetId/entityId` and it attempts to fetch an in-platform document for that combination. If it is not found, then the normal in-repo lookup occurs.

This is what previously didn't work in dev mode:
1. Live preview
   1. No way to inspect the actual output HTML document
2. No entity pages are served at localhost:5173
    1. the in platform page set pages do no appear on the index page
3. Testing changes to main.tsx
   1. Slug paths, head config, other wrappers
4. Testing adding a page set or changing a page set’s scope 
    1. (without it going live in production)

1, 2, and 3 now work

See https://github.com/yext/visual-editor/pull/251

https://github.com/user-attachments/assets/f7157d21-0409-42c5-9c7e-dc31e95d4537


packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -74,12 +88,43 @@ export const serverRenderRoute =
...
 
+      if (visualEditorOverrides) {
+        templateModuleInternal.config = {
+          templateType: "entity",
+          hydrate: true,
+          name: visualEditorOverrides.pageSet.id,
+          stream: {
+            $id: visualEditorOverrides.pageSet.id,
+            filter: {
+              entityTypes: visualEditorOverrides.pageSet.scope.entityTypes
```
mkilpatrick, line 99: Slightly more concise:
`visualEditorOverrides?.pageSet?.scope?.entityTypes?.map(scopeItem => scopeItem.name) ?? undefined`

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -74,12 +88,43 @@ export const serverRenderRoute =
...
+                .length
+                ? visualEditorOverrides.pageSet.scope.entityTypes.map(
+                    (scopeItem) => scopeItem.name
+                  )
+                : undefined,
+              savedFilterIds: visualEditorOverrides.pageSet.scope.savedFilters
```
mkilpatrick, line 105: Can do here too

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -74,12 +88,43 @@ export const serverRenderRoute =
         return;
       }
 
+      if (visualEditorOverrides) {
+        templateModuleInternal.config = {
```
mkilpatrick, line 92: Move this into its own function

packages/pages/src/dev/server/ssr/generateTestData.ts
```diff
@@ -113,6 +119,10 @@ export const generateTestDataForPage = async (
 
   args.push("--featureName", `"${featureName}"`);
 
+  if (visualEditorOverrides) {
+    args.push("--allFields");
```
mkilpatrick, line 123: Oh wow, didn't even realize this was an option.

packages/pages/src/dev/server/middleware/types.ts
```diff
@@ -0,0 +1,75 @@
...
+  codeTemplate: string;
+  /**
+   * Assignments define when to use this page set in new deployments for the site. They support the following:
+   * - GLOBAL to use this page set in all new deployments
+   * - PRODUCTION to use this page set in all new deployments for the production branch of the site
+   * - STAGING to use this page set in all new deployments for staging branches of the site
+   * - <branch name> to use this page set in all new deployments for the specified branch
+   * - <branch id> to use this page set in all new deployments for the specified branch
+   */
+  assignments: string[];
```
mkilpatrick, line 38: Idk if we should model out the full thing. These will diverge over time. I think we just model out the things that are relevant to pagesjs.

packages/pages/src/dev/server/ssr/findTemplateModuleInternal.ts
```diff
@@ -32,7 +33,10 @@ export const findTemplateModuleInternalByName = async (
       templateFilepath
     );
 
-    if (templateModuleInternal.config.templateType !== "entity") {
+    if (
+      !isInPlatformPageSet &&
```
mkilpatrick, line 37: Why do we need this check?

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -89,8 +134,12 @@ export const serverRenderRoute =
         return;
       }
 
-      const overrides = JSON.parse(req?.body?.overrides ?? "{}");
-      merge(document, overrides);
+      if (visualEditorOverrides) {
+        document.__.theme = visualEditorOverrides.theme ?? "{}";
+        document.__.layout =
```
mkilpatrick, line 139: How did this work before? Didn't the overrides just apply over the document? Why is it different now? Is it because overrides contains more things than it did before?

packages/pages/src/dev/server/ssr/findTemplateModuleInternal.ts
benlife5: At this point, it still thinks the in-repo template is a static template because it does not have a stream config. After we find it here, the stream config is added

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
benlife5: Yeah it used to just merge the document with whatever was send overrides. Overrides now has pageset, entityId, and locale, which wouldn't normally be fields on the document. 
I guess I could pass `{entityId: x, overrides: {}...}` from platform and then merge anything in that subkey into the document if we wanted to keep it more flexible

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
benlife5: getPageSet in YSS returns an empty array for entityTypes if they are not part of the page set scope, but generate-test-data interprets as "entity type must be in []" so we have to change empty arrays to undefined here which requires the longer code. As we modify generate-test-data, might be able to fix

packages/pages/src/dev/server/ssr/inPlatformPageSets.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (filters?.locale) {
+    args.push("--locale", filters.locale);
+  }
+  if (filters?.slug) {
+    args.push("--slug", filters.slug);
+  }
+
+  return await spawnPageSetCommands(
+    process.stdout,
+    "$ALPHA/bazel-bin/gocode/src/yext/cli/yext pages visual-editor document --xYextLocalCliGateway",
```
benlife5, line 58: Once the CLI is good to go, will update this to not use the local version

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -34,53 +47,82 @@ export const serverRenderRoute =
...
+          : await findTemplateModuleInternalByName(
+              vite,
+              feature,
+              templateFilepaths
+            );
       if (!templateModuleInternal) {
         send404(
           res,
-          `Cannot find template corresponding to feature: ${feature}`
+          `Cannot find template corresponding to feature: ${pageSet ? pageSet.code_template : feature}`
```
asanehisa, line 106: Cannot find template corresponding to feature: <pageSet.code_template> seems incorrect? More like "Cannot find template corresponding to <pageSet.code_template>" or "Cannot find template corresponding to feature <feature>" ?

benlife5: Good call

asanehisa (approved):

packages/pages/src/dev/server/server.ts
```diff
@@ -110,10 +115,15 @@ export const createServer = async (
   // Redirect urls with a final slash to their canonical url without the slash
   app.use(finalSlashRedirect);
 
-  // Call generateTestData to ensure we have data to populate the index page.
+  let inPlatformPageSets: PageSetConfig[] = [];
   // If the user specifies dynamicGenerateData = false we assume they have localData already.
   if (dynamicGenerateData) {
+    // Call generateTestData to ensure we have data to populate the index page.
     await generateTestData(projectStructure.config.scope);
+    if (siteId) {
```
mkilpatrick, line 123: Maybe there's a way around needing this?

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
```diff
@@ -34,53 +47,84 @@ export const serverRenderRoute =
...
-        vite,
-        templateFilepaths,
-        false,
-        staticURL,
-        locale
-      );
+      // First, match the feature name to an in-platform page set id
+      let pageSet: PageSetConfig | undefined = undefined;
+      for (const ps of inPlatformPageSets) {
+        if (ps.id === feature) {
```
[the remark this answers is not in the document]
benlife5: I think feature name is the page set id for in-platform page sets. OctopusServer.GetSiteConfig returns something like this:
```
...
"site_features": [
        {
          "name": "locations-directory-region",
          "stream_id": "63590-locations-directory-region-stream-n5s2avmpvz",
          "template_type": "JS",
          "entity_page_set": {},
          "additional_properties": {
            "displayName": "Locations Directory Region"
          },
          "code_template": "directory",
          "layout_required": true,
          "feature_origin": "PLATFORM"
        },
        ...
```
Regardless, this is just a variable name for the parsed value of x in :5173/x/entityId, so I can update to something else if that would be less confusing

packages/pages/src/dev/server/middleware/getPageSetConfig.ts
```diff
@@ -0,0 +1,31 @@
+import { VisualEditorPreviewOverrides } from "./types.js";
+import { TemplateConfigInternal } from "../../../common/src/template/internal/types.js";
+
+export const getPageSetConfig = (
+  visualEditorOverrides: VisualEditorPreviewOverrides
+): TemplateConfigInternal => {
+  return {
+    templateType: "entity",
+    hydrate: true,
+    name: visualEditorOverrides.pageSet.id,
```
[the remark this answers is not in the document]
benlife5: Ah whoops, this file was from an old attempt, will delete

packages/pages/src/dev/server/middleware/serverRenderSlugRoute.ts
```diff
@@ -47,18 +59,33 @@ export const serverRenderSlugRoute =
...
+      // First, try to get an entity document
+      let document = await getDocument(
         dynamicGenerateData,
         vite,
         slug,
         projectStructure
       );
+      // If the document is not found, and in-platform page sets are present,
+      // loop through each page set and attempt to get an in-platform-based document
+      if (!document && siteId && inPlatformPageSets.length) {
```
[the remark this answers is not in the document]
benlife5: Done

packages/pages/src/dev/server/server.ts
benlife5: I guess we could look up the page sets by account, but then we'd still need some way to disambiguate between multiple sites if they the same page set names

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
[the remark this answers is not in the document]
mkilpatrick: For in-repo templates/pagesets, the generated templates.json sets features.name to the name of the template. So I think you're right that in-platform pagesets the feature name is the pageset.id.  code_template relates to the name of the template since multiple pagesets share the same template. 

You're saying we need the pageset.id in the no-prod-url path to properly look things up? Like if we used code_template (which is how we have it for repo-based templates) then you wouldn't be able to determine the proper things?

packages/pages/src/dev/server/server.ts
mkilpatrick: pageset.Id should be unique though right?

packages/pages/src/dev/server/middleware/serverRenderRoute.ts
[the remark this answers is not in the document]
benlife5: Discussed, we need the page set id to get the right document, and a user could find the page set id in the editor's url path

mkilpatrick (dismissed):

benlife5:
The CLI has been released and works with these changes 👍

mkilpatrick (approved):

PR #574, opened 2025-03-27
fix: page set commands shell vulnerability 

benlife5:


asanehisa (approved):

packages/pages/src/dev/server/ssr/inPlatformPageSets.ts
```diff
@@ -55,8 +55,8 @@ export const getInPlatformPageSetDocuments = async (
 
   return await spawnPageSetCommands(
     process.stdout,
-    "yext pages visual-editor document",
-    args,
+    "yext",
+    ["pages", "visual-editor", "document", ...args],
```
mkilpatrick, line 59: Should we fix generateTestData too? Looks like it's the same. Double check that everything still works of course with shell: true no longer being set.

mkilpatrick: Hmm ok we can return to this later if necessary.

mkilpatrick (approved):

PR #579, opened 2025-05-02
fix(dev): dev server does not work with visual editor plugin

benlife5:
Feature generation occurs before we start the vite server but `edit.tsx` is created by the visual editor plugin _after_ we create the vite server. As a result, there is no `edit` static page generated and dev mode does not work

### DIFF
--- a/packages/pages/src/dev/server/server.ts
+++ b/packages/pages/src/dev/server/server.ts
@@ -25,6 +25,7 @@ import {
   getInPlatformPageSets,
   PageSetConfig,
 } from "./ssr/inPlatformPageSets.js";
+import runSubProcess from "../../util/runSubprocess.js";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const SERVERLESS_FUNCTION_POST_REQUEST_LIMIT = "50mb";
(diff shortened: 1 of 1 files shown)


packages/pages/src/dev/server/server.ts
```diff
@@ -118,12 +119,14 @@ export const createServer = async (
   let inPlatformPageSets: PageSetConfig[] = [];
   // If the user specifies dynamicGenerateData = false we assume they have localData already.
   if (dynamicGenerateData) {
-    // Call generateTestData to ensure we have data to populate the index page.
-    await generateTestData(projectStructure.config.scope);
     if (siteId) {
+      // If siteId is provided, regenerate templates to handle files created by the visual editor plugin
+      runSubProcess("pages generate templates", []);
```
benlife5, line 124: @mkilpatrick In `packages/pages/src/dev/dev.ts:53`, we handle a bunch of cases before we make this call (check for `config.yaml`, `scope` flag, `noGenFeatures` flag). Not sure if we wanted to respect those here (pretty easy to do but will add a bunch of arguments to `createServer`)

mkilpatrick: I know that plugin order matters. I think we needed the VE plugin to come last for some reason but we should verify that. Does swapping the order still work and fix things here? There's also the `enforce:post`  we could potentially add to the pagesjs plugin.

benlife5: The current flow is `pages generate templates` -> `createViteServer()` -> ve plugin adds the files as part of the `configureServer` step but we need to generate templates after the ve plugin runs. I wouldn't think there's a way for the vite plugin can run before the vite server is created

mkilpatrick: It makes sense what you're doing here, though I'm hoping we can figure out a better alternative.

benlife5:
https://github.com/yext/pages/pull/582

PR #580, opened 2025-05-05
WIP: feat(prod): add universe to __meta field

jwartofsky-yext:
Adds the universe field to __meta

### DIFF
--- a/packages/pages/src/common/src/template/types.ts
+++ b/packages/pages/src/common/src/template/types.ts
@@ -218,6 +218,7 @@ export interface TemplateProps<T = Record<string, any>> {
   __meta: {
     /** Specifies if the data is returned in development or production mode */
     mode: "development" | "production";
+    universe: "development" | "qa" | "sandbox" | "production";
   };
   /**
    * Set in the preview context of the generatepagecontent API endpoint. Since
--- a/packages/pages/src/common/src/module/types.ts
+++ b/packages/pages/src/common/src/module/types.ts
@@ -39,6 +39,7 @@ export interface ModuleProps<T = Record<string, any>> {
   __meta: {
     /** Specifies if the data is returned in development or production mode */
     mode: "development" | "production";
+    universe: "development" | "qa" | "sandbox" | "production";
   };
   /** The document to use for AnalyticsProvider */
   document: T;
--- a/packages/pages/src/common/src/template/head.test.ts
+++ b/packages/pages/src/common/src/template/head.test.ts
@@ -179,7 +179,10 @@ describe("getLang", () => {
       document: { locale: lang },
       path: "",
       relativePrefixToRoot: "",
-      __meta: { mode: "development" },
+      __meta: {
+        mode: "development",
+        universe: "production",
+      },
     };
 
     expect(getLang(undefined, props)).toEqual(lang);
@@ -196,5 +199,6 @@ const templateProps: TemplateRenderProps = {
   document: {},
   __meta: {
     mode: "development",
+    universe: "production",
   },
 };
--- a/packages/pages/src/dev/server/ssr/propsLoader.ts
+++ b/packages/pages/src/dev/server/ssr/propsLoader.ts
@@ -21,7 +21,10 @@ export const propsLoader = async ({
 
   let templateProps: TemplateProps = {
     document: document,
-    __meta: { mode: "development" },
+    __meta: {
+      mode: "development",
+      universe: "production", // TODO get from yext init
+    },
   };
 
   if (transformProps) {
--- a/packages/pages/src/scaffold/modules/templates.ts
+++ b/packages/pages/src/scaffold/modules/templates.ts
@@ -28,7 +28,8 @@ const templateData: ModuleProps = {
     }
   }, 
   __meta: {
-    mode: "production"
+    mode: "production",
+    universe: "production"
   }
 }
 
--- a/packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.test.tsx
+++ b/packages/pages/src/vite-plugin/build/buildStart/rendering/templateUtils.test.tsx
@@ -38,6 +38,7 @@ const baseProps: TemplateProps = {
   document: {},
   __meta: {
     mode: "development",
+    universe: "development",
   },
 };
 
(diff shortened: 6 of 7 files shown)


packages/pages/src/common/src/template/types.ts
```diff
@@ -218,6 +225,7 @@ export interface TemplateProps<T = Record<string, any>> {
   __meta: {
     /** Specifies if the data is returned in development or production mode */
     mode: "development" | "production";
+    universe: YextEnv;
```
asanehisa, line 228: include comment above

jwartofsky-yext: Done

packages/pages/src/dev/server/ssr/yextEnv.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (
+      typeof data === "object" &&
+      data !== null &&
+      "env" in data &&
+      typeof (data as any).env === "string"
+    ) {
+      const env = (data as { env: string }).env as string;
+
+      const validEnvs: YextEnv[] = [
+        "development",
```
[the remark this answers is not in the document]
jwartofsky-yext: Verified that these are correct for "development", "qa", "sandbox", and "production"

packages/pages/src/dev/server/ssr/yextEnv.ts
```diff
@@ -0,0 +1,61 @@
...
+      data = JSON.parse(content);
+    } catch (parseErr) {
+      console.error(
+        `Failed to parse JSON in ${filePath}:`,
+        (parseErr as Error).message
+      );
+      return "production";
+    }
+
+    if (
```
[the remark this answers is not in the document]
mkilpatrick: You can check that the env value is one of the 4 we know of if you want to by adding basically one more line. The unique error isn't really valuable as a user has no way to do anything about it. But you could log that still.

```
const env = data['env'];
if (!validEnvs.includes(env)) {
 console.log():
 return "production";
}
return env'
```

packages/pages/src/dev/server/ssr/yextEnv.ts
mkilpatrick, line 40: Do the as const thing next to the type to have a valid list and the types from it.

packages/pages/src/dev/server/ssr/yextEnv.ts
```diff
@@ -0,0 +1,61 @@
...
+      typeof data === "object" &&
```
mkilpatrick, line 32: But I'm really saying none of this type checking is necessary to simply get the value of a key from a json object. If the key doesn't exist you'll get undefined.

asanehisa: with this if data is undefined, will it error on undefined.env?

packages/pages/src/common/src/template/types.ts
```diff
@@ -205,12 +205,21 @@ export type Manifest = {
...
  * @internal
  */
-export type YextEnv = "development" | "qa" | "sandbox" | "production";
+export type YextEnv = (typeof YEXT_ENVS)[number];
+/**
+ * Returns true if the "env" string is a YextEnv
+ *
+ * @internal
+ */
+export function isYextEnv(env: string): env is YextEnv {
```
mkilpatrick, line 220: This would be a util and shouldn't be in types.ts

packages/pages/src/dev/server/ssr/yextEnv.ts
```diff
@@ -17,7 +17,7 @@ export function getEnvFromYextCredential(): YextEnv {
 
     const content = readFileSync(filePath, "utf8");
 
-    let data: unknown;
+    let data: any;
```
mkilpatrick, line 20: Record<string, string>

packages/pages/src/dev/server/ssr/yextEnv.ts
```diff
@@ -28,34 +28,23 @@ export function getEnvFromYextCredential(): YextEnv {
...
-      const env = (data as { env: string }).env as string;
-
-      const validEnvs: YextEnv[] = [
-        "development",
-        "qa",
-        "sandbox",
-        "production",
-      ];
-      if (validEnvs.includes(env as YextEnv)) {
+    const env = (data as { env: string }).env as string;
```
mkilpatrick, line 31: Again, just do `const env = data['env']`. It's valid json if we got here. Your type casting here doesn't provide value because you just do `as string` anyways, so regardless it will be treated as a string no matter what the object were to be.

packages/pages/src/dev/server/ssr/yextEnv.ts
```diff
@@ -28,34 +28,23 @@ export function getEnvFromYextCredential(): YextEnv {
...
+    if (env) {
```
mkilpatrick, line 32: There are still way too many conditions here for a simple _return value if exists and is valid, otherwise return a default_.

packages/pages/src/dev/server/ssr/yextEnv.ts
```diff
@@ -28,17 +28,12 @@ export function getEnvFromYextCredential(): YextEnv {
...
-    if (env) {
-      if (isYextEnv(env)) {
-        return env as YextEnv;
-      } else {
-        console.warn(`Invalid "env" value in ${filePath}: ${env}`);
-      }
-    } else {
-      console.warn(`"env" field not found in: ${filePath}`);
+    const env: string = data["env"];
+    if ((YEXT_ENVS as readonly string[]).includes(env)) {
```
mkilpatrick, line 32: Why do you need `as readonly string[]`?

jwartofsky-yext: If I remove that I see

'TS2345: Argument of type  string  is not assignable to parameter of type'

mkilpatrick: You right

mkilpatrick (approved):

benlife5 (approved):

jwartofsky-yext:
Also going to need to add "partition" here

Hopefully it works the same way

PR #584, opened 2025-06-21
chore(deps): swap from eslint to oxlint

mkilpatrick:


benlife5 (approved):

mkilpatrick:
> Is this just b/c oxlint is faster?

Faster and requires less config/deps.

PR #611, opened 2026-04-22
fix: getDocumentByLocale not filtering out unused templates

mkilpatrick:
`Multiple templates found for slug` was being returned when the same entity was in scope for two templates, even if they used different slug fields. This is because all templates were being looked at even if no documents were returned for them.


asanehisa (approved):

jwartofsky-yext (approved):
lgtm!

PR #616, opened 2026-06-08
feat: add reverse-proxy-prefix option to build

jwartofsky-yext:
After running "npm run build -- --reverse-proxy-prefix=www.brand.com/locations", I see these files modified:

**vite.config.ts**
```
import { defineConfig } from "vite";
import react from "@vitejs/plugin-react";
import yextSSG from "@yext/pages/vite-plugin";
import { yextVisualEditorPlugin } from "@yext/visual-editor/plugin"

export default defineConfig({
    plugins: [react(), yextVisualEditorPlugin(), yextSSG()],
    build: {
        assetsDir: "locations/assets"
    }
});
```

**config.yaml**
```
buildConfiguration:
  buildCommand: npm run build
  installDependenciesStep:
    command: npm install
    requiredFiles:
      - package.json
      - package-lock.json
      - .npmrc
livePreviewConfiguration:
  setupCommand: ":"
sitemap:
  excludeList:
    - edit
serving:
  reverseProxyPrefix: www.brand.com/locations
dynamicRoutes:
  - from: /assets/*
    to: /locations/assets/:splat
    status: 200
```

The **dist/assets** are also updated:
image

See example of a run here: https://github.com/YextSolutions/pages-visual-editor-starter/pull/330


packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -0,0 +1,216 @@
...
+      `Invalid reverseProxyPrefix "${reverseProxyPrefix}". Expected a non-empty subpath after the first slash.`
+    );
+  }
+
+  return {
+    reverseProxyPrefix,
+    assetsDir: `${subpath}/assets`,
+    dynamicRoute: {
+      from: "/assets/*",
+      to: `/${subpath}/assets/:splat`,
```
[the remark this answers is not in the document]
asanehisa: Seems like valid feedback?

packages/pages/src/prod/prod.ts
```diff
@@ -26,5 +41,9 @@ export const prodCommand = (program: Command) => {
     .option("--noBuild", "Disable build step")
     .option("--noRender", "Disable render step")
     .option("--scope  <string>", "The subfolder to scope from")
+    .option(
+      "--reverse-proxy-prefix <string>",
```
benlife5: The other options here are camel case but I think we're already a little inconsistent with kebab case vs. camel case.

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -0,0 +1,217 @@
+import fs from "node:fs";
+import path from "node:path";
+import { Node, ObjectLiteralExpression, Project, SyntaxKind } from "ts-morph";
+import YAML from "yaml";
+import logger from "../vite-plugin/log.js";
+
+export interface ReverseProxyOverride {
```
benlife5: Does this need to be exported?

packages/pages/src/prod/prod.ts
jwartofsky-yext: yes that is inconsistent

under build all of the options are kebab case:
```
...
    .command("build")
    .description("Build site using Vite")
    .option("--scope <string>", "The subfolder to scope from")
    .option("--plugin-filesize-limit <number>", "The max size of a single plugin file in MB", "10")
    .option(
      "--plugin-total-filesize-limit <number>",
      "The max size of all plugin files combined in MB",
      "10"
    )
...
```

Since this is a passthrough type of argument, I think it should match build

packages/pages/src/prod/prod.ts
```diff
@@ -5,16 +5,31 @@ interface ProdArgs {
...
   const command = "yext pages";
 
   if (!noBuild) {
-    await runSubprocess(command, ["build", scope ? `--scope ${scope}` : ""]);
+    const buildArgs = ["build"];
+    if (scope) {
+      buildArgs.push(`--scope ${scope}`);
+    }
+    if (reverseProxyPrefix) {
+      buildArgs.push(`--reverse-proxy-prefix ${reverseProxyPrefix}`);
```
mkilpatrick: yext pages does not have this flag. What is the intention? The only thing that needs to change is the build, therefore the buildCommand in config.yaml needs the flag.

packages/pages/src/util/reverseProxyOverride.ts
mkilpatrick: Make this a type instead of an interface

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -0,0 +1,217 @@
...
+ */
+export const applyReverseProxyOverride = (
+  scope: string | undefined,
+  reverseProxyPrefix: string
+): void => {
+  const finisher = logger.timedLog({
+    startLog: "Applying reverse proxy override",
+  });
+  const reverseProxyOverride = buildReverseProxyOverride(reverseProxyPrefix);
+  const configYamlPath = path.resolve(scope ?? "", "config.yaml");
```
mkilpatrick: You should use the functions of projectStructure - getConfigYamlPath, same for others

packages/pages/src/prod/prod.ts
jwartofsky-yext: This just allows you to run npm run prod with the reverse proxy prefix argument, which will pass it to as one of the build args.

I suppose we don't need this option available. I will remove it

mkilpatrick: You can add it to the npm prod command, but here this is passing the arg to the yext build command which is wrong. The yext pages prod cmd just ultimately calls the build cmd in the config.yaml.

mkilpatrick: You don't need to modify the buildCommand as I previously said

packages/pages/src/util/reverseProxyOverride.ts
jwartofsky-yext: projectStructure isn't created yet. ProjectStructure init reads the vite config, but since we are modifying the vite config it's a bit strange to call that first before modifying it.

I think it might be more confusing to use a "half-baked" project structure here just to call getConfigYamlPath. That might entice future devs to reuse this "project structure" which might have bad values for other fields.

I can extract this logic into a shared place and have the project structure flows reference it so we don't have random path logic here.

mkilpatrick: Ok sg, extract the logic

jwartofsky-yext: Done

Created "paths.ts" which has various helpers to get paths like getViteConfigPath(scope: string, options: PathOptions)

PathOptions can include fallbackToRoot, which is a common pattern we re-use in some places and don't want in others

packages/pages/src/common/src/project/structure.ts
```diff
@@ -299,14 +300,14 @@ export class ProjectStructure {
    * @returns the {@link Path} to the config.yaml file, taking scope into account.
    */
   getConfigYamlPath = () => {
-    return new Path(pathLib.join(this.config.scope ?? "", this.config.rootFiles.config));
+    return getConfigYamlPath(this.config.scope);
```
mkilpatrick: This doesn't technically use this.config.rootFiles.config anymore. I know in practice we only ever pass scope in so it's not exactly breaking, but it does break the current contract. Perhaps we should refactor ProjectStructure to only allow scope to come through?

packages/pages/src/common/src/project/paths.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+const getScopedFilePath = (fileName: string, scope?: string): Path => {
+  return new Path(pathLib.resolve(scope ?? "", fileName));
+};
+
+/**
+ * Returns the scoped config.yaml path.
+ */
+export const getConfigYamlPath = (scope?: string): Path => {
+  return getScopedFilePath("config.yaml", scope);
```
mkilpatrick: Based on the comment below, we probably want these filenames to come from this.config.rootFiles.config

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -2,34 +2,73 @@ import fs from "node:fs";
...
   }
 
-  const subpath = reverseProxyPrefix.substring(firstSlashIndex + 1);
+  let rawSubpath = "";
+  if (trimmedReverseProxyPrefix.includes("://")) {
+    try {
+      rawSubpath = new URL(trimmedReverseProxyPrefix).pathname;
+    } catch {
+      throw new Error(
+        `Invalid reverseProxyPrefix "${reverseProxyPrefix}". Expected a valid URL or host and subpath like "www.brand.com/locations".`
```
mkilpatrick: I think we want to enforce that there is no scheme passed through RPP instead of parsing it if it does

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -2,34 +2,73 @@ import fs from "node:fs";
...
+      );
+    }
+  } else {
+    try {
+      rawSubpath = new URL(`https://${trimmedReverseProxyPrefix}`).pathname;
+    } catch {
+      throw new Error(
+        `Invalid reverseProxyPrefix "${reverseProxyPrefix}". Expected a host and subpath like "www.brand.com/locations".`
```
mkilpatrick: Irrelevant if you change based on my comment above, but we DON'T expect a host

packages/pages/src/util/viteConfig.ts
```diff
@@ -1,17 +1,12 @@
 import { existsSync } from "node:fs";
-import path from "node:path";
-
-const VITE_CONFIG = "vite.config.js";
+import { getViteConfigPath } from "../common/src/project/paths.js";
 
 export const scopedViteConfigPath = (scope?: string) => {
```
mkilpatrick: Does this function do anything? You can just call getViteConfigPath(scope).getAbsolutePath() directly

jwartofsky-yext: It does some extra logic with existsSync

In an earlier patchset I put this logic inside of paths.ts, but that caused various issues with playwright because it became node-dependant

packages/pages/src/common/src/project/paths.ts
jwartofsky-yext: we would have to pass the project structure in in that case, in which case this should just live in structure.ts

I'm beginning to think that having these things share code is not the best idea

I'm looking at various options

packages/pages/src/util/reverseProxyOverride.ts
jwartofsky-yext: Do you mean that we don't expect the protocol, ex: "https://"?

The example from hitchhikers clearly has "www.brand.com/locations"

https://hitchhikers.yext.com/docs/pages/reverse-proxy/

image

packages/pages/src/common/src/project/paths.ts
jwartofsky-yext: Reverted so structure.ts is unchanged.

We will just keep the logic independent for now instead of a huge refactor

packages/pages/src/util/reverseProxyOverride.ts
jwartofsky-yext: Done

packages/pages/src/util/viteConfig.ts
jwartofsky-yext: Newest patchset removes all changes to this file

packages/pages/src/util/reverseProxyOverride.ts
mkilpatrick: Ah yea, sorry, got my terminology wrong there.

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -0,0 +1,250 @@
...
+export const applyReverseProxyOverride = (
+  scope: string | undefined,
+  reverseProxyPrefix: string
+): void => {
+  const finisher = logger.timedLog({
+    startLog: "Applying reverse proxy override",
+  });
+  const reverseProxyOverride = buildReverseProxyOverride(reverseProxyPrefix);
+  const configYamlPath = path.resolve(scope ?? "", "config.yaml");
+  const viteConfigPath = path.resolve(scope ?? "", "vite.config.js");
```
mkilpatrick: I think technically this is still wrong. A scope can be passed in but it'll fall back to root level if a scoped version doesn't exist. You could use scopedViteConfigPath in packages/pages/src/util/viteConfig.ts, though there isn't one specifically for the config.yaml.

Honestly, it might be simplest to just call this in build.ts: `const projectStructure = await ProjectStructure.init({scope});` and pass projectStructure to the code in this file. Update structure.ts to export `getViteConfigPath` and remove viteConfig.ts entirely and update all callers of `scopedViteConfigPath` to use `projectStructure.getViteConfigPath` since they all already have projectStructure.

It's a smaller refactor but not huge. Any reason not to do that?

jwartofsky-yext: ProjectStructure.init uses the vite config which this function is modifying

Creating the project structure early would be a bit of a chicken and egg problem. It might be more accurate to use new ProjectStructure instead since that doesn't read the repo contents. It would just be using default values and wouldn't help for vite.cofig.js path, but we could use it for the config.yaml path. The current setup is much cleaner IMO.

The current behavior fails if we pass a scope that does not exist. Do we want it to fallback and modify the root scope if the target scope doesn't exist?

```
In structure.ts, ProjectStructure.init(...) resolves a vite.config.js path and uses it to derive project structure values.
Specifically it uses the vite config to compute:
config.subfolders.assets via determineAssetsFilepath(...)
config.subfolders.public via determinePublicFilepath(...)
So new ProjectStructure(...) does not read the repo, but ProjectStructure.init(...) does use vite.config.js to fill in those derived paths.
```

packages/pages/src/common/src/project/structure.ts
```diff
@@ -302,6 +305,13 @@ export class ProjectStructure {
     return new Path(pathLib.join(this.config.scope ?? "", this.config.rootFiles.config));
   };
 
+  /**
+   * @returns the {@link Path} to the vite.config.js file, taking scope into account.
+   */
+  getViteConfigPath = () => {
```
mkilpatrick, line 315: The robot is right. This should return viteConfigPath from above.

mkilpatrick, line 315: Per my last comment, remove viteConfig.ts and update callers to use this func

mkilpatrick (approved):

packages/pages/src/util/viteConfig.ts
```diff
@@ -1,18 +0,0 @@
-import { existsSync } from "node:fs";
-import path from "node:path";
-
-const VITE_CONFIG = "vite.config.js";
-
-export const scopedViteConfigPath = (scope?: string) => {
```
asanehisa, line 6: is it safe to delete this exported function?

jwartofsky-yext: This function isn't exported publicly, so users of yext/pages weren't able to call it themselves

So it's safe to delete

PR #620, opened 2026-06-17
chore: adjust @types/node

asanehisa:
Upgraded `@types/node` too high in my last PR when it should align with lowest node version we support

### DIFF
--- a/packages/pages/package.json
+++ b/packages/pages/package.json
@@ -92,7 +92,7 @@
     "@types/glob": "^8.1.0",
     "@types/lodash": "^4.17.0",
     "@types/mime-types": "^2.1.4",
-    "@types/node": "^25.9.3",
+    "@types/node": "^20.0.0",
     "@types/prompts": "^2.4.9",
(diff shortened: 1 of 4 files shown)


packages/pages/package.json
```diff
@@ -92,7 +92,7 @@
     "@types/glob": "^8.1.0",
     "@types/lodash": "^4.17.0",
     "@types/mime-types": "^2.1.4",
-    "@types/node": "^25.9.3",
+    "@types/node": "^20.0.0",
```
mkilpatrick, line 95: Should this be the latest minor? [20.19.43](https://www.npmjs.com/package/@types/node/v/20.19.43)

mkilpatrick (approved):

packages/pages/package.json
asanehisa: Yup latest minor works! It's pinned to ^20.0.0 so it uses the latest minor in major 20

mkilpatrick (approved):

PR #624, opened 2026-07-23
fix: apply reverse proxy prefix to config

a-friedman:
The previous logic loaded the vite configuration with `ProjectStructure.init` before applying the reverse proxy prefix, so asset paths did not have the correct prefix and returned a 404. Update to work properly.

TEST=auto

Ran unit tests

### DIFF
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -1,6 +1,7 @@
 import { Command } from "commander";
 import { build } from "vite";
 import { ProjectStructure } from "../common/src/project/structure.js";
+import { applyReverseProxy } from "../util/applyReverseProxy.js";
 
 /**
  * The arguments passed to the build CLI command.
@@ -19,6 +20,7 @@ export interface BuildArgs {
 const handler = async (buildArgs: BuildArgs) => {
   const { scope, pluginFilesizeLimit, pluginTotalFilesizeLimit, reverseProxyPrefix } = buildArgs;
   const projectStructure = await ProjectStructure.init({ scope }, reverseProxyPrefix);
+  applyReverseProxy(projectStructure);
 
   // Pass CLI arguments as env variables to use in vite-plugin
   if (scope) {
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -1,7 +1,10 @@
 import { Command } from "commander";
 import { build } from "vite";
 import { ProjectStructure } from "../common/src/project/structure.js";
-import { applyReverseProxyOverride } from "../util/reverseProxyOverride.js";
+import {
+  applyReverseProxyOverride,
+  buildReverseProxyOverride,
+} from "../util/reverseProxyOverride.js";
 
 /**
  * The arguments passed to the build CLI command.
@@ -20,10 +23,16 @@ export interface BuildArgs {
 const handler = async (buildArgs: BuildArgs) => {
   const { scope, pluginFilesizeLimit, pluginTotalFilesizeLimit, reverseProxyPrefix } = buildArgs;
   const trimmedReverseProxyPrefix = reverseProxyPrefix?.trim();
-  const projectStructure = await ProjectStructure.init({ scope });
+  const projectStructure = await ProjectStructure.init({
+    scope,
+    reverseProxyPrefix: trimmedReverseProxyPrefix || undefined,
+  });
 
-  if (trimmedReverseProxyPrefix) {
-    applyReverseProxyOverride(projectStructure, trimmedReverseProxyPrefix);
+  if (projectStructure.config.reverseProxyPrefix) {
+    const reverseProxyOverride = buildReverseProxyOverride(
+      projectStructure.config.reverseProxyPrefix
+    );
+    applyReverseProxyOverride(projectStructure, reverseProxyOverride);
   }
 
   // Pass CLI arguments as env variables to use in vite-plugin
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -1,6 +1,6 @@
 import { Command } from "commander";
 import { build } from "vite";
-import { ProjectStructure } from "../common/src/project/structure.js";
+import { parseReverseProxyPrefix, ProjectStructure } from "../common/src/project/structure.js";
 import { applyReverseProxy } from "../util/applyReverseProxy.js";
 
 /**
@@ -19,8 +19,8 @@ export interface BuildArgs {
  */
 const handler = async (buildArgs: BuildArgs) => {
   const { scope, pluginFilesizeLimit, pluginTotalFilesizeLimit, reverseProxyPrefix } = buildArgs;
-  const projectStructure = await ProjectStructure.init({ scope }, reverseProxyPrefix);
-  applyReverseProxy(projectStructure);
+  const parsedReverseProxyPrefix =
+    reverseProxyPrefix === undefined ? undefined : parseReverseProxyPrefix(reverseProxyPrefix);
 
   // Pass CLI arguments as env variables to use in vite-plugin
   if (scope) {
@@ -28,6 +28,9 @@ const handler = async (buildArgs: BuildArgs) => {
   }
   process.env.YEXT_PAGES_PLUGIN_FILESIZE_LIMIT = String(pluginFilesizeLimit);
   process.env.YEXT_PAGES_PLUGIN_TOTAL_FILESIZE_LIMIT = String(pluginTotalFilesizeLimit);
+
+  await applyReverseProxy(scope, parsedReverseProxyPrefix);
+  const projectStructure = await ProjectStructure.init({ scope });
 
   await build({
     configFile: projectStructure.getViteConfigPath()?.getAbsolutePath(),
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -19,8 +19,7 @@ export interface BuildArgs {
  */
 const handler = async (buildArgs: BuildArgs) => {
   const { scope, pluginFilesizeLimit, pluginTotalFilesizeLimit, reverseProxyPrefix } = buildArgs;
-  const parsedReverseProxyPrefix =
-    reverseProxyPrefix === undefined ? undefined : parseReverseProxyPrefix(reverseProxyPrefix);
+  const parsedReverseProxyPrefix = parseReverseProxyPrefix(reverseProxyPrefix);
 
   // Pass CLI arguments as env variables to use in vite-plugin
   if (scope) {
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -1,7 +1,7 @@
 import { Command } from "commander";
 import { build } from "vite";
-import { parseReverseProxyPrefix, ProjectStructure } from "../common/src/project/structure.js";
-import { applyReverseProxy } from "../util/applyReverseProxy.js";
+import { ProjectStructure } from "../common/src/project/structure.js";
+import { applyReverseProxy, parseReverseProxyPrefix } from "../util/applyReverseProxy.js";
 
 /**
  * The arguments passed to the build CLI command.
--- a/packages/pages/src/build/build.ts
+++ b/packages/pages/src/build/build.ts
@@ -1,10 +1,6 @@
 import { Command } from "commander";
 import { build } from "vite";
 import { ProjectStructure } from "../common/src/project/structure.js";
-import {
-  applyReverseProxyOverride,
-  buildReverseProxyOverride,
-} from "../util/reverseProxyOverride.js";
 
 /**
  * The arguments passed to the build CLI command.
@@ -22,18 +18,7 @@ export interface BuildArgs {
  */
 const handler = async (buildArgs: BuildArgs) => {
   const { scope, pluginFilesizeLimit, pluginTotalFilesizeLimit, reverseProxyPrefix } = buildArgs;
-  const trimmedReverseProxyPrefix = reverseProxyPrefix?.trim();
-  const projectStructure = await ProjectStructure.init({
-    scope,
-    reverseProxyPrefix: trimmedReverseProxyPrefix || undefined,
-  });
-
-  if (projectStructure.config.reverseProxyPrefix) {
-    const reverseProxyOverride = buildReverseProxyOverride(
-      projectStructure.config.reverseProxyPrefix
-    );
-    applyReverseProxyOverride(projectStructure, reverseProxyOverride);
-  }
+  const projectStructure = await ProjectStructure.init({ scope }, reverseProxyPrefix);
 
   // Pass CLI arguments as env variables to use in vite-plugin
   if (scope) {
--- a/packages/pages/src/common/src/project/structure.ts
+++ b/packages/pages/src/common/src/project/structure.ts
@@ -2,7 +2,11 @@ import pathLib from "node:path";
 import merge from "lodash/merge.js";
 import fs from "node:fs";
 import { Path } from "./path.js";
-import { parseReverseProxyPrefix } from "../../../util/reverseProxyOverride.js";
+import {
+  applyReverseProxyOverride,
+  buildReverseProxyOverride,
+  type ReverseProxyOverride,
+} from "../../../util/reverseProxyOverride.js";
 import { determineAssetsFilepath } from "../assets/getAssetsFilepath.js";
 import { determinePublicFilepath } from "../assets/getPublicFilepath.js";
 
@@ -143,9 +147,9 @@ export interface ProjectStructureConfig {
   scope?: string;
 
   /**
-   * The reverse proxy prefix for the current build.
+   * The build configuration derived from the reverse proxy prefix.
    */
-  reverseProxyPrefix?: string;
+  reverseProxyOverride?: ReverseProxyOverride;
 }
 
 const DEFAULT_ASSETS_DIR = "assets";
@@ -221,16 +225,20 @@ export class ProjectStructure {
     this.config = mergedConfig;
   }
 
-  static init = async (projectStructureConfig?: Optional<ProjectStructureConfig>) => {
+  static init = async (
+    projectStructureConfig?: Optional<ProjectStructureConfig>,
+    reverseProxyPrefix?: string
+  ) => {
     const projectStructure = new ProjectStructure(projectStructureConfig);
     const config = projectStructure.config;
 
     const viteConfigPath = projectStructure.getViteConfigPath()?.getAbsolutePath() ?? "";
 
-    if (config.reverseProxyPrefix) {
-      const parsedReverseProxyPrefix = parseReverseProxyPrefix(config.reverseProxyPrefix);
-      config.reverseProxyPrefix = parsedReverseProxyPrefix.reverseProxyPrefix;
-      config.subfolders.assets = `${parsedReverseProxyPrefix.subpath}/assets`;
+    if (reverseProxyPrefix) {
+      const reverseProxyOverride = buildReverseProxyOverride(reverseProxyPrefix);
+      config.reverseProxyOverride = reverseProxyOverride;
+      config.subfolders.assets = reverseProxyOverride.assetsDir;
+      applyReverseProxyOverride(projectStructure);
     } else {
       // TODO: handle other extensions
       const assetsDir = await determineAssetsFilepath(DEFAULT_ASSETS_DIR, viteConfigPath);
--- a/packages/pages/src/common/src/project/structure.ts
+++ b/packages/pages/src/common/src/project/structure.ts
@@ -2,11 +2,6 @@ import pathLib from "node:path";
 import merge from "lodash/merge.js";
 import fs from "node:fs";
 import { Path } from "./path.js";
-import {
-  applyReverseProxyOverride,
-  buildReverseProxyOverride,
-  type ReverseProxyOverride,
-} from "../../../util/reverseProxyOverride.js";
 import { determineAssetsFilepath } from "../assets/getAssetsFilepath.js";
 import { determinePublicFilepath } from "../assets/getPublicFilepath.js";
 
@@ -118,6 +113,24 @@ export interface EnvVar {
 }
 
 /**
+ * Build configuration derived from a reverse proxy prefix.
+ */
+export type ReverseProxyOverride = {
+  reverseProxyPrefix: string;
+  assetsDir: string;
+  dynamicRoute: {
+    from: string;
+    to: string;
+    status: number;
+  };
+};
+
+export type ParsedReverseProxyPrefix = {
+  reverseProxyPrefix: string;
+  subpath: string;
+};
+
+/**
  * The configuration structure of a project.
  *
  * @public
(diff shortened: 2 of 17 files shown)


packages/pages/src/util/applyReverseProxy.ts
```diff
@@ -98,7 +115,7 @@ export const applyReverseProxyOverride = (
   }
 
   updateConfigYaml(configYamlPath, reverseProxyOverride);
-  updateViteConfig(viteConfigPath, reverseProxyOverride.assetsDir);
+  updateViteConfig(viteConfigPath, projectStructure.config.subfolders.assets);
```
jwartofsky-yext: Do we not want the reverseProxyOverride assetsDir here?

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -64,6 +68,20 @@ export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReversePr
...
+ * Validates and parses a reverse proxy prefix into the concrete build-time
+ * values needed to initialize the project and update its configuration.
+ */
+export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReverseProxyOverride => {
+  const parsedReverseProxyPrefix = parseReverseProxyPrefix(reverseProxyPrefix);
+  const { subpath } = parsedReverseProxyPrefix;
+
+  return {
+    reverseProxyPrefix: parsedReverseProxyPrefix.reverseProxyPrefix,
     assetsDir: `${subpath}/assets`,
```
mkilpatrick: This should really come from the projectStructure too because right now it's doing some hardcoding of 'assets'

packages/pages/src/common/src/project/structure.ts
```diff
@@ -211,24 +217,30 @@ export class ProjectStructure {
...
+    const projectStructure = new ProjectStructure(projectStructureConfig);
+    const config = projectStructure.config;
 
     const viteConfigPath = projectStructure.getViteConfigPath()?.getAbsolutePath() ?? "";
 
-    // TODO: handle other extensions
-    const assetsDir = await determineAssetsFilepath(DEFAULT_ASSETS_DIR, viteConfigPath);
+    if (config.reverseProxyPrefix) {
+      const parsedReverseProxyPrefix = parseReverseProxyPrefix(config.reverseProxyPrefix);
+      config.reverseProxyPrefix = parsedReverseProxyPrefix.reverseProxyPrefix;
```
mkilpatrick: This is already set

packages/pages/src/build/build.ts
```diff
@@ -20,10 +23,16 @@ export interface BuildArgs {
 const handler = async (buildArgs: BuildArgs) => {
   const { scope, pluginFilesizeLimit, pluginTotalFilesizeLimit, reverseProxyPrefix } = buildArgs;
   const trimmedReverseProxyPrefix = reverseProxyPrefix?.trim();
-  const projectStructure = await ProjectStructure.init({ scope });
+  const projectStructure = await ProjectStructure.init({
```
mkilpatrick: Maybe we should take in the RPP as is (non-trimmed, which can happen in the PS.init) and we store the RPP payload and store it in on the PS. It feels weird that the PS requires the RPP but then we separately do other RPP things we need to pass around. Since RPP actually influences the PS I think the PS should control everything about it.

packages/pages/src/util/applyReverseProxy.ts
a-friedman: This will be the reverse proxy assetsDir. Part of this change is to update the ProjectStructure config to have the reverse proxy included.

packages/pages/src/util/reverseProxyOverride.ts
a-friedman: Refactored to include the full reverseProxyOverride object on the projectStructure config

packages/pages/src/common/src/project/structure.ts
```diff
@@ -2,7 +2,11 @@ import pathLib from "node:path";
 import merge from "lodash/merge.js";
 import fs from "node:fs";
 import { Path } from "./path.js";
-import { parseReverseProxyPrefix } from "../../../util/reverseProxyOverride.js";
+import {
+  applyReverseProxyOverride,
+  buildReverseProxyOverride,
+  type ReverseProxyOverride,
```
mkilpatrick: Let's keep this type local to this file

packages/pages/src/common/src/project/structure.ts
mkilpatrick: Aren't we losing this now?

packages/pages/src/common/src/project/structure.ts
```diff
@@ -221,16 +225,20 @@ export class ProjectStructure {
...
 
-    if (config.reverseProxyPrefix) {
-      const parsedReverseProxyPrefix = parseReverseProxyPrefix(config.reverseProxyPrefix);
-      config.reverseProxyPrefix = parsedReverseProxyPrefix.reverseProxyPrefix;
-      config.subfolders.assets = `${parsedReverseProxyPrefix.subpath}/assets`;
+    if (reverseProxyPrefix) {
+      const reverseProxyOverride = buildReverseProxyOverride(reverseProxyPrefix);
+      config.reverseProxyOverride = reverseProxyOverride;
+      config.subfolders.assets = reverseProxyOverride.assetsDir;
+      applyReverseProxyOverride(projectStructure);
```
mkilpatrick: I'm debating if this (apply) should live here or not. I'm sort of thinking it should live back in build.ts because it is doing more than just creating the project structure.

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -95,10 +95,12 @@ export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReversePr
...
  * read so the normal build pipeline picks up the reverse proxy override.
  */
-export const applyReverseProxyOverride = (
-  projectStructure: ProjectStructure,
-  reverseProxyOverride: ReverseProxyOverride
-): void => {
+export const applyReverseProxyOverride = (projectStructure: ProjectStructure): void => {
+  const reverseProxyOverride = projectStructure.config.reverseProxyOverride;
+  if (!reverseProxyOverride) {
+    throw new Error("Cannot apply reverse proxy override because none is configured.");
```
mkilpatrick: I wonder if we should always just have this called and noop instead of erroring.

packages/pages/src/util/reverseProxyOverride.test.ts
```diff
@@ -302,10 +301,7 @@ describe("applyReverseProxyOverride", () => {
       );
       process.chdir(tempDir);
 
-      applyReverseProxyOverride(
-        new ProjectStructure({ scope: "brand" }),
-        "www.brand.com/locations"
-      );
+      await ProjectStructure.init({ scope: "brand" }, "www.brand.com/locations");
```
mkilpatrick: This file should test things in reverseProxyOverride, not ProjectStructure

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -16,10 +21,9 @@ type ReverseProxyOverride = {
 };
 
 /**
- * Parses a reverse proxy prefix into the concrete build-time values needed to
- * update config.yaml and vite.config.js.
+ * Validates and parses a reverse proxy prefix into its normalized value and subpath.
  */
-export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReverseProxyOverride => {
+export const parseReverseProxyPrefix = (reverseProxyPrefix: string): ParsedReverseProxyPrefix => {
```
mkilpatrick: Add tests

packages/pages/src/util/reverseProxyOverride.ts
mkilpatrick: This is still hardcoded.

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -64,6 +68,20 @@ export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReversePr
...
+ */
+export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReverseProxyOverride => {
+  const parsedReverseProxyPrefix = parseReverseProxyPrefix(reverseProxyPrefix);
+  const { subpath } = parsedReverseProxyPrefix;
+
+  return {
+    reverseProxyPrefix: parsedReverseProxyPrefix.reverseProxyPrefix,
     assetsDir: `${subpath}/assets`,
     dynamicRoute: {
       from: "/assets/*",
```
mkilpatrick: Not you, but these are also hardcoded and don't reflect the assetsDir of the PS.

packages/pages/src/util/reverseProxyOverride.ts
```diff
@@ -64,6 +68,20 @@ export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReversePr
...
     reverseProxyPrefix: trimmedReverseProxyPrefix,
+    subpath,
+  };
+};
+
+/**
+ * Validates and parses a reverse proxy prefix into the concrete build-time
+ * values needed to initialize the project and update its configuration.
+ */
+export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReverseProxyOverride => {
```
mkilpatrick: Given my comments below, I think it makes sense to move these parsing functions into the structure.ts and have this file only really contain applyReverseProxyOverride (renamed to applyReverseProxy as well the filename...and corresponding tests/filenames).

jwartofsky-yext (dismissed):

packages/pages/src/util/reverseProxyOverride.ts
a-friedman: Changed to take projectStructure as an arg and build the paths from that

packages/pages/src/util/reverseProxyOverride.ts
a-friedman: Done

packages/pages/src/util/reverseProxyOverride.ts
a-friedman: Done

packages/pages/src/util/reverseProxyOverride.ts
a-friedman: Done

packages/pages/src/util/reverseProxyOverride.test.ts
a-friedman: Done

packages/pages/src/util/reverseProxyOverride.ts
a-friedman: Done

packages/pages/src/common/src/project/structure.ts
a-friedman: Done

packages/pages/src/common/src/project/structure.ts
a-friedman: I'm not very familiar with Vite/PagesJS, but codex seems to think this is the correct solution. An excerpt from its output:

- assetsDir is a Vite output path. RPP changes it from assets to blog/assets.
- publicDir is a Vite source directory. RPP does not change where public files are read from.
The RPP branch intentionally avoids determining either value from vite.config.js, because doing so imports the config too early. It derives the new assets output from the structure’s current assets value. Later, when Vite loads the rewritten config, the plugin initializes another ProjectStructure and determines the real publicDir.
So the initial RPP structure is temporarily incomplete regarding a custom publicDir, but nothing consumes that field before Vite reinitializes it.

Does that sound correct to you?

packages/pages/src/common/src/project/structure.ts
a-friedman: Done

packages/pages/src/common/src/project/structure.ts
```diff
@@ -140,12 +158,93 @@ export interface ProjectStructureConfig {
...
+
+  if (!trimmedReverseProxyPrefix.includes("/")) {
+    throw new Error(
+      `Invalid reverseProxyPrefix "${reverseProxyPrefix}". Expected a host and subpath like "www.brand.com/locations".`
+    );
+  }
+
+  const subpathAfterHost = trimmedReverseProxyPrefix.substring(
+    trimmedReverseProxyPrefix.indexOf("/") + 1
+  );
```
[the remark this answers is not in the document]
mkilpatrick: Heed the robots

packages/pages/src/common/src/project/structure.ts
```diff
@@ -211,24 +310,33 @@ export class ProjectStructure {
...
+    reverseProxyPrefix?: string
+  ) => {
+    const projectStructure = new ProjectStructure(projectStructureConfig);
+    const config = projectStructure.config;
 
     const viteConfigPath = projectStructure.getViteConfigPath()?.getAbsolutePath() ?? "";
 
-    // TODO: handle other extensions
-    const assetsDir = await determineAssetsFilepath(DEFAULT_ASSETS_DIR, viteConfigPath);
+    if (reverseProxyPrefix) {
```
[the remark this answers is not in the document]
mkilpatrick: I think this is fine to leave as is

packages/pages/src/common/src/project/structure.ts
```diff
@@ -112,6 +112,24 @@ export interface EnvVar {
...
+  reverseProxyPrefix: string;
+  assetsDir: string;
+  dynamicRoute: {
+    from: string;
+    to: string;
+    status: number;
+  };
+};
+
+export type ParsedReverseProxyPrefix = {
```
mkilpatrick: Don't export this since it's only used locally to this file.

packages/pages/src/common/src/project/structure.ts
mkilpatrick: I semi follow what it's saying. The publicDir is determined by the vite.config.js (or falls back to the default). The RP flow has no bearing on that, which is true. I _think_ what it's saying is that we end up rewriting the vite config file when there is an RP and somewhere later we init PS again, so it has no real effect to do it now. But it'd be correct to determine it regardless of RPP. I think you should just do this outside of the if/else case.

But separately, I'm pretty sure we'll need to pass RPP to all usages of ProjectStructure.init. For instance, `packages/pages/src/vite-plugin/plugin.ts` is the actual build plugin that gets called from the underlying vite js build command in src/build/build.ts and it doesn't use the RPP. It might technically be fine because by then we've already rewritten the vite.config.js so downstream everything works, but it's also weird that the side effect just happens to work. It feels brittle.

I keep debating on changing it back or not. There's no real great way here IMO due to the chicken and egg problem. So the tldr for this is to just do the first thing I said and move this outside of the if/else and be done with it. No other PS.inits need to be changed.

packages/pages/src/common/src/project/structure.ts
a-friedman: Done

packages/pages/src/common/src/project/structure.ts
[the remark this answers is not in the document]
a-friedman: Done

packages/pages/src/common/src/project/structure.ts
a-friedman: Discussed on slack, new commit implements as discussed

packages/pages/src/util/applyReverseProxy.ts
```diff
@@ -97,6 +38,15 @@ export const applyReverseProxyOverride = (
...
+    projectStructure.config.subfolders.assets,
+    viteConfigPath
+  );
+  const reverseProxyOverride = buildReverseProxyOverride(
+    originalAssetsPath,
+    parsedReverseProxyPrefix
+  );
+
   updateConfigYaml(configYamlPath, reverseProxyOverride);
   updateViteConfig(viteConfigPath, reverseProxyOverride.assetsDir);
```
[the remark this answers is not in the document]
mkilpatrick: Yea this would be bad

packages/pages/src/util/applyReverseProxy.ts
```diff
@@ -97,6 +38,15 @@ export const applyReverseProxyOverride = (
     throw new Error(`Cannot apply reverseProxyPrefix because ${viteConfigPath} does not exist.`);
   }
 
+  const originalAssetsPath = await determineAssetsFilepath(
```
mkilpatrick: This already gets called in PS so you should just be able to use PS.config.subfolders.assets directly.

packages/pages/src/common/src/project/structure.ts
```diff
@@ -146,6 +164,79 @@ const DEFAULT_ASSETS_DIR = "assets";
 
 const DEFAULT_PUBLIC_DIR = "public";
 
+/**
+ * Validates and parses a reverse proxy prefix into its normalized value and subpath.
+ */
+export const parseReverseProxyPrefix = (reverseProxyPrefix: string): ParsedReverseProxyPrefix => {
```
mkilpatrick: All of this code doesn't need to be part of structure.ts anymore and you can move it back to applyReverseProxy.

packages/pages/src/common/src/project/structure.ts
```diff
@@ -211,24 +302,20 @@ export class ProjectStructure {
...
     // TODO: handle other extensions
-    const assetsDir = await determineAssetsFilepath(DEFAULT_ASSETS_DIR, viteConfigPath);
-
-    config.subfolders.assets = assetsDir;
-
-    const publicDir = await determinePublicFilepath(DEFAULT_PUBLIC_DIR, viteConfigPath);
+    config.subfolders.assets = await determineAssetsFilepath(DEFAULT_ASSETS_DIR, viteConfigPath);
 
-    config.subfolders.public = publicDir;
+    config.subfolders.public = await determinePublicFilepath(DEFAULT_PUBLIC_DIR, viteConfigPath);
```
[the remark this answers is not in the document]
a-friedman: Done

packages/pages/src/util/applyReverseProxy.ts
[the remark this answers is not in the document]
a-friedman: Done

packages/pages/src/common/src/project/structure.ts
a-friedman: Done

packages/pages/src/util/applyReverseProxy.ts
a-friedman: The ProjectStructure here is made through the constructor, not `init`, more as just a clean way to get the configYamlPath/viteConfigPath/assetPath, so determineAssetsFilepath isn't called as part of it. Also, as part of the refactor in the most recent commit, this was updated to use a new function that doesn't execute the vite config to get the asset filepath. The two asset reads (this one and the one in PS.init) have different purposes. This one is to get the original asset path without executing the config and update it to have the correct RPP. The init one is to set the PS config value to the proper RPPed asset path written by this one

packages/pages/src/util/fileContentHash.ts
```diff
@@ -0,0 +1,10 @@
+import fs from "node:fs";
+
+/**
+ * Returns a stable hash of a file's contents for use in module cache keys.
+ */
+export const getFileContentHash = async (filepath: string): Promise<string> => {
+  const contents = new TextEncoder().encode(fs.readFileSync(filepath, "utf-8"));
```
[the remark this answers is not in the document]
mkilpatrick: Sounds like we should do this. Just curious, why did you need to change to the globalThis way?

packages/pages/src/build/build.ts
```diff
@@ -19,12 +19,8 @@ export interface BuildArgs {
...
 const handler = async (buildArgs: BuildArgs) => {
   const { scope, pluginFilesizeLimit, pluginTotalFilesizeLimit, reverseProxyPrefix } = buildArgs;
-  const trimmedReverseProxyPrefix = reverseProxyPrefix?.trim();
-  const projectStructure = await ProjectStructure.init({ scope });
-
-  if (trimmedReverseProxyPrefix) {
-    applyReverseProxyOverride(projectStructure, trimmedReverseProxyPrefix);
-  }
+  const parsedReverseProxyPrefix =
+    reverseProxyPrefix === undefined ? undefined : parseReverseProxyPrefix(reverseProxyPrefix);
```
mkilpatrick: Probably makes sense to just have parseReverseProxyPrefix return undefined

mkilpatrick (commented):
otherwise lg

packages/pages/src/util/fileContentHash.ts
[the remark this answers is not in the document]
a-friedman: Done

By using `globalThis`, we use the built-in Web Crypto API, which avoids importing `node:crypto`. This improves efficiency since we avoid importing large polyfills.

packages/pages/src/build/build.ts
a-friedman: Done

packages/pages/src/util/fileContentHash.ts
```diff
@@ -0,0 +1,10 @@
+import fs from "node:fs";
+
+/**
+ * Returns a stable hash of a file's contents for use in module cache keys.
+ */
+export const getFileContentHash = async (filepath: string): Promise<string> => {
```
briantstephan, line 6: Does this need to be async?

packages/pages/src/util/applyReverseProxy.ts
```diff
@@ -64,10 +76,27 @@ export const buildReverseProxyOverride = (reverseProxyPrefix: string): ReversePr
 
   return {
     reverseProxyPrefix: trimmedReverseProxyPrefix,
-    assetsDir: `${subpath}/assets`,
+    subpath,
+  };
+};
+
+/**
+ * Validates and parses a reverse proxy prefix into the concrete build-time
```
briantstephan: nit, this doesn't validate, `parseReverseProxyPrefix` does the validation

packages/pages/src/util/applyReverseProxy.ts
```diff
@@ -97,6 +129,19 @@ export const applyReverseProxyOverride = (
     throw new Error(`Cannot apply reverseProxyPrefix because ${viteConfigPath} does not exist.`);
   }
 
+  const configuredAssetsPath = readAssetsDirFromViteConfigSource(
+    projectStructure.config.subfolders.assets,
+    viteConfigPath
+  );
+  const assetsPathPrefix = `${parsedReverseProxyPrefix.subpath}/`;
```
briantstephan: I might just be confused here, but it seems like if the build runs again with a different reverse-proxy prefix (for example, `stores` instead of `locations`), the old prefix would get included in the new asset path (e.g. `stores/locations/assets`). Do we need to strip any existing prefix before applying the new one?

Feel free to ack if I'm missing something.

packages/pages/src/util/fileContentHash.ts
a-friedman: Yes, because `globalThis.crypto.subtle.digest` is async and returns a promise

packages/pages/src/util/applyReverseProxy.ts
a-friedman: Done

packages/pages/src/util/applyReverseProxy.ts
a-friedman: Good catch. Updated to read the RPP from config.yaml and remove the prefix parsed from that instead of the prefix passed in.

briantstephan (approved):

jwartofsky-yext (approved):

PR #626, opened 2026-07-28
fix: make getFileContentHash sync

a-friedman:
Brian pointed out that this function no longer needed to be async after updating to use `createHash`. Change to be sync

TEST=auto

Ran unit tests


asanehisa (approved):

briantstephan (approved):
Thanks!

mkilpatrick (approved):

PR #627, opened 2026-07-28
chore: pin npm version to 11.18.0

a-friedman:
The latest version of npm is now 12, which is incompatible with node 20. Pin to npm 11

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Update npm
         # npm version npm 11.5.1 or later needed for Trusted Publishing
-        run: npm install -g npm@11.18.0
+        run: npm install -g npm@11
 
       - name: Verify package version and get npm tag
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - name: Update npm
         # npm version npm 11.5.1 or later needed for Trusted Publishing
-        run: npm install -g npm@latest
+        run: npm install -g npm@11.18.0
```
asanehisa: just npm@11 will work! https://github.com/yext/visual-editor/commit/8cba4f7a2ed55c8e558b63dc54eb70e9b9473caf

briantstephan (dismissed):

.github/workflows/publish.yml
a-friedman: Done

briantstephan (approved):

